PR #14, opened 2018-07-09
Import via replcompletion

sebasguts:


sebasguts:
I removed the remark about built-in functions.

PR #19, opened 2018-08-03
libjulia ignores .juliarc.jl, let GAP load it if wanted

ThomasBreuer:
- added the user preference `IncludeJuliarcFile` for the package
  `JuliaInterface`

- if this user preference has a suitable value then include the
  `.juliarc.jl` file in question into Julia


sebasguts:
Thank you :)

PR #20, opened 2018-08-14
Replace GAP orbit implementation

fingolfin:
The new one is much faster. Indeed, for me it is faster than the Julia
version. However, that also could be optimized; the main reason I did not yet
do that is that one needs to access a hash function for the orbit elements
(here: GAP integers), and I did not yet bother to figure out how to do that.

### DIFF
--- a/JuliaInterface/example/orbits.gi
+++ b/JuliaInterface/example/orbits.gi
@@ -1,30 +1,29 @@
 LoadPackage( "JuliaInterface" );
 
 bahn := function( element, generators, action )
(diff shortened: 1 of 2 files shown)


fingolfin:
My plan is to also improve the Julia side of this. But as long as I only have `JuliaInterface` in GAP for doing it, this is very painful; it'd be much more convenient if I could instead use LibGAP.jl and then develop this from within the Julia REPL. So in a sense, this is blocked by issue #32

PR #23, opened 2018-08-21
Change some INTOBJ_INT uses

fingolfin:
Instead use ObjInt_Int8, ObjInt_UInt8 etc. to deal with potential overflows:
not all 64 bit values fit into a GAP immediate integers; and of course on 32bit
systems things are different again

That said, I am not even sure Julia works in 32bit mode? In that case, the changes for (u)int32_t are not needed, and I would recommend adding a `static_assert` somewhere for rejecting builds against a 32bit GAP.

### DIFF
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -208,10 +208,10 @@ Obj __ConvertedFromJulia_internal( jl_value_t* julia_obj )
 
     // small int
     if(jl_typeis(julia_obj, jl_int64_type)){
-        return INTOBJ_INT( jl_unbox_int64( julia_obj ) );
+        return ObjInt_Int8( jl_unbox_int64( julia_obj ) );
     }
     if(jl_typeis(julia_obj, jl_int32_type)){
-        return INTOBJ_INT( jl_unbox_int32( julia_obj ) );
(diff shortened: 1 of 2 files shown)


fieker:
On Tue, Aug 21, 2018 at 02:05:39PM +0000, Max Horn wrote:
> Instead use ObjInt_Int8, ObjInt_UInt8 etc. to deal with potential overflows:
> not all 64 bit values fit into a GAP immediate integers; and of course on 32bit
> systems things are different again
> 
> That said, I am not even sure Julia works in 32bit mode? In that case, the changes for (u)int32_t are not needed, and I would recommend adding a `static_assert` somewhere for rejecting builds against a 32bit GAP.
There is a 32 bit julia and as of 6 months ago, most of Nemo/Hecke also
worked in 32 bits (there were a few int/long mistakes in the
c-interface)...
I am not sure though that we are going to support this long term...
(My old notebook was/is 32 bits, so every longer train trip I debugged
the 32 bit version)
Claus

sebasguts:
@fieker Do you think it is really necessary to support 32 bit?

I would like to just do `static_asserts` that just stops compiling JuliaInterface on 32 bit instead of having to keep in mind those differences all the time. There are not that many 32bit systems around anymore, specially on real computers.

PR #24, opened 2018-09-17
re-added the julia version to the package banner

ThomasBreuer:
- added the component `BannerFunction` to `JuliaInterface/PackageInfo.g`,
  which can produce a banner string at runtime, at least from GAP 4.10 on;

- changed the `Dependencies` component to require at least GAP 4.10

### DIFF
--- a/JuliaInterface/PackageInfo.g
+++ b/JuliaInterface/PackageInfo.g
@@ -19,7 +19,7 @@ Persons := [
     IsMaintainer := true,
     FirstNames := "Sebastian",
     LastName := "Gutsche",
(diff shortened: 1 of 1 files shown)


fingolfin:
Note that GAP, if linked against Julia, also prints the Julia version as parts of its banner.

sebasguts:
I am thinking that we might want to also print the Julia version in the JuliaInterface banner.

From the GAP banner, it is easily overlooked, and it might be beneficial to have it printed again in the JuliaInterface banner. Although, we might just want to use the string from GAP again (if it stores it), instead of asking Julia again.

PR #44, opened 2018-09-23
C func call

sebasguts:


JuliaInterface/src/JuliaInterface.c
```diff
@@ -207,6 +207,107 @@ Obj NewJuliaFunc(jl_function_t* function, int autoConvert)
...
 }
 
+/*
+ * C function pointer calls
+ */
+
+typedef Obj (*func_ptr_0_args)();
+Obj DoCallJuliaCFunc0Arg( Obj func )
+{
+    func_ptr_0_args function = (func_ptr_0_args)jl_unbox_voidpointer((jl_value_t*)FEXS_FUNC(func));
```
fingolfin: I would put `jl_unbox_voidpointer((jl_value_t*)FEXS_FUNC(func))` into a helper function. That'll make the code more readable, and also make it trivial to change how the func pointer is stored.

JuliaInterface/src/JuliaInterface.c
```diff
@@ -207,6 +207,107 @@ Obj NewJuliaFunc(jl_function_t* function, int autoConvert)
...
+    return function(arg1,arg2,arg3,arg4,arg5,arg6);
+}
+
+Obj NewJuliaCFunc(void* function, int nr_args, char* args)
+{
+
+    Obj func = NewFunctionC("", nr_args, args, 0);
+
+    switch(nr_args){
+        case 0:
```
fingolfin: Off-topic, but: it'd be great if there was a `.clang-format` file in this repo. Then the code in this file could be formatted uniformly. I normally try to adapt to whatever code formatting conventions are used by a project, but here, I really can't tell what those are, resp. find them so different from any I used before that I have a hard time following them. In either case, a `.clang-format` would simplify writing patches for me a lot.

JuliaInterface/src/JuliaInterface.c
```diff
@@ -207,6 +207,107 @@ Obj NewJuliaFunc(jl_function_t* function, int autoConvert)
...
+            SET_HDLR_FUNC(func, 3, DoCallJuliaCFunc3Arg);
+            break;
+        case 4:
+            SET_HDLR_FUNC(func, 4, DoCallJuliaCFunc4Arg);
+            break;
+        case 5:
+            SET_HDLR_FUNC(func, 5, DoCallJuliaCFunc5Arg);
+            break;
+        case 6:
+            SET_HDLR_FUNC(func, 6, DoCallJuliaCFunc6Arg);
```
fingolfin: While you can do that, the more "natural" way would be to pass the handler in the fourth argument to `NewFunctionC`. But I guess it doesn't really matter in the end.

JuliaInterface/src/JuliaInterface.c
```diff
@@ -207,6 +207,107 @@ Obj NewJuliaFunc(jl_function_t* function, int autoConvert)
...
+            break;
+        default:
+            ErrorQuit("Only 0-6 arguments are supported", 0, 0);
```
fingolfin: Perhaps move this error to *before* calling `NewFunctionC`? Of course it shouldn't *really* matter thanks to garbage collection, but it just feels "more natural" to me to just always perform argument validation before using those arguments.

JuliaInterface/gap/BindCFunction.gi
```diff
@@ -27,15 +27,16 @@ InstallGlobalFunction( JuliaBindCFunction,
         return;
     fi;
 
-    cfunction_call_string := Concatenation( "@cfunction(GAP.prepare_func_for_gap(", julia_name,")" );
```
fingolfin: Also, is `prepare_func_for_gap` still used anywhere else? If not, remove it?

JuliaInterface/gap/BindCFunction.gi
```diff
@@ -27,15 +27,16 @@ InstallGlobalFunction( JuliaBindCFunction,
...
     od;
-    Remove( cfunction_call_string );
-    cfunction_call_string := Concatenation( cfunction_call_string, "))" );
-
-    return _JuliaBindCFunction( cfunction_call_string, nr_args, arg_names );
+    cfunction_call_string := Concatenation( cfunction_call_string, "));" );
+    cfunction_call_string := Concatenation( cfunction_call_string, "\nreinterpret(Ptr{Cvoid},temp)");
+    function_ptr := JuliaEvalString( cfunction_call_string );
+    arg_string := JoinStringsWithSeparator( arg_names, "," );
+    return _NewJuliaCFunc( function_ptr, nr_args, arg_string );
```
fingolfin: This code really could use some comments!

JuliaInterface/example/function_perform.gi
```diff
@@ -0,0 +1,48 @@
+LoadPackage( "JuliaInterface" );
+
+dirs:= DirectoriesPackageLibrary( "JuliaInterface", "example" );
+
+JuliaIncludeFile( Filename( dirs, "function_perform.jl" ) );
+
+if not IsBound(_JuliaFunctionByModuleNoConv) then
+  _JuliaFunctionByModuleNoConv:= _JuliaFunctionByModule;
+  _JuliaFunctionNoConv:= _JuliaFunction;
```
fingolfin: This `if` block is not really needed.

And beyond that, as I mentioned elsewhere, I'd really merge `_JuliaFunctionByModuleNoConv`, `_JuliaFunctionByModule`, `_JuliaFunctionNoConv`, `_JuliaFunction` into a single function.

JuliaInterface/gap/BindCFunction.gi
```diff
@@ -10,7 +10,7 @@
 
 InstallGlobalFunction( JuliaBindCFunction,
   function( julia_name, nr_args, arg_names )
-    local cfunction_call_string, i, cfunc;
+    local cfunction_call_string, i, cfunc, function_ptr, arg_string;
```
fingolfin: Why does one need to specify `nr_args`? Isn't that always equal to `Length(arg_names)`?

JuliaInterface/example/function_perform.gi
```diff
@@ -0,0 +1,48 @@
...
+blafuncC := JuliaBindCFunction( "blafunc", 2, ["a", "b"] );
+
+GASMAN("collect"); ListX([1..10^5], [1..10], {i,j} -> i);; time;
+GASMAN("collect"); ListX([1..10^5], [1..10], ReturnFirst);; time;
+GASMAN("collect"); ListX([1..10^5], [1..10], myfunc);; time;
+GASMAN("collect"); ListX([1..10^5], [1..10], blafunc);; time;
+# GASMAN("collect"); ListX([1..10^5], [1..10], myfuncC);; time;
+GASMAN("collect"); ListX([1..10^5], [1..10], blafuncC);; time;
+GASMAN("collect"); ListX([1..10^5], [1..10], myfuncNoConv);; time;
+GASMAN("collect"); ListX([1..10^5], [1..10], blafuncNoConv);; time;
```
fingolfin: Those `time;` statements are only valid in the interpreter; putting them into `.g` or `.gi` file is not really appropriate.

JuliaInterface/example/function_perform.jl
```diff
@@ -0,0 +1,25 @@
...
+    return a;
+end
+
+function myfunc( a::Int64, b::Int64 )
+    return a;
+end
+
+end
+
+function blafunc(a, b)
```
fingolfin: I understand that this is just my hacked together example, but if this is added to the repo, perhaps let's choose some better names? E.g. 
* Stuff -> function_perform to match the file (and BTW, perhaps use `function_calls` instead? That'd at least match the GAP nomenclature, and also fits with the `jl_call*` function names
* myFunc -> returnFirstWithMethodDispatch
* blafunc -> returnFirstGeneric

Well, or something along these lines.

JuliaInterface/example/function_perform.gi
sebasguts: I will leave them in for now. I would prefer to have two versions of this file
1. a testfile
2. a version that I can copy&paste for demo proposes

sebasguts:
I think I addressed all comments (except the clang-format one, we will do this in a single commit later).

JuliaInterface/configure.ac
```diff
@@ -55,7 +55,7 @@ AC_ARG_WITH([julia],
 )
 JULIA_CPPFLAGS="-I$JULIA/include/julia"
 JULIA_LDPATH="$JULIA/lib"
-JULIA_LDFLAGS="-L$JULIA_LDPATH"
+JULIA_LDFLAGS="-L$JULIA_LDPATH -Wl,-rpath=$JULIA_LDPATH"
```
fingolfin: This is not needed anymore

PR #62, opened 2018-10-09
Added new module objects and structure.

### LINKED ISSUE #56: Revise `ImportJuliaModuleIntoGAP`
@sebasguts came up with the following plan: the global variable `Julia` will point to a custom GAP object, for which methods for `\.` and `IsBound.` (and possibly `RecNames`) are installed. When accessing `Julia.foobar`, it queries if `Main.foobar` exists; if not, it errors out, otherwise that object is returned.

On the next level, Julia module objects (i.e., `jl_module_t`) are wrapped into custom GAP objects (similar or even identical to the `Julia` object). Then for these the `\.` method of course performs the lookup relative to the `jl_module_t` (in the `Julia` object, this would then be `jl_main_module`).

sebasguts:
Julia modules are now bona fide GAP objects
which one can use to access arbitrary functions
and submodules of them.

Resolves #56

### DIFF
--- a/JuliaInterface/gap/JuliaInterface.gd
+++ b/JuliaInterface/gap/JuliaInterface.gd
@@ -52,122 +52,6 @@ DeclareOperation( "ConvertedFromJulia", [ IsJuliaObject ] );
 #!  Like structural copy, also converts the contents of a Julia list recursively to GAP objects.
 DeclareGlobalFunction( "StructuralConvertedFromJulia" );
 
-## Internal
-BindGlobal( "_JULIAINTERFACE_MODULE_NAME",
-  function( module_name )
-    if Length( module_name ) = 0 then
-        return "Main";
-    else
-        return module_name[ 1 ];
-    fi;
-end );
-
-## Internal
-BindGlobal( "_JULIAINTERFACE_PREPARE_RECORD",
-  function( module_name )
-    local no_import;
-
-    if module_name = "Main" and IsBound( Julia.Main ) then
-        return module_name;
-    fi;
-
-    if module_name = "GAP" then
-        Julia.GAP := rec( _JULIAINTERFACE_NOT_IMPORTED_YET := true );
-        return module_name;
-    fi;
-    
-    if IsBound( Julia.(module_name) ) and ( not IsBound( Julia.(module_name)._JULIAINTERFACE_NOT_IMPORTED_YET ) ) then
-        return fail;
-    fi;
-
-    if not IsBound( Julia.(module_name) ) then
-        no_import := ValueOption( "NoImport" );
-        if no_import = fail then
-            no_import := false;
-        fi;
-        if no_import = false then
-            JuliaEvalString( Concatenation( "import ", module_name ) );
-        fi;
-        Julia.(module_name) := rec( _JULIAINTERFACE_NOT_IMPORTED_YET := true );
-    fi;
-
-    return module_name;
-
-end );
-
-#! @Arguments function_name[, module_name]
-#! @Description
-#!  Stores the Julia function <A>function_name</A> from the module <A>module_name</A>
-#!  in the <C>Julia</C> record, as <C>Julia</C>.<A>module_name</A>.<A>function_name</A>,
-#!  to have quick access if the function is needed in several places. The default value for
-#!  <A>module_name</A> is <C>"Main"</C>.
-DeclareGlobalFunction( "BindJuliaFunc" );
-InstallGlobalFunction( BindJuliaFunc,
-  function( julia_name, module_name... )
-    module_name := _JULIAINTERFACE_MODULE_NAME( module_name );
-    if IsBound( Julia.(module_name) ) and IsBound( Julia.(module_name).(julia_name) ) then
-        return;
-    fi;
-    module_name := _JULIAINTERFACE_PREPARE_RECORD( module_name );
-    if module_name = fail then
-        return;
-    fi;
-    Julia.(module_name).(julia_name) := JuliaFunction( julia_name, module_name );
-end );
-
-#! @Arguments object_name[, module_name]
-#! @Description
-#!  Stores the Julia object <A>object_name</A> from the module <A>module_name</A>
-#!  in the <C>Julia</C> record, as <C>Julia</C>.<A>module_name</A>.<A>object_name</A>,
-#!  to have quick access if the object is needed in several places. The default value for
-#!  <A>module_name</A> is <C>"Main"</C>.
-DeclareGlobalFunction( "BindJuliaObj" );
-InstallGlobalFunction( BindJuliaObj,
-  function( julia_name, module_name... )
-    module_name := _JULIAINTERFACE_MODULE_NAME( module_name );
-    if IsBound( Julia.(module_name) ) and IsBound( Julia.(module_name).(julia_name) ) then
-        return;
-    fi;
-    module_name := _JULIAINTERFACE_PREPARE_RECORD( module_name );
-    if module_name = fail then
-        return;
-    fi;
-    Julia.(module_name).(julia_name) := JuliaGetGlobalVariable( julia_name, module_name );
-end );
-
-#! @Arguments function_name[, module_name]
-#! @Description
-#!  Returns the julia function <A>function_name</A> from the module <A>module_name</A>.
-#!  As a side effect, it also
-#!  stores the Julia function <A>function_name</A> from the module <A>module_name</A>
-#!  in the <C>Julia</C> record, as <C>Julia</C>.<A>module_name</A>.<A>function_name</A>,
-#!  to have quick access if the function is needed in several places. The default value for
-#!  <A>module_name</A> is <C>"Main"</C>.
-DeclareGlobalFunction( "GetJuliaFunc" );
-InstallGlobalFunction( GetJuliaFunc,
-  function( julia_name, module_name... )
-    module_name := _JULIAINTERFACE_MODULE_NAME( module_name );
-    BindJuliaFunc( julia_name, module_name );
-    return Julia.(module_name).(julia_name);
-end );
-
-#! @Arguments object_name[, module_name]
-#! @Description
-#!  Returns the julia object <A>object_name</A> from the module <A>module_name</A>.
-#!  As a side effect, it also
-#!  stores the Julia object <A>object_name</A> from the module <A>module_name</A>
-#!  in the <C>Julia</C> record, as <C>Julia</C>.<A>module_name</A>.<A>object_name</A>,
-#!  to have quick access if the object is needed in several places. The default value for
-#!  <A>module_name</A> is <C>"Main"</C>.
-DeclareGlobalFunction( "GetJuliaObj" );
-InstallGlobalFunction( GetJuliaObj,
-  function( julia_name, module_name... )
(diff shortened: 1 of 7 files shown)


JuliaInterface/gap/JuliaInterface.gd
```diff
@@ -203,6 +87,23 @@ DeclareGlobalFunction( "JuliaIncludeFile" );
...
+#!  TODO
+DeclareGlobalVariable( "Julia" );
+
+#! @Arguments module
+#! @Returns a list of strings
+#! @Description
+#!  Returns a list of names of components currently bound
+#!  in <A>module</A>. Please note that this does usually
+#!  not reflect all symbols bound to a module on the
+#!  &Julia; side.
```
ThomasBreuer: Perhaps be more explicit here.
I think the idea is that `module` shows what the GAP side currently knows about the Julia module,
and that there may be more symbols, even if the Julia module had been imported some time ago.
(In particular, the value returned by the function is not cached.)

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,74 @@ InstallMethod( CallFuncList,
...
+InstallMethod( IsBound\.,
+               [ IsJuliaModule, IsPosInt ],
+    {module,rnum} -> IsBound\.( module!.storage, rnum  ) );
+
+
+InstallMethod( Unbind\.,
+               [ IsJuliaModule, IsPosInt ],
+  function( module, rnum )
+    Unbind\.( module!.storage, rnum );
+end );
```
ThomasBreuer: Just for curiosity:
Why would one want to unbind symbols?
(One situation would be in the `Main` module if one could unbind Julia variables;
but is this possible?)

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -140,18 +195,11 @@ InstallGlobalFunction( ImportJuliaModuleIntoGAP,
...
-    od;
-    julia_list_func := JuliaFunction( "get_variable_symbols_in_module", "GAPUtils" );
-    variable_list := StructuralConvertedFromJulia( julia_list_func( JuliaModule( name ) ) );
-    for i in variable_list do
-        current_module_rec.(i) := JuliaGetGlobalVariable( i, name );
+    current_module := Julia.(name);
+    julia_list_func := JuliaFunction( "get_symbols_in_module", "GAPUtils" );
+    list := StructuralConvertedFromJulia( julia_list_func( current_module!.julia_pointer ) );
+    for i in list do
+        i := current_module.(i);
```
ThomasBreuer: I would just call `\.` but not assign the return value.

JuliaInterface/src/JuliaInterface.c
```diff
@@ -376,25 +376,41 @@ Obj NewJuliaObj(jl_value_t * C)
...
+    }
+    ErrorMayQuit("argument is not a julia object or string", 0, 0);
+    return 0;
+}
+
+
 /*
- * Returns the function with name <string> from the Julia main module.
+ * Returns the function from the Object <func>
+ * or the function with name <string> from
```
ThomasBreuer: There is no &lt;string&gt; anymore to refer to.

JuliaInterface/src/JuliaInterface.c
```diff
@@ -376,25 +376,41 @@ Obj NewJuliaObj(jl_value_t * C)
...
-Obj Func_JuliaFunctionNoConv(Obj self, Obj string)
+Obj Func_JuliaFunctionNoConv(Obj self, Obj func)
 {
-    jl_function_t * function =
-        jl_get_function(jl_main_module, CSTR_STRING(string));
-    if (function == 0)
-        ErrorMayQuit("Function is not defined in julia", 0, 0);
-    return NewJuliaFunc(function, 0);
+    jl_function_t * function = get_function_from_obj_or_string(func);
+    return NewJuliaFunc(function, 1);
```
ThomasBreuer: The second argument should be 0.

JuliaInterface/src/JuliaInterface.c
```diff
@@ -748,10 +764,20 @@ Obj Func_JuliaGetGlobalVariable(Obj self, Obj name)
...
-Obj Func_JuliaGetGlobalVariableByModule(Obj self, Obj name, Obj module_name)
+Obj Func_JuliaGetGlobalVariableByModule(Obj self, Obj name, Obj module)
 {
-    jl_sym_t *    symbol = jl_symbol(CSTR_STRING(name));
-    jl_module_t * module_t = get_module_from_string(CSTR_STRING(module_name));
+    jl_module_t * module_t = 0;
+    if (IS_JULIA_OBJ(module)) {
+        module_t = (jl_module_t *)GET_JULIA_OBJ(module);
+    }
+    if (IS_STRING_REP(module)) {
```
ThomasBreuer: why not `else if`?

JuliaInterface/gap/JuliaInterface.gi
sebasguts: Well, technically, any import could override variables in `Main`, so `Unbind` could potentially be useful.

Granted, I just added it for completeness ;)

JuliaInterface/gap/JuliaInterface.gi
sebasguts: Good idea.

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,74 @@ InstallMethod( CallFuncList,
...
-##
-##  We support '<juliaobj>[<i>]' and '<juliaobj>[<i>, <j>]' in general.
-##  This is useful for example if <juliaobj> is a tuple.
-##
-BindJuliaFunc( "getindex", "Base" );
+BindGlobal( "_JULIA_MODULE_TYPE", JuliaEvalString( "Module" ) );
+BindGlobal( "_JULIA_FUNCTION_TYPE", JuliaEvalString( "Function" ) );
+BindGlobal( "_JULIA_ISA", JuliaFunction( "isa" ) );
+
+BindGlobal( "_NewJuliaModule",
```
fingolfin: I find this name confusing: It refers to `IsJuliaModule` type objects, of course, but I keep reading it as "create a new Julia module", i.e., on the Julia side. So it is ambiguous. Perhaps: `_WrapJuliaModule`?

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,74 @@ InstallMethod( CallFuncList,
...
+
+    rnam := NameRNam( rnum );
+
+    global_variable := _JuliaGetGlobalVariableByModule( rnam, module!.julia_pointer );
+
+    if ConvertedFromJulia( _JULIA_ISA( global_variable, _JULIA_FUNCTION_TYPE ) ) then
+        global_variable := _JuliaFunction( global_variable );
+    elif ConvertedFromJulia( _JULIA_ISA( global_variable, _JULIA_MODULE_TYPE ) ) then
+        global_variable := _NewJuliaModule( rnam, global_variable );
+    fi;
```
fingolfin: Perhaps the above code should eventually go into a new high(er) level helper, say `JuliaGetGlobalVariableByModule`? (No need to change it in this PR, though)

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,74 @@ InstallMethod( CallFuncList,
...
+
+    \.\:\=( module!.storage, rnum, global_variable );
+    return global_variable;
+
+end );
+
+InstallMethod( \.\:\=,
+               [ IsJuliaModule, IsPosInt, IsObject ],
+  function( module, rnum, obj )
+    \.\:\=( module!.storage, rnum, obj );
```
fingolfin: Do we really want to support this??? I would rather call `ErrorNoReturn("Assigning to Julia module wrapper is not supported")`. Or else, if it was supported, I'd expect that it'd also affect the Julia side.

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,74 @@ InstallMethod( CallFuncList,
...
+end );
+
+
+InstallMethod( IsBound\.,
+               [ IsJuliaModule, IsPosInt ],
+    {module,rnum} -> IsBound\.( module!.storage, rnum  ) );
```
fingolfin: Hmm, that's not the semantic I wold expect. I'd expect this to look more like this:
```gap
function(module, rnum)
  if IsBound\.( module!.storage, rnum  ) then return true; fi;
  rnam := NameRNam( rnum );
  # test if Julia knows a variable with that name
  return fail <> _JuliaGetGlobalVariableByModule( rnam, module!.julia_pointer );
end;
```
Well, this assumes that `_JuliaGetGlobalVariableByModule` returns fail if there is no such variable, which may be wrong (I did not check what it does), but I hope the general idea is clear.

JuliaInterface/src/JuliaInterface.c
sebasguts: Indeed.

JuliaInterface/gap/JuliaInterface.gi
fingolfin: I would call `Error` here, just as for assignment.

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,74 @@ InstallMethod( CallFuncList,
...
+BindGlobal( "_NewJuliaModule",
+  function( name, julia_pointer )
+    local module;
+
+    module := rec( storage := rec( ),
+                   julia_pointer := julia_pointer );
+
+    Objectify( TheTypeOfJuliaModules, module );
+
+    SetName( module, name );
```
fingolfin: The name is what is printed, right? Instead of printing just, say `Main` or `JSON`, I think it'd be nicer if we printed e.g. `Julia module "Main"` or `JuliaModule("Main")` or whatever. Perhaps with a custom `ViewObj` or `ViewString` method?

JuliaInterface/gap/JuliaInterface.gi
sebasguts: I intentionally not did this. I think it is a good idea to have `JuliaGetGlobalVariableByModule` return an object. I am not sure it should really check for the type.

JuliaInterface/gap/JuliaInterface.gi
sebasguts: Yep, that makes more sense, as the binding happens automatically. Did not think about this.

JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -57,11 +57,81 @@ InstallMethod( CallFuncList,
...
+
+    module := rec( storage := rec( ),
+                   julia_pointer := julia_pointer );
+
+    ObjectifyWithAttributes( module, TheTypeOfJuliaModules,
+                             Name, Concatenation( "<Julia module ", name, ">" ) );
+
+    return module;
+
+end );
```
ThomasBreuer: If you use `ObjectifyWithAttributes` then you can turn the whole function body into a call
`return ObjectifyWithAttributes( rec( ... ), ... );`.

sebasguts: Okay, we can keep that in mind, but I think it is more readable this way.

PR #64, opened 2018-10-10
New makefiles

sebasguts:
* Using gac for JuliaExperimental
* Configure and Make for GAPJulia


.gitignore
```diff
@@ -18,7 +18,7 @@
 /doc/*.tex
 /doc/*.toc
 /doc/manual.pdf
-/doc/NormalizInterface-bib.xml.bib
```
fingolfin: There is more Normaliz specific stuff below (including the /test-suite.log and more)

JuliaExperimental/.gitignore
```diff
@@ -18,46 +18,31 @@
 /doc/*.tex
 /doc/*.toc
 /doc/manual.pdf
-/doc/NormalizInterface-bib.xml.bib
```
fingolfin: So right now there are three .gitignore files. OK, it makes some sense that each of the GAP package has its own, but then the one on the global scope could / should be trimmed down a lot.

Or you merge them into a single .gitignore at the root, which then replaces rules like `/doc/*.tex` by `doc/*.tex`

configure
```diff
@@ -0,0 +1,45 @@
...
+    echo "is not properly compiled (do \"./configure && make\" there first)."
+    echo
+    echo Aborting... No Makefile is generated.
+    echo
+    exit 1
+fi
+
+if [[ "$GAPPATH" != /* ]]; then
+    GAPPATH=../${GAPPATH}
+fi
```
fingolfin: It might be safer to simply convert the path into a global one, say via `GAPPATH=$(cd ${GAPPATH} && pwd)`.

configure
```diff
@@ -0,0 +1,45 @@
...
+
+echo "all: ${packages}" > Makefile
+
+for package in $packages; do
+    echo "Configuring ${package}"
+    cd $package
+    ./configure $GAPPATH
+    cd ..
+    sed "s/@@PACKAGE@@/${package}/g" Makefile.in >> Makefile
+done
```
fingolfin: Is it useful to generate the `Makefile` this way, instead of simply adding it directly to the repository? That certainly would make it easier to edit it in the future, and add subproject specific build rules.

PR #66, opened 2018-10-10
Update JuliaExperimental for Julia 1.0 and latest JuliaInterface

fingolfin:
Note that the test still produce diffs and even crash for me, I'll look into that next.

### DIFF
--- a/JuliaExperimental/julia/zlattice.jl
+++ b/JuliaExperimental/julia/zlattice.jl
@@ -15,7 +15,7 @@ import Base: abs, convert, copy, deepcopy, diag, haskey, inv, lcm, length,
 export LLLReducedGramMat, ShortestVectors, OrthogonalEmbeddings
 
 
-doc"""
+raw"""
     LLLReducedGramMat( grammatrix::Array{Int,2}, y::Rational{Int} = 3//4 )
 > Return a dictionary with the following components.
 >   `remainder`:      the reduced Gram matrix (`Array{Rational{Int},2}`)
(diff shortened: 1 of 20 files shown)


JuliaExperimental/tst/testall.g
```diff
@@ -1,35 +1,14 @@
 #
-# JuliaExperimental: Experimental code for the GAP Julia integration
+# JuliaInterface: Test interface to julia
```
ThomasBreuer: really?

JuliaExperimental/julia/zlattice.jl
```diff
@@ -9,13 +9,15 @@ module GAPZLattice
...
+import LinearAlgebra: diag
+
+import Base: abs, convert, copy, deepcopy, haskey, inv, lcm, length,
+             map, push!, sign, size, sum, trunc, zero, zeros
 
 export LLLReducedGramMat, ShortestVectors, OrthogonalEmbeddings
 
 
-doc"""
+raw"""
```
ThomasBreuer: raw?

JuliaExperimental/tst/testall.g
fingolfin: Ouch

JuliaExperimental/julia/zlattice.jl
fingolfin: The problem is the `$` later in the string, which Julia tries to interpolate, like in a shell script. We could remove that; but @sebasguts suggested this alternative change, which disables variable interpolation for the string.

sebasguts: We could probably also do `\$`, but I would try to investigate into this when we actually start to build julia documentation.

Issue #32, opened 2018-09-19
Merge LibGAP.jl into GAPJulia ?


fingolfin:
... as discussed during GAP Days Fall 2018

sebasguts:
Currently, the Julia package manager seems to be very strict about the structure of a Julia module.
It needs to have a `deps` folder and `Manifest.toml` and `Project.toml` in its root folder.
So how would we want to unify it?

Currently, `LibGAP.jl` is not a proper Julia module at all. We could just make it another subfolder in this repo, and be okay with that we could currently not use it as a proper Julia module, or giving the whole Repo the structure of a Julia module and just having the two GAP packages around.

wbhart:
I my opinion, it is worth thinking about whether a separate LibGAP package could have wider benefits for the community at large. I had always envisioned the two being a single Julia package, but it need not be that way if there is a good reason to keep them separated.

In short, you may increase the usership of GAP in the Julia community with one or the other approach.

ThomasBreuer:
`LibGAP.jl` and GAP's `GAPJulia` package belong together because they provide the two directions
of an interface between GAP and Julia, thus one wants to have them together;
from the viewpoint of Julia, it would be strange if one could not call some GAP functionality
just because it needs Julia.
(But this is only a logical dependency not a technical one,
and it would not cause real problems to have separate repositories for `LibGAP.jl` and `GAPJulia`
as long as there are no technical dependencies.)

If Julia wants the repository of a Julia module to have a particular structure,
can't we provide a repository with that structure for `LibGAP.jl` and then add the stuff from `GAPJulia` to it,
and Julia's package manager will be happy?
(This is how I interpret the last part in the above comment by @sebasguts.)

sebasguts:
Yes, that is indeed correct. We could even still pack it in GAP's `pkg` folder. It would just look ugly.

fingolfin:
I propose the following plan: First, we update https://github.com/oscar-system/LibGAP.jl to work with the current version of GAPJulia (possibly on a branch there, as to not negatively affect the work by @CarloSircana in his `CarloLibGAPTest` repo). Once that works, we simply subtree-merge this into the GAPJulia repo, into a subdir `LibGAP.jl`. 

[Alternatively, we could reverse the order, and first subtree-merge LibGAP.jl, then work on getting it to run].

[...]

PR #69, opened 2018-10-11
Reduce INITIALIZE_JULIA_CPOINTER calls, resp. move to right package

fingolfin:


sebasguts:
@ThomasBreuer @fingolfin I will have a look at them later. Some of the removed pointers might still be in used by LibGAP.jl, but potentially unnecessary. I think we should better push for #32 and then see what we can safely remove.

I am sorry for the bad design at that point, i.e., declaring pointers here that are just used somewhere invisible from this repo.

fingolfin:
I actually checked against LibGAP.jl, too

sebasguts:
Thank you :)

PR #72, opened 2018-10-15
Made julia error messages handled by GAP

### LINKED ISSUE #60: Error messages are just printed, not given to GAP
Currently, the content of Julia error message is just printed to `stderr`, not returned to GAP.

Fix this by providing an `IOBuffer` for the `showerror` call, read out the string in this `IOBuffer`, and give it back to GAP.

sebasguts:
Closes #60

### DIFF
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -4,6 +4,11 @@
 
 #include "JuliaInterface.h"
 
+jl_value_t *    JULIA_ERROR_IOBuffer;
+jl_function_t * JULIA_FUNC_take_inplace;
+jl_function_t * JULIA_FUNC_String_constructor;
+jl_function_t * JULIA_FUNC_showerror;
+
 #include "gap_macros.c"
 
 Obj TheTypeJuliaObject;
@@ -909,6 +914,13 @@ static Int InitKernel(StructInitInfo * module)
     CopyObjFuncs[T_JULIA_OBJ] = &JuliaObjCopyFunc;
     CleanObjFuncs[T_JULIA_OBJ] = &JuliaObjCleanFunc;
     IsMutableObjFuncs[T_JULIA_OBJ] = &JuliaObjIsMutableFunc;
+
+    // Initialize necessary variables for error handling
+    JULIA_ERROR_IOBuffer =
+        jl_eval_string("GAP_JULIA_ERROR_IO_BUFFER = Base.IOBuffer()");
+    JULIA_FUNC_take_inplace = jl_get_function(jl_base_module, "take!");
+    JULIA_FUNC_String_constructor = jl_get_function(jl_base_module, "String");
+    JULIA_FUNC_showerror = jl_get_function(jl_base_module, "showerror");
 
     // Initialize GAP function pointers in Julia
     JuliaInitializeGAPFunctionPointers();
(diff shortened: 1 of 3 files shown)


JuliaInterface/src/JuliaInterface.c
```diff
@@ -910,6 +915,13 @@ static Int InitKernel(StructInitInfo * module)
     CleanObjFuncs[T_JULIA_OBJ] = &JuliaObjCleanFunc;
     IsMutableObjFuncs[T_JULIA_OBJ] = &JuliaObjIsMutableFunc;
 
+    // Initialize necessary variables for error handling
+    JULIA_ERROR_IOBuffer =
+        jl_eval_string("GAP_JULIA_ERROR_IO_BUFFER = IOBuffer()");
+    JULIA_FUNC_take_inplace = (jl_function_t *)jl_eval_string("take!");
+    JULIA_FUNC_String_constructor = (jl_function_t *)jl_eval_string("String");
+    JULIA_FUNC_showerror = (jl_function_t *)jl_eval_string("showerror");
```
fingolfin: Nipick: here you use `jl_eval_string("showerror")`; in the old code you use `jl_get_function(jl_base_module, "showerror")`. The latter *seems* somewhat more robust to me, i.e., in case somebody defined another `showerror` in the namespace active in Julia when `JuliaInterface` is loaded, which hides `Base.showerror`. But perhaps that's not possible anyway or whatever (I don't know Julia well enough, I am just guessing around here)

JuliaInterface/src/JuliaInterface.c
```diff
@@ -910,6 +915,13 @@ static Int InitKernel(StructInitInfo * module)
     CleanObjFuncs[T_JULIA_OBJ] = &JuliaObjCleanFunc;
     IsMutableObjFuncs[T_JULIA_OBJ] = &JuliaObjIsMutableFunc;
 
+    // Initialize necessary variables for error handling
+    JULIA_ERROR_IOBuffer =
+        jl_eval_string("GAP_JULIA_ERROR_IO_BUFFER = Base.IOBuffer()");
```
fingolfin: Sorry for another nitpick, but: is `GAP_JULIA_ERROR_IO_BUFFER` ever used later on? Or are there plans to use it? Otherwise, shouldn't this just be `jl_eval_string("Base.IOBuffer()");`, or even `jl_call0(jl_get_function(jl_base_module, "IOBuffer"))`?

sebasguts: I think both options are not really safe, actually, it is basically a GC issue.

For the second, we would still need to pin the new object during the computation. We can do that, I just wanted to avoid new object creation.

For the first, the object might get deleted some time before even the first error occurs. I am not sure if you can leave it pinned forever.

So, if any of those solutions should be used, it would have to be the second. I can implement that.

fingolfin: I see. Well, so let's leave this as-is for now. But on the long run, we need a mechanism to easily "pin" variables. One way would be to add them to the GAP module state, as those automatically get marked. Or we could add these to "our" Julia module. Or we could set a custom `ExtraMarkFuncBags`

sebasguts: I have honestly tried to get rid of this, but every try I made to create the new `IOBuffer` in the error handling macro lead to a segfault. Trying to pin it using `JL_GC_PUSH1` in the `KernelInit` does also not work.

I could assign it to a global variable in GAP, but then it would be dangling there, which is not better or worse, I guess. So I would guess we can leave it be for now.

PR #80, opened 2018-10-16
HACK on LibGAP.jl -- do not merge

fingolfin:
I am putting this here so that @sebasguts can take any bits that are helpful.


fingolfin:
I squashed it into a single commit with a sane commit message, so from my POV, this could indeed be merged.

PR #90, opened 2018-10-19
Enable Julia code coverage

fingolfin:
This is just a proof of concept, and to see if it works. If it does, we need to implement a way to conditionally enable code coverage in Julia.


fingolfin:
OK, I had to fix https://github.com/JuliaCI/Coverage.jl and get them to make a new release (0.7.0, now out), but this PR finally is "ready" -- except for those crashes, which of course wreck most of the coverage data :-(

fingolfin:
@sebasguts this is now really ready for merging!

PR #91, opened 2018-10-19
JuliaInterface: remove more dead code

fingolfin:
All of the removed functions are potentially useful. However, I believe that
we will be better off reimplementing them in Julia, and then exporting these
Julia functions to GAP.

But maybe my believe is wrong... perhaps if one needs to use these functions in tight loop, having them as pure C is beneficial... but then again, I don't think using them in a tight loop is a good idea either way. I mean, setting a global Julia variable is not something a lot of code should do, if any; certainly not in a tight loop. Likewise, converting from/to GapObj wrappers strikes me as something you want to avoid in efficient code anyway, and would ideally do only once at the start and once at the end.

### DIFF
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -811,6 +811,19 @@ static Obj FuncJuliaModule(Obj self, Obj name)
     return NewJuliaObj((jl_value_t *)julia_module);
 }
 
+// Sets the value of the julia identifier <name> to the julia value the julia
+// object GAP object <julia_val> points to.
+// This function is for debugging purposes
+static Obj FuncJuliaSetVal(Obj self, Obj name, Obj julia_val)
+{
+    jl_value_t * julia_obj = GET_JULIA_OBJ(julia_val);
+    jl_sym_t *   julia_symbol = jl_symbol(CSTR_STRING(name));
+    JULIAINTERFACE_EXCEPTION_HANDLER
+    jl_set_global(jl_main_module, julia_symbol, julia_obj);
+    JULIAINTERFACE_EXCEPTION_HANDLER
+    return 0;
+}
+
 // Returns the julia object GAP object that holds a pointer to the value
 // currently bound to the julia identifier <name>.
 static Obj Func_JuliaGetGlobalVariable(Obj self, Obj name)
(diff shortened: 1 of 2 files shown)


JuliaInterface/src/JuliaInterface.c
```diff
@@ -811,18 +811,6 @@ static Obj FuncJuliaModule(Obj self, Obj name)
     return NewJuliaObj((jl_value_t *)julia_module);
 }
 
-// Sets the value of the julia identifier <name> to the julia value the julia
-// object GAP object <julia_val> points to.
-static Obj FuncJuliaSetVal(Obj self, Obj name, Obj julia_val)
```
sebasguts: I am pretty much against removing this function without implementing it again. I use it often for debugging purposes. We could hide it using `_` or provide a julia implementation. A Julia implementation could look like
```
function set_global(name,val)
           name_sym = Symbol(name)
           eval(:($name_sym=$val))
       end
```

JuliaInterface/src/JuliaInterface.c
```diff
@@ -900,35 +888,6 @@ static Obj Func_JuliaSetGAPFuncAsJuliaObjFunc(Obj self,
     return NULL;
 }
 
-static Obj FuncJuliaSetAsJuliaPointer(Obj self, Obj obj)
```
sebasguts: I would guess this function does nothing anymore anyway, so it could be savely removed

sebasguts:
I am not sure it is super useful to go ahead and just remove functionality. Moving it to Julia is fine, but what is the point in removing functionality that seems unused?

fingolfin:
For starters, it makes it super confusing when looking at the code. Like when I recently worked on adding tests for the "BindCFunction" code, and run into a crash, I wasted quite some time on the old C function for that which was still around but not actually used. That wouldn't have happened if it had just been removed.

Moreover, this unused code tends to bitrot quite quickly -- nobody tests it, and we freely and quickly change things right now. and if if somebody needs some functionality, and sees "oh, there is already a function for it", and tries to use it, and it crashes, it can again waste quite some time until they finally realize "gee, this function cannot work".

So, I believe that one should either remove functions, or else actually use them -- at the very least, some tests should trigger the function and verify it keeps working. But just doing that "to get coverage up" if there is no clear scenario in which they will be useful, seems like a waste of time to me.

JuliaInterface/src/JuliaInterface.c
fingolfin: Of course if you the function, we can keep it, but then there should be a test case for it. And perhaps a comment saying "this function is useful for debugging".

sebasguts:
Agreed that unused code is confusing for starters, and even for myself (@sebasguts worst enemy is still past @sebasguts).

So, I would proceed as follows for now: I will edit the PR to keep my `JuliaSetVal` function, and then add a comment and tests. Replacing it via a pure Julia function, while it seems tempting, might actually not help much, as the access to a pure Julia function needs much more functionality in this package then just the C-Code, so this function is still available when something else does not load. That's why I really want to keep it. Maybe we could move it into a separate file, say `internal.c` or `debug.c`, would that be okay for you?

fingolfin:
As I said before: I am perfectly happy with keeping `JuliaSetVal` or any other functions, as long as there are tests for them, or at least comments that indicate what they are meant to be used for.

In this case, I just added a test for `JuliaSetVal`; combined with the comment you wrote, from my POV, we could merge this now, possibly after squashing it.

sebasguts:
Thank you :)

Issue #70, opened 2018-10-11
JuliaExperimental: replace INTOBJ_INT by ObjInt_Int8


fingolfin:
Using `INTOBJ_INT ` is not safe for inputs that don't fit into a GAP immediate integer. The simple fix is to use `ObjInt_Int8` instead. But I don't quite know how to test that code, so I'll leave it to somebody else to make that change.

fingolfin:
This was resolved by PR #97

Issue #71, opened 2018-10-11
JuliaExperimental: incorrect use of AbsInt and GcdInt; create_rational is dangerous


fingolfin:
JuliaExperimental calls `AbsInt` and `GcdInt` with rationals as arguments, which cannot work and will lead to crashes and/or corrupt data.

Note there is a kernel function `AbsRat` which can be used instead of `AbsInt`. This function is also available on the GAP level as `ABS_RAT`, so it could also be called as `GAP.GAPFuncs.ABS_RAT`. However, this may be slower compared to the ccall, see also issue #68 . Another thing one could try is to simply use `ccall(:AbsRat, ...`, i.e. without setting up a wrapper in `gap_macros.c`

For `GcdInt`, it is unclear to me what the intended meaning is when computing the gcd of two rationals. I'd say it is always 1 (and GAP agrees), except of course if both inputs are 0.

Lastly, `create_rational` is rather dangerous, as it allows creating denormalized rationals, such as 2/1, 8/4,  3/0 or 2/-1. GAP is *not* prepared for this -- it expects rationals to always have a denominator > 1. If that rules is violated, then e.g. the rational `2/1` in GAP is not equal to the integer 2.

I suggest to instead call the GAP kernel function `QuoRat`, which accepts both rationals and integers as arguments.

Lastly, with the current master branch, the wrapper functions `MyFuncLT` in the kernel are no longer necessary, as `LT` etc. all were converted from macros into functions.

PR #97, opened 2018-10-25
Restructure relations of LibGAP.jl and JuliaInterface

sebasguts:
* Made JuliaInterface load the necessary part of LibGAP.jl
* overhauled `GAP.GAPFuncs` to dynamically fetch functions from GAP


sebasguts:
I am still working on this, no need to review right now.

sebasguts:
@fingolfin @ThomasBreuer I am finished with this so far, please have a look.

@ThomasBreuer I think almost all of the `ccalls` in `JuliaExperimental` could be removed, and `Main.GAP.GAPFuncs.Whatever` could be used instead. Please note that in this functions `Int64`s are handled transparently, which means that you can just pass them to those functions without converting them to GAP integers before.

JuliaExperimental/julia/gaprat.jl
```diff
@@ -140,14 +140,14 @@ end
...
-    return Bool( ccall( Main.gap_MyFuncEQ, Int,
+    int_ptr = ccall( :ObjInt_Int, Ptr{Cvoid}, (Int,), b )
+    return Bool( ccall( :MyFuncEQ, Int,
                 (Ptr{Cvoid}, Ptr{Cvoid}), a.obj.ptr, int_ptr ) )
 end
 
 function ==( a::Int, b::GAPRat )
-    int_ptr = ccall( Main.gap_INTOBJ_INT, Ptr{Cvoid}, (Int,), a )
-    return Bool( ccall( Main.gap_MyFuncEQ, Int,
+    int_ptr = ccall( :ObjInt_Int, Ptr{Cvoid}, (Int,), a )
```
fingolfin: Great, it seems you resolve #70

JuliaExperimental/julia/gaprat.jl
```diff
@@ -319,40 +319,40 @@ end
...
 function abs( a::GAPRat )
-    ptr = ccall( Main.gap_AbsInt, Ptr{Cvoid},
+    ptr = ccall( :AbsInt, Ptr{Cvoid},
                 (Ptr{Cvoid},), a.obj.ptr )
     return GAPRat( GapObj(ptr) )
 end
 
 function gcd( a::GAPRat, b::GAPRat )
-    ptr = ccall( Main.gap_GcdInt, Ptr{Cvoid},
+    ptr = ccall( :GcdInt, Ptr{Cvoid},
```
fingolfin: Meta-level: note that I don't believe this code is correct to start with, see issue #71. Alas, of course you are just converting it here. Once we start adding tests for it, we should quickly see if I am right or not :)

JuliaExperimental/src/gap_macros.c
```diff
@@ -66,26 +66,3 @@ Obj create_rational(int numerator, int denominator)
...
-    INITIALIZE_JULIA_CPOINTER(MyFuncQUO);
-    INITIALIZE_JULIA_CPOINTER(MyFuncPOW);
-    INITIALIZE_JULIA_CPOINTER(MyFuncMOD);
-    INITIALIZE_JULIA_CPOINTER(AbsInt);
-    INITIALIZE_JULIA_CPOINTER(create_rational);
-    INITIALIZE_JULIA_CPOINTER(DEN_RAT);
-    INITIALIZE_JULIA_CPOINTER(GcdInt);
-    INITIALIZE_JULIA_CPOINTER(INTOBJ_INT);
-    INITIALIZE_JULIA_CPOINTER(NUM_RAT);
-}
```
fingolfin: Good riddance :)

JuliaInterface/example/orbits.jl
```diff
@@ -1,24 +1,13 @@
...
-    generator_length = gap_LengthPlist(generators)
+    action_func = GAP.GapFunc( action )
+    generator_length = length(generators)
     while length(work_set) != 0
         current_element = pop!(work_set)
         for current_generator_number = 1:generator_length
-            current_generator = gap_ListElement(generators,current_generator_number)
-            current_result = gap_CallFunc2Args(action,current_element,current_generator)
+            current_generator = generators[current_generator_number]
+            current_result = action(current_element,current_generator)
```
fingolfin: Much nicer

JuliaInterface/read.g
```diff
@@ -11,6 +11,9 @@ dirs:= DirectoriesPackageLibrary( "JuliaInterface", "julia" );
 
 JuliaIncludeFile( Filename( dirs, "gaptypes.jl" ) );
 
-AddGapJuliaFuncs();
-
 ImportJuliaModuleIntoGAP( "GAP" );
+
+if not IsBound( JULIAINTERNAL_LOADED_FROM_JULIA ) then
```
fingolfin: Minor (less important) idea: So, right now `JULIAINTERNAL_LOADED_FROM_JULIA` is set by some GAP code which the Julia code generates and runs... Wouldn't it make more sense to instead set such a variable from within the Julia code which we execute (or not) below? I.e. the code in `libgap.jl` could set some Julia or GAP global variable.

That would make the connection more direct.

JuliaExperimental/julia/gaprat.jl
sebasguts: To be fair, this is just a conversion. This code is broken (for example the GapObj call can be removed, and so on. Once this is merged, I will clean up this whole file again

@ThomasBreuer be prepared ;)

JuliaInterface/read.g
sebasguts: I do not think that is possible actually.

The variable is set within the GAP startup in `LibGAP.jl`, and there just because it makes the file loading there explicit (instead of letting the `libgap.jl` file being loaded by GAP, we load it directly.

I do not think this could be set somewhere else, actually.

JuliaExperimental/julia/gaprat.jl
fingolfin: Fair enough

JuliaExperimental/julia/gaprat.jl
```diff
@@ -12,8 +12,6 @@ import Base: zero, -, one, inv, ==, isless, +, *, //, ^, mod, iszero, string,
 
 import Main.GAP: GapObj
 
-import Main.GAP.GAPFuncs: SUM
-
```
fingolfin: I had to add this to make it possible to load this file (good thing we fixed the tests :-)

PR #98, opened 2018-10-29
Fix crash in our tests for JuliaCFuncs

fingolfin:


fingolfin:
The tests are not (yet) disabled on master

PR #99, opened 2018-10-29
Install Nemo and cache it (WIP)

fingolfin:


fingolfin:
To my surprise, installing Nemo.jl only adds ~10 minutes to the Travis build.

fingolfin:
This seems to be working fine now. There is the question of what happens if the Julia code in Nemo "should be" recompiled due to a change in the Julia nightly... not sure whether Julia does that automatically (probably not?), or else how to trigger it manually... I *think* (but might be wrong) that the part taking up most of the time is compiling all the C libraries, so as long as we can avoid that, it should still be useful. Perhaps @wbhart or some other Julia experts have some insights on that?

wbhart:
I believe Julia will recompile if its binary changes. However, it won't recompile Flint and so on, I think, as this shouldn't be needed. You are right that the first build takes a while, due to building all the C libraries. We surely want to eventually move over to using binaries to speed that up.

PR #103, opened 2018-10-30
Fix crashes, add checks and tests, start revising the exception handlingg

fingolfin:


JuliaInterface/src/JuliaInterface.h
```diff
@@ -4,13 +4,6 @@
 #include "src/compiled.h" /* GAP headers */
 #include <julia.h>
 
-extern void handle_jl_exception(void);
```
sebasguts: Can we leave the macro in the header? If anything, this should be part of the official API.

fingolfin: As explained in the commit message, it should be replaced by something that external code actually can safely use w/o having to hard code deep knowledge about Julia internals. This function isn't doing that, and can't.

I plan to add a proper replacement later on, but for now, I think it's better to not expose this low-level function.

The replacement will likely consists of our own set of TRY/CATCH macros; or perhaps a `BEGIN_JULIA / END_JULIA` pair. But I think it's easier to just show the code once it is ready. In the meantime, I didn't want to pile on too many changes in one PR, also to avoid conflicts with your work.

PR #104, opened 2018-10-30
Redo handling of GAP functions in Julia

sebasguts:


fingolfin:
For optimal performance, though, we should have code like this on the Julia side:
```julia
function(func::GapFunc)(arg1, arg2)
    result = ccall(:call_gap_func_2args,Any,
                        (MPtr,Any,Any),func.ptr, arg1, arg2 )
    return result
end
```
On the C side, `call_gap_func_2args` would be a thin wrapper around `CALL_2ARGS`

sebasguts:
Agreed. This is a first approximation.

Note that we would need to distinguish `T_FUNCTION`s from other GAP callables on the Julia side then, or duplicate the generic code at the end for all of those.

PR #107, opened 2018-10-30
Revise how Julia exceptions are handled

fingolfin:


fingolfin:
Good point. No, it is not really. I'll clang-format this PR, too.

fingolfin:
BTW, the names BEGIN_JULIA / END_JULIA are of course not great, feel free to rename them.

Next, one thing that I dislike about them is that one *must not use `return` or `ErrorQuit` inside such a block. That's easy to forget. I'll at least add a comment.

sebasguts:
I think the names are okay for now, no need to change them. try/catch are not really fitting here, so okay.

Can we configure `clang-format` to do the indentation? I would really like it.

fingolfin:
Hmm, looking at https://clang.llvm.org/docs/ClangFormatStyleOptions.html this can be done with `MacroBlockBegin` and `MacroBlockEnd`. However, this requires a recent version of `clang-format` I think (LLVM 4.x or newer). I've updated the PR now.

sebasguts:
Nice one, thank you :)

PR #108, opened 2018-10-30
Add various argument checks, add tests

fingolfin:


fingolfin:
Actually, I just also improved the autoconversion test (via a second commit) a little bit, to actually check the difference it makes...

PR #109, opened 2018-10-30
Added GapFFE type

sebasguts:
Currently, this causes me severe trouble in Julia, so WIP

### DIFF
--- a/JuliaInterface/julia/gaptypes.jl
+++ b/JuliaInterface/julia/gaptypes.jl
@@ -83,6 +83,13 @@ struct GapFunc
 end
 
 """
+    GapFEE
+
+> Wraps a pointer to a GAP FFE immediate object
+> This type is defined in the JuliaInterface C code.
+"""
+
+"""
     (func::GapFunc)(args...)
 
 > This function makes it possible to call GapFunc objects on
--- a/JuliaInterface/read.g
+++ b/JuliaInterface/read.g
@@ -11,6 +11,8 @@ dirs:= DirectoriesPackageLibrary( "JuliaInterface", "julia" );
 
 JuliaIncludeFile( Filename( dirs, "gaptypes.jl" ) );
 
+_InitializeGAPFFEType();
+
 ImportJuliaModuleIntoGAP( "GAP" );
 
 if not IsBound( JULIAINTERNAL_LOADED_FROM_JULIA ) then
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -8,8 +8,7 @@ static jl_value_t *    JULIA_ERROR_IOBuffer;
 static jl_function_t * JULIA_FUNC_take_inplace;
 static jl_function_t * JULIA_FUNC_String_constructor;
 static jl_function_t * JULIA_FUNC_showerror;
-
-#include "gap_macros.c"
+static jl_datatype_t * JULIA_GAPFFE_type;
 
 Obj TheTypeJuliaObject;
 UInt T_JULIA_OBJ;
@@ -30,6 +29,46 @@ void handle_jl_exception(void)
         jl_call1(JULIA_FUNC_String_constructor, string_object);
     ErrorMayQuit(jl_string_data(string_object), 0, 0);
 }
+
+jl_module_t * get_module_from_string(char * name)
+{
+    // It suffices to use JULIAINTERFACE_EXCEPTION_HANDLER here, as
+    // jl_eval_string is part of the jlapi, so don't have to be wrapped in
+    // JL_TRY/JL_CATCH.
+    jl_value_t * module_value = jl_eval_string(name);
+    JULIAINTERFACE_EXCEPTION_HANDLER
+    if (!jl_is_module(module_value))
+        ErrorQuit("Not a module", 0, 0);
+    return (jl_module_t *)module_value;
+}
+
+// This function needs to be called after the `gaptypes.jl`
+// file is loaded into Julia and the GAP module is present.
+// Therefore we do not call in InitKernel, but in the `read.g` file.
+Obj Func_InitializeGAPFFEType()
+{
+    JULIA_GAPFFE_type = (jl_datatype_t *)jl_new_primitivetype(
+        (jl_value_t *)jl_symbol("GapFFE"), get_module_from_string("GAP"),
+        jl_any_type, jl_emptysvec, sizeof(Obj) * 8);
+    jl_set_const(get_module_from_string("GAP"), jl_symbol("GapFFE"),
+                 (jl_value_t *)JULIA_GAPFFE_type);
+    return NULL;
+}
+
+jl_value_t * gap_box_gapffe(Obj value)
+{
+    jl_ptls_t    ptls = jl_get_ptls_states();
+    jl_value_t * v = jl_gc_alloc_typed(ptls, sizeof(Obj), JULIA_GAPFFE_type);
+    *(Obj *)jl_data_ptr(v) = value;
+    return v;
+}
+
+Obj gap_unbox_gapffe(jl_value_t * gapffe)
+{
+    return *(Obj *)jl_data_ptr(gapffe);
+}
+
+#include "gap_macros.c"
 
 // FIXME: get rid of IS_JULIA_FUNC??
 static inline Int IS_JULIA_FUNC(Obj obj)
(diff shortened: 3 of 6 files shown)


JuliaInterface/src/JuliaInterface.c
```diff
@@ -36,6 +35,43 @@ static void handle_jl_exception(void)
...
+typedef struct _jl_gapffe{
+    void* ptr;
+} gapffe_t;
+
+gapffe_t* new_julia_gapffe(Obj value){
+    jl_ptls_t ptls = jl_get_ptls_states();
+    gapffe_t *t = (gapffe_t*)jl_gc_alloc_typed(ptls, sizeof(gapffe_t), JULIA_GAPFFE_type);
+    t->ptr = (void*)value;
+    return t;
+}
```
fingolfin: Clang-format all?

JuliaInterface/src/gap_macros.c
```diff
@@ -1,13 +1,14 @@
 #include "libgap-api.h"
 
+jl_module_t* get_module_from_string(char*);
```
fingolfin: Huh, doesn't this PR *remove* get_module_from_string ?

JuliaInterface/src/JuliaInterface.c
```diff
@@ -36,6 +35,43 @@ static void handle_jl_exception(void)
...
+    // JL_TRY/JL_CATCH.
+    jl_value_t * module_value = jl_eval_string(name);
+    JULIAINTERFACE_EXCEPTION_HANDLER
+    if (!jl_is_module(module_value))
+        ErrorQuit("Not a module", 0, 0);
+    return (jl_module_t *)module_value;
+}
+
+Obj Func_InitializeGAPFFEType(){
+    JULIA_GAPFFE_type = (jl_datatype_t*)jl_new_datatype(jl_symbol("GapFFE"),
```
fingolfin: Why that cast for the return value? It's not needed, is it?

JuliaInterface/src/JuliaInterface.c
```diff
@@ -36,6 +35,43 @@ static void handle_jl_exception(void)
...
+                                                        get_module_from_string("GAP"),
+                                                        jl_any_type,
+                                                        jl_emptysvec,
+                                                        jl_svec(1,jl_symbol("ptr")),
+                                                        jl_svec(1,jl_voidpointer_type),
+                                                        0,0,1);
```
fingolfin: I though some more about this, and I now think it would be better to create a [primitive type](https://docs.julialang.org/en/v1/manual/types/#Primitive-Types-1) instead, using `jl_new_primitivetype` -- because what you are stuffing into instances of this type are bits which do *not* form valid `jl_value_t` pointers, but rather are pure bits.
So roughly this:
```C
jl_value_t *name = (jl_value_t*)jl_symbol("GapFFE");
jl_module_t *module = get_module_from_string("GAP");
JULIA_GAPFFE_type = jl_new_primitivetype(name, module, jl_any_type, jl_emptysvec, 8 * sizeof(Obj));
jl_set_const(module, name, JULIA_GAPFFE_type);
```

JuliaInterface/src/gap_macros.c
sebasguts: No, it just moves it all the way to the top.

sebasguts:
I adressed all the changes, and made GAPFFE a primitve type.

JuliaInterface/src/gap_macros.c
fingolfin: OK, but then why is this decl needed, considering that the function implementation now is moved to before the `#include "gap_macros.c"`?

JuliaInterface/src/gap_macros.c
```diff
@@ -23,6 +24,9 @@ Obj gap_julia(jl_value_t * julia_obj)
     if (IsGapObj(julia_obj)) {
         return (Obj)(julia_obj);
     }
+    if (jl_typeis(julia_obj,JULIA_GAPFFE_type)){
```
fingolfin: This is still not clang-formatted. I now made PR #110 to help us automatically track these things, *if* we can agree that's a good thing. If you rather not be forced to clang-format your commits, I can also shut up about it.

JuliaInterface/julia/gaptypes.jl
```diff
@@ -82,6 +82,13 @@ struct GapFunc
     ptr::MPtr
 end
 
+"""
+    GapFEE
+
+> Holds a pointer to a GAP FFE immediate object
```
fingolfin:
```suggestion
> Holds a GAP FFE immediate object.
```
Could also change "Holds" to "Wraps" or so?

JuliaInterface/julia/gaptypes.jl
```diff
@@ -82,6 +82,13 @@ struct GapFunc
...
+> (Gets defined in the JuliaInterface C code)
```
fingolfin:
```suggestion
> This type is defined in the JuliaInterface C code.
```

JuliaInterface/read.g
```diff
@@ -11,6 +11,8 @@ dirs:= DirectoriesPackageLibrary( "JuliaInterface", "julia" );
 
 JuliaIncludeFile( Filename( dirs, "gaptypes.jl" ) );
 
+_InitializeGAPFFEType();
```
fingolfin: Why not do that in `InitKernel`?

JuliaInterface/src/JuliaInterface.c
```diff
@@ -36,6 +35,43 @@ static void handle_jl_exception(void)
...
+    jl_value_t * module_value = jl_eval_string(name);
+    JULIAINTERFACE_EXCEPTION_HANDLER
+    if (!jl_is_module(module_value))
+        ErrorQuit("Not a module", 0, 0);
+    return (jl_module_t *)module_value;
+}
+
+Obj Func_InitializeGAPFFEType()
+{
+    JULIA_GAPFFE_type = (jl_datatype_t *)jl_new_primitivetype(
```
fingolfin:
```suggestion
    JULIA_GAPFFE_type = jl_new_primitivetype(
```
AFAICT there is no need to cast to `(jl_datatype_t *)` here, the return value of `jl_new_primitivetype ` already is `jl_datatype_t *`

JuliaInterface/src/JuliaInterface.c
```diff
@@ -488,19 +524,6 @@ jl_function_t * get_function_from_obj_or_string(Obj func)
...
-{
-    // It suffices to use JULIAINTERFACE_EXCEPTION_HANDLER here, as
-    // jl_eval_string is part of the jlapi, so don't have to be wrapped in
-    // JL_TRY/JL_CATCH.
-    jl_value_t * module_value = jl_eval_string(name);
-    JULIAINTERFACE_EXCEPTION_HANDLER
-    if (!jl_is_module(module_value))
-        ErrorQuit("Not a module", 0, 0);
-    return (jl_module_t *)module_value;
-}
```
fingolfin: Instead of moving this code around, you could also add a function declaration at the top of the file. Not that I care where it is, but moving functions complicates reviewing (e.g. now I did not check whether you made any change to the function other than moving it, because doing so requires too much work and I can't be bothered to invest that).

JuliaInterface/src/JuliaInterface.c
```diff
@@ -933,6 +956,7 @@ static StructGVarFunc GVarFuncs[] = {
     GVAR_FUNC(JuliaModule, 1, "name"),
     GVAR_FUNC(_ConvertedFromJulia_record_dict, 1, "dict"),
     GVAR_FUNC(_NewJuliaCFunc, 2, "ptr,arg_names"),
+    GVAR_FUNC(_InitializeGAPFFEType,0,""),
```
fingolfin: Hm, this also does not look clang-formatted

JuliaInterface/read.g
sebasguts: Because we need to have the GAP module first, which gets loaded into Julia after the package `.so` is loaded into GAP.

JuliaInterface/src/JuliaInterface.c
sebasguts: Okay, for next time. I think we will have to do some cleanup soon anyway.

JuliaInterface/src/gap_macros.c
sebasguts: Sorry, I just sometimes forget. I think the formatting is a really good thing and should be enforced!

JuliaInterface/read.g
fingolfin: Fair enough. It would be helpful if there was a comment somewhere explaining that, though...

PR #115, opened 2018-10-31
Add some LibGAP.jl tests and run them on Travis

### LINKED ISSUE #102: How can we add tests for the Julia side of `GAPJulia`?
I would like to add some tests to the Julia side of `GAPJulia`. The problem is that we need a path to an appropriate GAP executable to start GAP in Julia. So how can we add a portable test setup on the Julia side? Should we use environment variables?

fingolfin:
Resolves #102

### DIFF
--- a/LibGAP.jl/test/runtests_includes.jl
+++ b/LibGAP.jl/test/runtests_includes.jl
@@ -1,0 +1,6 @@
+
+@test LibGAP.CSTR_STRING(GAP.GAPFuncs.String(GAP.GAPFuncs.PROD(2^59,2^59))) == "332306998946228968225951765070086144"
+
+l = LibGAP.to_gap([1,2,3])
+
+@test l[1] == 1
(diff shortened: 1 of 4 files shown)


LibGAP.jl/test/runtests_includes.jl
```diff
@@ -0,0 +1,6 @@
+
+@test LibGAP.CSTR_STRING(GAP.GAPFuncs.String(GAP.GAPFuncs.PROD(2^59,2^59))) == "332306998946228968225951765070086144"
```
sebasguts: Bikeshed time: I think we should put such a test in a different file, say `basic.jl` or `convert.jl`. `runtests_includes.jl` should then just include those files.

The double indirection as opposed to the standard Julia testing is because we autogenerate `runtests.jl`

fingolfin: I propose something else: let's put the autogenerated data into a seprate file, say `gap_path.jl`, which then is included by a static, non generated `runtests.jl`

PR #117, opened 2018-11-01
Replace JuliaInterface/src/gap_macros.c by two new source files calls.c, convert.c

fingolfin:
These files are also compiled separately.

More conversion functions could be moved to convert.c, but I thought I'd better get this out ASAP, so we don't diverge too much.


sebasguts:
Can you clang-format it, so the tests pass? Or should I merge and then clang-format?

fingolfin:
Fixed.

PR #118, opened 2018-11-01
Get rid of LibGAP module

sebasguts:
This PR gets rid of the `LibGAP` module.


fingolfin:
Seems I forgot to clang format a header file...

PR #120, opened 2018-11-01
Compute coverage for LibGAP.jl, for realz this time

fingolfin:


fingolfin:
Should be fixed now, we'll see

PR #121, opened 2018-11-01
WIP: tweak to_gap; add some convenience helpers

fingolfin:
However, this code crashes GAP, hard... Specifially, the part
with the convenience helpers. Just reading this code in causes a crash.
Makes me wonder if we did something serious wrong in the GAP.GAPFuncs
implementations?!?

HELP! Why does it crash?

### DIFF
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -20,12 +20,6 @@ static jl_datatype_t * JULIA_GAPFFE_type;
 Obj  TheTypeJuliaObject;
 UInt T_JULIA_OBJ;
 
-Obj JI_SUM(Obj a, Obj b)
-{
-    return SUM(a, b);
-}
-
-
 void handle_jl_exception(void)
 {
     jl_call2(JULIA_FUNC_showerror, JULIA_ERROR_IOBuffer,
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -20,6 +20,7 @@ static jl_datatype_t * JULIA_GAPFFE_type;
 Obj  TheTypeJuliaObject;
 UInt T_JULIA_OBJ;
 
+
 void handle_jl_exception(void)
 {
     jl_call2(JULIA_FUNC_showerror, JULIA_ERROR_IOBuffer,
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -20,6 +20,11 @@ static jl_datatype_t * JULIA_GAPFFE_type;
 Obj  TheTypeJuliaObject;
 UInt T_JULIA_OBJ;
 
+Obj JI_SUM(Obj a, Obj b)
+{
+    return SUM(a, b);
+}
+
 
 void handle_jl_exception(void)
 {
--- a/LibGAP.jl/src/conversion.jl
+++ b/LibGAP.jl/src/conversion.jl
@@ -10,15 +10,8 @@ function to_gap(v :: Array{T, 1} where T) :: MPtr
     for i in 1:length(v)
         l[i] = v[i]
     end
-    # FIXME: size of l is never set, thus remains zero
     return l
 end
-
-# FIXME: this method makes no sense: it returns the value
-# computed by `map`, which will *not* be a valid GAP object
-#function to_gap(v :: AbstractArray)
-#    return map(to_gap, v)
-#end
 
 ## Conversion from GAP
 ##
--- a/LibGAP.jl/src/conversion.jl
+++ b/LibGAP.jl/src/conversion.jl
@@ -1,28 +1,33 @@
 ## Conversion to GAP
 
-to_gap(str :: String)         = MakeString(str)
-to_gap(v :: Any)              = v
-
-
-function to_gap( v :: Bool ) :: MPtr
-    if v
-        return True
-    else
-        return False
-    end
-end
+to_gap(v :: MPtr)       = v
+to_gap(v :: GapFFE)     = v
+to_gap(v :: String)     = MakeString(v)
+to_gap(v :: Bool)       = v ? True : False
 
 function to_gap(v :: Array{T, 1} where T) :: MPtr
     l = NewPlist(length(v))
     for i in 1:length(v)
         l[i] = v[i]
     end
+    # FIXME: size of l is never set, thus remains zero
     return l
 end
 
-function to_gap(v :: AbstractArray)
-    return map(to_gap, v)
-end
+# FIXME: this method makes no sense: it returns the value
+# computed by `map`, which will *not* be a valid GAP object
+#function to_gap(v :: AbstractArray)
+#    return map(to_gap, v)
+#end
+
+## Conversion from GAP
+##
+## FIXME: these are not very user friendly (user has to know type of the GAP
+## object in advanced), and some may even crash. It is not even clear that
+## here is the right place to perform most of these conversions, maybe they
+## should be done on the GAP and/or in the JuliaInterface GAP kernel
+## extension instead. We need to work this out...
+
 
 from_gap_string(obj :: MPtr) = CSTR_STRING(obj)
 
--- a/LibGAP.jl/src/gap.jl
+++ b/LibGAP.jl/src/gap.jl
@@ -57,19 +57,3 @@ for types in typecombinations
         end
     end
 end
-
-
-# #
-# +(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.SUM(x, y)
-# -(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.DIFF(x, y)
-# *(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.PROD(x, y)
-# /(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.QUO(x, y)
-# ^(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.POW(x, y)
-# mod(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.MOD(x, y)
-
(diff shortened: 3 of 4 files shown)


sebasguts:
@fingolfin With your convenience functions you overload arithmetic for Julia`s `Int64`, which subsequently causes Julia to crash (which is fair I guess).

So instead of adding all those functionality for `GAPInputType`, we need to be more specific. I will look into it.

fingolfin:
Ahh, indeed *facepalm*. Well, then, I guess one has to install the following combos:
  mptr/ffe/int  plus mptr/ffe
  mptr/ffe  plus mptr/ffe/int
So two union types are needed.

Feel free to rewrite this PR, or close it and make your own, borrowing whatever you might find useful in here.

sebasguts:
Last question: Do you think `QUO` should be `/` or `//`?

fingolfin:
`QUO` is tricky. If both inputs are ints, then `//` indeed *seems more appropriate. OTOH, for dividing group elements (permutations or matrices etc.), factor groups etc., it would seem annoying to have to use `//` instead of `/`.  

So, I think I'd actually use custom functions for both `/` and `//`, which both call `QUO`, but filter their inputs: `/` would do something different if both inputs are any kind of integers, otherwise call `QUO`. While `//` would only accept inputs which are integers or rationals (ideally, it'd allow mixing any GAP integer or GAP rationals with any Julia integer, including e.g. uint16 or bigint -- that's for the future, though)

But that's all quite complicated, which is why I chose to simply map `/` to `QUO` for starters and ignore the inconsistency -- we have much bigger fish to fry, like lots of crashes here and there, and missing tests (adding those tests tends to reveals those crashes...). I'd just do that, and then file an issue (and perhaps add a FIXME comment) on this.

But of course feel free to ignore that and go ahead and try to do it "right" (but please write a comment for people who want to understand what is going on :)

LibGAP.jl/src/gap.jl
```diff
@@ -37,16 +38,37 @@ Base.setproperty!(x::MPtr, f::Symbol, v) = GAP.GAPFuncs.ASS_REC(x, RNamObj(f), t
...
-#
-+(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.SUM(x, y)
--(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.DIFF(x, y)
-*(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.PROD(x, y)
-/(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.QUO(x, y)
-^(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.POW(x, y)
-mod(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.MOD(x, y)
+typecombinations = [[:GAPInputType_noint,:GAPInputType_noint],
+                    [:GAPInputType_noint,:Int64],
+                    [:Int64,:GAPInputType_noint]]
```
fingolfin: Warum reichen nicht zwei Combos: `:GAPInputType_noint, :GAPInputType]` und `[:GAPInputType, :GAPInputType_noint]`?

LibGAP.jl/src/gap.jl
```diff
@@ -3,7 +3,8 @@ import Base: convert, getindex, setindex!, length, show
 const True = GAP.GAPFuncs.ReturnTrue()
 const False = GAP.GAPFuncs.ReturnFalse()
 
-const GAPInputType = Union{MPtr,Int64,GAP.GapFFE}
+const GAPInputType_noint = Union{MPtr,GAP.GapFFE}
+const GAPInputType = Union{GAPInputType_noint,Int64}
```
fingolfin: Warum eigentlich `GAPInputType`? 

Vielleicht sollte dies der neue `GapObj`-Typ werden?

LibGAP.jl/src/gap.jl
```diff
@@ -37,16 +38,37 @@ Base.setproperty!(x::MPtr, f::Symbol, v) = GAP.GAPFuncs.ASS_REC(x, RNamObj(f), t
...
 
-<(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.LT(x, y)
-==(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.EQ(x, y)
+for types in typecombinations
+    for func in function_combinations
+        @eval begin
+            $(func[1])(x::$(types[1]),y::$(types[2])) = GAP.GAPFuncs.$(func[2])(x,y)
+        end
+    end
+end
```
fingolfin: Sehr schön!

In der Zukunft können wir natürlich noch viele mehr machen, z. B. `lcm`, `gcd`, `rem`, ... Aber da sollte man dann evtl. überlegen, ob das so sinnvoll ist.

Frage: wenn man `<` und `==` definiert, stellt dann Julia automatisch `>`, `<=`, `>=` bereit? Falls nicht, sollte man das noch tun (kann man aus `LT` basteln durch tauschen der Eingaben bzw. `!` verwenden).

JuliaInterface/src/JuliaInterface.c
```diff
@@ -20,6 +20,11 @@ static jl_datatype_t * JULIA_GAPFFE_type;
 Obj  TheTypeJuliaObject;
 UInt T_JULIA_OBJ;
 
+Obj JI_SUM(Obj a, Obj b)
+{
+    return SUM(a, b);
+}
```
fingolfin: Das hier kann natürlich weg, dass ist ein Überbleibsel meiner Versuche zu benchmarken, wie der Performance-Unterschied zwischen `ccall(:JI_SUM...)` und `GAP.GAPFuncs.SUM` ist. (Das sollte man natürlich trotzdem noch machen).

LibGAP.jl/src/gap.jl
```diff
@@ -27,9 +28,48 @@ Base.length(x::MPtr) = GAP.GAPFuncs.Length(x)
...
+# #
+# +(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.SUM(x, y)
+# -(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.DIFF(x, y)
+# *(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.PROD(x, y)
+# /(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.QUO(x, y)
+# ^(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.POW(x, y)
+# mod(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.MOD(x, y)
+
+# <(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.LT(x, y)
+# ==(x::GAPInputType, y::GAPInputType) = GAP.GAPFuncs.EQ(x, y)
```
fingolfin: Das kann jetzt natürlich weg.

sebasguts:
For now, I will leave it to `/`, as it is `/` in GAP.

Anyhow, I agree on the rest.

LibGAP.jl/src/gap.jl
sebasguts: Because this would give several (equal) options for the two types in `:GAPInputType_noint`. This way, we only have one possibility for each type combination

LibGAP.jl/src/gap.jl
sebasguts: Yes, Julia overloads it. But it still does not really work as intended, see
```
julia> x < 1
GAP: false

julia> x > 1
GAP: true
```
Maybe we should convert bools automatically?

LibGAP.jl/src/gap.jl
sebasguts: Indeed. We could use `GAPObj` as a name

sebasguts: But, I will introduce this in a second PR, as it will need several changes

LibGAP.jl/src/gap.jl
fingolfin: Are Julia true/false singletons? Then yes. Otherwise I'd just insert a call to from_gap_bool

fingolfin: Ah, I see, there are `jl_true` and `jl_false`. So yes, I'd simply always auto-convert these between GAP and Julia.

fingolfin: And sorry for my german comments, I had a braino there.

sebasguts: No worries. I will autoconvert them.

LibGAP.jl/src/conversion.jl
```diff
@@ -1,28 +1,33 @@
...
+to_gap(v :: GapFFE)     = v
+to_gap(v :: String)     = MakeString(v)
+to_gap(v :: Bool)       = v ? True : False
 
 function to_gap(v :: Array{T, 1} where T) :: MPtr
     l = NewPlist(length(v))
     for i in 1:length(v)
         l[i] = v[i]
     end
+    # FIXME: size of l is never set, thus remains zero
```
fingolfin: Actually, I just realized this complaint is wrong! Because you are using `l[i] = ...` defined below, which calls `ASS_LIST`, which *does* adjust the list size. So all should be fine. Of course best would be to verify this is accurate by adding a test which does something like `@test GAP.GAPFuncs.Length(to_gap([1,2,3])) == 3`.

LibGAP.jl/src/conversion.jl
```diff
@@ -1,28 +1,33 @@
...
 end
 
-function to_gap(v :: AbstractArray)
-    return map(to_gap, v)
-end
+# FIXME: this method makes no sense: it returns the value
+# computed by `map`, which will *not* be a valid GAP object
+#function to_gap(v :: AbstractArray)
+#    return map(to_gap, v)
+#end
```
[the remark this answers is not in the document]
sebasguts: I removed it. It was introduced fo match the conversion functionality from `JuliaInterface`, but probably is not that useful . . .

sebasguts:
I will merge after the checks are completed, and add some tests now.

PR #127, opened 2018-11-04
Use julia_gap and gap_julia in more places

fingolfin:
... for consistency in the auto-conversion behaviour of JuliaInterface versus
LibGAP.jl, and for ease of use.

There is more work to be done here, though, and we need to properly
document the rules for when and where which conversion is applied,
and why.

### DIFF
--- a/JuliaInterface/gap/JuliaInterface.gi
+++ b/JuliaInterface/gap/JuliaInterface.gi
@@ -28,6 +28,10 @@ InstallMethod( ConvertedFromJulia,
                [ IsJuliaObject ],
     _ConvertedFromJulia );
 
+InstallMethod( ConvertedFromJulia,
+               [ IsObject ],
+    IdFunc );
+
 InstallMethod( ConvertedToJulia,
                 [ IsObject ],
   function( obj )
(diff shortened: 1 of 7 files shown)


JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -28,6 +28,10 @@ InstallMethod( ConvertedFromJulia,
                [ IsJuliaObject ],
     _ConvertedFromJulia );
 
+InstallMethod( ConvertedFromJulia,
+               [ IsObject ],
+    ReturnFirst );
```
sebasguts: Nitpick: `IdFunc`

fingolfin:
My personal stance on this is this: This PR updates the existing tests for JuliaExperimental, so code covered by that is fine. All the other code in `JuliaExperimental` has a high chance of already being completely broken anyway. IMHO the only way to effectively move on with this project is that for all code, we either add comprehensive tests (then whoever makes changes has a chance of discovering anything they break, and subsequently can fix it), or else we remove that code eventually.

ThomasBreuer:
@sebasguts Yes, merging proposed changes in JuliaInterface is fine, independent of what they do with JuliaExperimental.
@fingolfin Adjusting tests in JuliaExperimental to changes in JuliaInterface is perhaps not recommended in general; the code is experimental, and it may well be that changes in JuliaInterface make bigger changes in JuliaExperimental sensible.

fingolfin:
Those test failures were real bugs, yay to the new tests that helped uncover them before merging!

@ThomasBreuer of course I am assuming that any adjustment to JuliaExperimental tests would involve some reasonable thinking, and questioning of what is there. Still, for tests to be most effective, it is important that they all pass (resp. that tests which are expected to fail are marked as such).

PR #132, opened 2018-11-05
Move arith.gi to JuliaInterface; add methods for EQ and LT, add some tests

fingolfin:
Contains PR #127


fingolfin:
I don't see an obstacle, but I am also not sure how much of an advantage there is. Would be useful to at least benchmark it, to get an idea. That said, I really think that all these convenience functions should not be used by code that wants highest performance: If you need to add a million Julia objects, don't write a GAP loop using GAP's wrapper around Julia's `+`; instead, write a Julia loop doing that and call it from GAP.

The advantage of keeping the code in GAP then is that it is very easy to debug and maintain there, and also to add new mappings. So, if I may make a recommendation, I'd leave `arith.gi` as it is for now, and at most add a TODO / an issue mentioning the possibility of replacing those by GAP kernel functions (installed into the various tables, like `EqFuncs` etc.), in the future. Then, let's focus on other things for now :).

sebasguts:
Agreed.

PR #133, opened 2018-11-05
Convert Julia bigints to GAP large ints

fingolfin:
The other conversion direction is missing because it is a bit fickly to correctly create Julia bigints from C; in particular, this code does it, taken from Julia's `src/dump.c`, but it uses `jl_gc_counted_malloc` which is not exported from the public Julia headers:
```C
    else if ((jl_value_t*)dt == jl_bigint_type) {
        jl_value_t *sizefield = jl_deserialize_value(s, NULL);
        int32_t sz = jl_unbox_int32(sizefield);
        int32_t nw = (sz == 0 ? 1 : (sz < 0 ? -sz : sz));
        size_t nb = nw * gmp_limb_size;
        void *buf = jl_gc_counted_malloc(nb);
        ios_read(s->s, (char*)buf, nb);
        jl_set_nth_field(v, 0, jl_box_int32(nw));
        jl_set_nth_field(v, 1, sizefield);
        jl_set_nth_field(v, 2, jl_box_voidpointer(buf));
    }
```
All in all, I think it might be better to write the GAP->Julia conversion in Julia, not in C, i.e., add another `BigInt` constructor which takes a GAP large int, and does the right thing. In any case, I won't solve this tonight.

### DIFF
--- a/JuliaInterface/src/JuliaInterface.c
+++ b/JuliaInterface/src/JuliaInterface.c
@@ -18,6 +18,8 @@ static jl_function_t * JULIA_FUNC_String_constructor;
 static jl_function_t * JULIA_FUNC_showerror;
 static jl_datatype_t * JULIA_GAPFFE_type;
 
+static jl_value_t * jl_bigint_type = NULL;
+
 Obj  TheTypeJuliaObject;
 UInt T_JULIA_OBJ;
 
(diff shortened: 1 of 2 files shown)


fingolfin:
Rebased

fingolfin:
To clarify: I think the conversion to Julia should work like this: add a new Julia function (as method for `big` and/or as a `BigInt` constructor) which takes as arguments an `Ptr{CVoid}` (or `Ptr{Limb}`) pointing to a buffer with the integer in it; a `Cint` giving the length of that buffer; and a boolean indicating the sign. The data is then *copied* into a new `BigInt`.

Once this is in place, we can simply call it from the kernel extensions with suitable arguments.

I won't work on this for now, though.

sebasguts:
@fingolfin I will try and write it.

PR #139, opened 2018-11-06
Move GAPFuncs to LibGap.jl and rename it to GAP.Globals

### LINKED ISSUE #123: Improve cache in getproperty for GAPFuncsType

fingolfin:
Resolves #123

### DIFF
--- a/JuliaInterface/src/convert.c
+++ b/JuliaInterface/src/convert.c
@@ -7,6 +7,9 @@
 // This function is used by LibGAP.jl
 jl_value_t * julia_gap(Obj obj)
 {
+    if (obj == 0) {
+        return jl_nothing;
+    }
     if (IS_INTOBJ(obj)) {
         return jl_box_int64(INT_INTOBJ(obj));
     }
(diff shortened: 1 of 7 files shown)


JuliaInterface/src/convert.c
```diff
@@ -7,6 +7,9 @@
 // This function is used by LibGAP.jl
 jl_value_t * julia_gap(Obj obj)
 {
+    if (obj == 0) {
```
sebasguts: Good catch.

fingolfin: Well, it crashed ... :-)

sebasguts: Picky computers. . .

PR #140, opened 2018-11-06
Tab completions for GAP globals

sebasguts:


LibGAP.jl/src/conversion.jl
```diff
@@ -39,11 +40,12 @@ end
 
 from_gap(x,::Type{Any})    = x
 from_gap(x,::Type{String}) = from_gap_string( x )
+from_gap(x,::Type{Symbol}) = from_gap_symbol( x )
```
fingolfin: Is there a deeper reason for having all these duplicate functions? I.e., any reason to have `from_gap_symbol`, doesn't `from_gap(x,::Type{Symbol})` suffice?

LibGAP.jl/src/conversion.jl
```diff
@@ -53,3 +55,5 @@ end
 from_gap_list_int16( x ) = from_gap_list_type( x, Int16 )
 from_gap_list_int32( x ) = from_gap_list_type( x, Int32 )
 from_gap_list_int64( x ) = from_gap_list_type( x, Int64 )
+from_gap_list_string( x ) = from_gap_list_type( x, String )
```
fingolfin: ... and this could perhaps become `from_gap(x,::Array{String})`, and so on? And then in fact, all of the above and including `from_gap_list_type` could be replaced by a single method `from_gap(x, ::Array{T} where T)`, no?

Not in this PR, mind you.

LibGAP.jl/src/conversion.jl
```diff
@@ -39,11 +40,12 @@ end
 
 from_gap(x,::Type{Any})    = x
 from_gap(x,::Type{String}) = from_gap_string( x )
+from_gap(x,::Type{Symbol}) = from_gap_symbol( x )
 from_gap(x,::Type{Bool})   = from_gap_bool( x )
 
 function from_gap_list_type( obj :: MPtr, element_type :: DataType ) 
     len_list = length(obj)
-    new_array = Array{element_type}(len_list)
+    new_array = Array{element_type,1}( undef, len_list)
```
fingolfin: So this was broken all the time?

I notice that `from_gap_list_type` has zero code coverage; would be great to have at least one minimal test for one of the various callers to it.

sebasguts:
Sorry for the missing tests. I intended to test the tab completion, but it is quite hard, as the list fluctuates with the given GAP version . . .

I will replace the `from_gap` things tomorrow morning.

fingolfin:
As for testing the tab completion: perhaps you could simply have a test like `@test "Print" in propertynames()` or so?

sebasguts:
Good idea. Lets see if this gets enough diff coverage (would guess so).

PR #144, opened 2018-11-07
Make gap error handler throw an error

### LINKED ISSUE #114: Julia side calling into GAP must intercept GAP errors, too
Just got this:
```
julia> l = LibGAP.to_gap([1,2,3])
GAP: [ 1, 2, 3 ]

julia> l[1]
1

julia> l[0]
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `[]' on 2 arguments

return;  # <- typed in by me

signal (11): Segmentation fault: 11
```

sebasguts:
Resolves #114

### DIFF
--- a/LibGAP.jl/src/initialization.jl
+++ b/LibGAP.jl/src/initialization.jl
@@ -25,7 +25,7 @@ read_sysinfo_gap = function(dir::String)
 end
 
 function error_handler(message)
-    print(message)
+    error("Error thrown by GAP")
 end
 
 error_handler_func = @cfunction(error_handler,Cvoid,(Ptr{Char},))
--- a/LibGAP.jl/src/initialization.jl
+++ b/LibGAP.jl/src/initialization.jl
@@ -24,11 +24,11 @@ read_sysinfo_gap = function(dir::String)
     return d
 end
 
-function error_handler(message)
+function error_handler()
     error("Error thrown by GAP")
 end
 
-error_handler_func = @cfunction(error_handler,Cvoid,(Ptr{Char},))
+error_handler_func = @cfunction(error_handler,Cvoid,())
 
 const pkgdir = realpath(dirname(@__FILE__))
 
--- a/LibGAP.jl/test/basics.jl
+++ b/LibGAP.jl/test/basics.jl
@@ -16,4 +16,7 @@
     @test GAP.True
     @test ! GAP.False
 
+
+    xx = GAP.to_gap([1,2,3])
+    @test_throws ErrorException xx[4]
 end


fingolfin:
But #112 is another PR, not an issue. Perhaps #114 was meant?

LibGAP.jl/src/initialization.jl
```diff
@@ -25,7 +25,7 @@ read_sysinfo_gap = function(dir::String)
 end
 
 function error_handler(message)
-    print(message)
+    error("Error thrown by GAP")
```
fingolfin: Could this perhaps include the message?

sebasguts: I will check how to achieve this. It seemed to me that the message gets printed by GAP anyway.

sebasguts:
The complicated solution should work this way as well: GAP catches Julia error and goes into the GAP error, and Julia catches the GAP error and goes into Julia error. I think this works in arbitrary recursion depth.

LibGAP.jl/src/initialization.jl
fingolfin: Oh, I see: I thought this is just a mater of printing `message`. But now looking at how the "error callback" is called, I realized that it actually gets no arguments -- and this signature here is incorrect.

fingolfin:
I am not so sure that this code would work right with arbitrary interleaved calls; indeed, I even think that it's not yet enough to support some GAP code from within Julia without any mutual recursion at all:. Because we are actually setting the `JumpToCatchCallback` in GAP, and that is used unconditionally; thus GAP code which uses CALL_WITH_CATCH for other purposes will not work, as those will be intercepted just like errors. This is a major flaw in the current libgap API that needs to be reported as an issue and be resolved. And doing that is non-trivial once you allow crisscrossing between GAP and Julia code; right now, we always go back in one swoop to the julia code; we skip any intermediate GAP error handling, which means that we may miss out on resource cleanup, or alternate error recovery strategies. The same holds for the other direction, too, I guess.

I am sorry for being a bit fuzzy, but I don't have time right now to work out exact examples and give more details; the whole "error handling" in GAP is a bit of a mess right now for various reasons, and also extremely complicated; I've been working on untangling it a bit for the past year, but its still far from me how to fix it in GAP alone, and I am far from being clear on how it should work when adding Julia as another player. (And once we also want to integrate it with Singular and Polymake, it'll be even more challenging; but that's actually a major point for using Julia as the central glue for all these systems, because this way, at least we only need to solve this very complicated domain (error handling between two interfaced systems) once for every pair, and not for all pairs.

sebasguts:
@fingolfin I rebased this commit and fixed the function signature.

For now, I gave up on the stream business.

We could for example create a GAP stream, and set `ERROR_OUTPUT` in GAP to it, then get the string from that stream. But this only seems like a half-baked solution, so I did not do it yet.

I would prefer to be able to give `GAP_Initiliaze` a file descriptor which it sets as GAP's `*stderr*`, but I was not yet able to find out how to do this.

PR #145, opened 2018-11-07
Changed conversion, added tests

sebasguts:


LibGAP.jl/src/conversion.jl
```diff
@@ -22,38 +24,15 @@ end
 ## extension instead. We need to work this out...
 
 
-from_gap_string(obj :: MPtr) = CSTR_STRING(obj)
-from_gap_symbol(obj :: MPtr) = Symbol(from_gap_string(obj))
+from_gap(obj        , ::Any)                  = obj
+from_gap(obj :: MPtr, ::Type{AbstractString}) = CSTR_STRING(obj)
```
fingolfin: As per the comment above, this (still) does not perform any input validation and will crash or produce garbage if called on a non-string GAP object. Something to fix in a future PR, I guess.

sebasguts:
Agreed. I will provide a PR which uses the C functions instead.

Issue #146, opened 2018-11-07
Share conversion code between LibGAP.jl and JuliaInterface


fingolfin:
Right now, many conversions are implemented twice: once in the GAP C kernel extension, see e.g. `_ConvertedFromJulia_internal` and `_ConvertedToJulia_internal`; and once in Julia, see `from_gap` and `to_gap` in `LibGAP.jl/src/conversion.jl`.

There are multiple problems with that:
* we have to duplicate a lot of work
* which conversions resp. types are supported can and will differ
* error checking differs
* one side may support other types than the other

Therefore, we should think about ways that allow us to implement all these conversions once for both sides. Whether this is done in C, GAP or Julia, or a mix of them does not matter that much to me, as long as it is efficient and available from both sides, and flexible enough for our purposes.

PR #147, opened 2018-11-08
Use JuliaInterface conversion methods in LibGAP.jl

sebasguts:
See #146

### DIFF
--- a/LibGAP.jl/src/ccalls.jl
+++ b/LibGAP.jl/src/ccalls.jl
@@ -85,6 +85,20 @@ function propertynames(funcobj::GlobalsType,private)
     return tuple(list_converted...)
 end
 
+function to_gap(obj::Any)
+    ret_ptr = ccall(:_ConvertedFromJulia_internal,Ptr{Cvoid},(Any,),obj)
+    ret_val = GET_FROM_GAP(ret_ptr)
+    return ret_val
+end
+
+
+function from_gap(obj::MPtr,::Any)
+    ret_val = ccall(:_ConvertedToJulia_internal,Any,(MPtr,),obj)
+    return ret_val
+end
+
+
+
 # For backwards compatibility
 # TODO: remove this again
 GAPFuncs = Globals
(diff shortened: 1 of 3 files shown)


LibGAP.jl/src/ccalls.jl
```diff
@@ -85,6 +85,20 @@ function propertynames(funcobj::GlobalsType,private)
...
 
+function to_gap(obj::Any)
+    ret_ptr = ccall(:_ConvertedFromJulia_internal,Ptr{Cvoid},(Any,),obj)
+    ret_val = GET_FROM_GAP(ret_ptr)
+    return ret_val
+end
+
+
+function from_gap(obj::MPtr,::Any)
+    ret_val = ccall(:_ConvertedToJulia_internal,Any,(MPtr,),obj)
```
fingolfin: So I guess the idea is to have this two methods as kind of the "default" fallback methods, and implement any custom user driven conversions in Julia code?

That is already an improvement, but of course still asymmetric. E.g. it would now be natural to implement the conversion of GAP bigints to Julia bigints in Julia; but then it would not be accessible in JuliaInterface... so I guess we'd also need to change `JuliaInterface` to call `to_gap` / `from_gap` in certain cases (but of course from a higher level function, not from `_ConvertedFromJulia_internal`, to avoid infinite recursion)

LibGAP.jl/src/conversion.jl
```diff
@@ -1,32 +1,29 @@
-## Conversion to GAP
 
-to_gap(v :: Int64)      = v
-to_gap(v :: MPtr)       = v
-to_gap(v :: GapFFE)     = v
-to_gap(v :: String)     = MakeString(v)
-to_gap(v :: Symbol)     = to_gap(string(v))
-to_gap(v :: Bool)       = v
+to_gap(x::Int64) = x
+to_gap(x::GapFFE) = x
```
fingolfin: Maybe keep `to_gap(x::MPtr)`?

LibGAP.jl/src/ccalls.jl
sebasguts: Yep, indeed. I think the ultimate goal needs to be exactly one conversion function for each type pair in GAP/Julia, either written on Julia side, so no code is actually duplicated.

PR #148, opened 2018-11-09
Add test for LibGAP.jl, fix bug in Base.string method

fingolfin:


LibGAP.jl/src/gap.jl
```diff
@@ -13,7 +13,7 @@ function Base.show( io::IO, obj::Union{MPtr,GAP.GapFFE} )
 end
 
 function Base.string( obj::Union{MPtr,GAP.GapFFE} )
-    str = GAP.String( obj )
+    str = GAP.Globals.String( obj )
     return CSTR_STRING( str )
```
sebasguts: Nitpick: Shouldn't we use `from_gap` here?

fingolfin: I think it's fine either way, but I'd prefer to merge it as-is, with minimal changes. But feel free to edit this PR, or change it in another PR.

sebasguts: It can be changed later, I guess.

PR #149, opened 2018-11-13
Remove GAPRat code

### LINKED ISSUE #71: JuliaExperimental: incorrect use of AbsInt and GcdInt; create_rational is dangerous

fingolfin:
This code is not needed anymore due the existence of the convenience wrappers
for arithmetic operations on GAP objects in Julia, resp. on Julia objects in
GAP.

Moreover, this code was buggy; and was not yet converted to use
`ForeignGAP.MPtr` (leading to GC crash). All in all, it seems best to
"conclude" this experiment by removing it.

Resolves #71


fingolfin:
@ThomasBreuer I think this is your code, isn't it? So I think you should say whether you agree with removing it, or whether you'd prefer to work on updating it.

PR #157, opened 2018-11-14
added first tests using Singular.jl

ThomasBreuer:


JuliaExperimental/gap/singular.g
```diff
@@ -497,7 +501,7 @@ InstallOtherMethod( \*,
                  Julia.Base.("*")( x, JuliaPointer( y ) ) );
     end );
 
-#T Calling 'divexact' causes a segmentation fault in Julia!
+#T Calling 'divexact' directly causes a segmentation fault in Julia!
```
fingolfin: Could you please clarify this? I.e., what invocation exactly crashes? Perhaps insert it as part of the comment?

JuliaExperimental/gap/singular.g
```diff
@@ -295,15 +293,17 @@ InstallOtherMethod( GcdOp,
...
     end );
 
 
 InstallOtherMethod( GcdOp,
     IsIdenticalObj,
     [ "IsSingularPolynomial", "IsSingularPolynomial" ],
     function( f, g )
-        return SingularElement( f, Julia.Base.gcd( f, g ) );
+        return SingularElement( f, Julia.Base.gcd(
+                   JuliaPointer( f ), JuliaPointer( g ) ) );
```
fingolfin: @sebasguts and me just discussed an extension to the conversion code that would in effect automatically call `JuliaPointer` to "unwrap" high-level wrapper objects, as long as those high-level objects are in a suitable filter (say, `IsAutomaticUnwrappingJuliaWrapper` or so...). Then the above change could be reverted to the old version.

JuliaExperimental/gap/singular.g
ThomasBreuer: @fingolfin Thanks for the hint.
In fact, now (with a new Julia and a new Singular) the "natural" way works.

JuliaExperimental/gap/singular.g
ThomasBreuer: Should this behavior (automatically unwrapping or not) really belong on an object,
or does it depend on the situation whether one wants to unwrap or not?

fingolfin: We have been discussing this as part of how to deal with extended automatic conversions of arguments for method calls (as opposed to just performing the absolute necessary conversions). We think we have a good plan for that now, and are working on implementing and writing up more about this. In the meantime, this is nothing that should delay this PR.

PR #160, opened 2018-11-14
Work on conversion

### LINKED ISSUE #116: Implement conversion between GAP large integers and Julia BigInt

### LINKED ISSUE #146: Share conversion code between LibGAP.jl and JuliaInterface

### LINKED ISSUE #162: Deal with Julia rationals 1//0 and -1//0
By chance I just learned that Julia allows rationals with denominator `0`, which are then normalized to `1//0` and `-1//0`. For these, `isinf` returns `true`, and they seem to behave like infinities.

I propose to convert these to GAP's `-infinity` and `+infinity`.

sebasguts:
For a detailed description on this work, see https://hackmd.io/-xkD8BtORfCcgi-TQsb8Eg?both

Resolves #162 
Resolves #146
Resolves #116


LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -23,8 +23,17 @@ julia_to_gap(x::BigInt)  = MakeObjInt(x)
...
 function julia_to_gap(x::Rational{T}) where T <: Integer
-    numer = julia_to_gap(numerator(x))
-    denom = julia_to_gap(denominator(x))
+    denom_julia = denominator(x)
+    numer_julia = numerator(x)
+    if denom_julia == 0
+        if numer_julia >= 0
+            return GAP.Globals.infinity
+        else
+            return -GAP.Globals.infinity
```
fingolfin: Perhaps add `Resolves #162` to the relevant commit message?

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -0,0 +1,167 @@
...
+    ## Check for correct type
+    if ! Globals.IsInt(x)
+        throw(ArgumentError("GAP object is not a large integer"))
+    end
+    ## get size of GAP BigInt (in limbs), multiply
+    ## by 64 to get bits
+    size_limbs = ccall(:GAP_SizeInt,Cint,(MPtr,),x)
+    size = abs(size_limbs * sizeof(UInt) * 8)
+    ## allocate new GMP
+    new_bigint = Base.GMP.MPZ.realloc2(size)
```
fingolfin: Also add "Resolves #116` to the relevant commit message?

LibGAP.jl/src/julia_to_gap.jl
sebasguts: I thought about it, and I would rather make a comment here in the PR, because I think we wanted to keep the commit messages clear from the `Closes #whatever` messages?

fingolfin: I certainly was never part of this "we" you mention ;-). But I really don't mind either way in this case, as long as the issues get referenced and closed

sebasguts: Maybe it was not "we", but "me", I am not sure anymore ;)

sebasguts:
@fingolfin @ThomasBreuer @frankluebeck I think this is finished for now. Sorry for the mess. Please have a look and review it.

@ThomasBreuer JuliaExperimental is not completely adjusted yet.

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -0,0 +1,99 @@
...
+julia_to_gap(x::UInt16, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive = Int64(x)
+julia_to_gap(x::UInt8, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive  = Int64(x)
+
+## BigInts
+julia_to_gap(x::BigInt, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive = MakeObjInt(x)
+
+## Rationals
+function julia_to_gap(x::Rational{T}, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive where T <: Integer
+    denom_julia = denominator(x)
+    numer_julia = numerator(x)
```
fingolfin: According to Codecov, there are no tests for this function at all, it is never called :-(

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -0,0 +1,99 @@
+## Converters
+
+## Default
+julia_to_gap(x::GAPInputType, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive = x
```
fingolfin: Please add a longer comment to this file which explains what is going on. Like: What is this `Val{Recursive}` bit doing; but also the general idea of what kind of methods, with what signature, one has to achieve what.

I actually need that explanation, as I don't understand right now why this complicated setup is better than this simpler one: For all "basic" types (= not containing subobjects that need to be converted), have methods like this:
```julia
julia_to_gap(x::Int32, recursion_dict) = Int64(x)
```
Then, have one generic redispatch function which handles all these primitive ones at once; and for all data types with subobjects, have both 1 and 2 arg methods:
```julia
julia_to_gap(x::Any) = julia_to_gap(x, nothing)

function julia_to_gap(obj::Array{T,1}, recursion_dict = IdDict()) where T
...
end
```

Perhaps the code the above produces is suboptimal? Well, I wouldn't ask this question if there was a comment explaining the design of the current approach :-)

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -0,0 +1,99 @@
...
+    size = length(obj)
+    array = Array{Any,1}(undef,size)
+    for i in 1:size
+        array[i] = obj[i]
+    end
+    return julia_to_gap(array, recursive, recursion_dict)
+end
+
+## Dictionaries
+function julia_to_gap(obj::Dict{T,S}, recursive::Val{Recursive}=Val(false), recursion_dict = IdDict()) where Recursive where S where T <: Union{Symbol,AbstractString}
```
fingolfin: According to Codecov, there are no tests for this function at all, it is never called :-(

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -0,0 +1,99 @@
...
+            end
+            recursion_dict[current_obj] = current_converted
+            ret_val[i] = current_converted
+        end
+    end
+    return ret_val
+end
+
+## Tuples
+function julia_to_gap(obj::Tuple, recursive::Val{Recursive}=Val(false), recursion_dict = IdDict()) where Recursive
```
fingolfin: According to Codecov, there are no tests for this function at all, it is never called :-(

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -0,0 +1,186 @@
...
+gap_to_julia(::Type{UInt128},x::Int64, recusive_dict = nothing) = trunc(UInt128,x)
+gap_to_julia(::Type{UInt64} ,x::Int64, recusive_dict = nothing) = trunc(UInt64 ,x)
+gap_to_julia(::Type{UInt32} ,x::Int64, recusive_dict = nothing) = trunc(UInt32 ,x)
+gap_to_julia(::Type{UInt16} ,x::Int64, recusive_dict = nothing) = trunc(UInt16 ,x)
+gap_to_julia(::Type{UInt8}  ,x::Int64, recusive_dict = nothing) = trunc(UInt8  ,x)
+
+## BigInts
+gap_to_julia(::Type{BigInt}, x::Int64, recusive_dict = nothing) = BigInt( x )
+
+function gap_to_julia(::Type{BigInt}, x::MPtr, recusive_dict = nothing)
```
fingolfin: According to Codecov, there are no tests for this function at all, it is never called :-(

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -0,0 +1,186 @@
...
+    return new_bigint
+end
+
+## Rationals
+function gap_to_julia(::Type{Rational{T}}, x::Int64, recusive_dict = nothing) where T <: Integer
+    numerator = gap_to_julia(T,x)
+    return numerator // T(1)
+end
+
+function gap_to_julia(::Type{Rational{T}}, x::MPtr, recusive_dict = nothing) where T <: Integer
```
fingolfin: According to Codecov, there are no tests for this function at all, it is never called :-( Same for the other rationals conversion above, and the float conversion below, and more -- see also https://codecov.io/gh/oscar-system/GAPJulia/pull/160/diff

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -0,0 +1,99 @@
...
+## BigInts
+julia_to_gap(x::BigInt, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive = MakeObjInt(x)
+
+## Rationals
+function julia_to_gap(x::Rational{T}, ::Val{Recursive}=Val(false), recursion_dict = nothing) where Recursive where T <: Integer
+    denom_julia = denominator(x)
+    numer_julia = numerator(x)
+    if denom_julia == 0
+        if numer_julia >= 0
+            return GAP.Globals.infinity
```
fingolfin: Should there perhaps also be a reverse conversion for this? But of course that's difficult, as on the GAP side, `infinity` and `-infinity` are objects, and "only" in the filter `IsCyc`, not in `IsRat`. Huh :/

ThomasBreuer: @fingolfin As far as I remember,
the object `infinity` had originally been designed just for having a natural object
that is larger than any integer, and that can be compared with integers via `=` and `<`.
In particular, there were no arithmetic operations for `infinity`,
and also `-infinity` did not yet exist.
Because of the wish to compare the new object `infinity` with integers,
the family `CyclotomicsFamily` was a natural choice for it,
and thus `infinity` is in the filter `IsCyclotomic` --but not in `IsCyc`, `IsRat`, or `IsInt`.

Before the introduction of `infinity`, other GAP objects of different types had been used
for this kind of comparison, for example `E(3)`,
assuming that GAP regards them as larger than integers.
Using `infinity` in such situations is nicer because it is self-explanatory;
besides that, I do not think that `infinity` is useful.

LibGAP.jl/src/gap_to_julia.jl
sebasguts: I think the codecov output is wrong. The test is in `conversion.jl`, line 22, and if I remove this function, the tests fail.

LibGAP.jl/src/julia_to_gap.jl
fingolfin: I am not sure how this is relevant for the discussion at hand, though?

BTW, I added `-infinity`, and I find it and `infinity` quite useful for all kinds of things, e.g. for polynomial degree functions, or arbitrary valuations; as default values for minimum / maximum operations; etc.. But whether one finds them useful or not is kind of irrelevant for this discussion. What we have to deal with is issue #162. We can either throw an error if we encounter the Julia rationals 1//0 and -1//0; or convert them to infinity and -infinity (as this PR does). In the latter case, the question arises whether we want to make the conversion roundtrip or not. While in general, I am in favor for clean roundtrip conversions, but is of course not always possible; but then perhaps there should be a comment here pointing that out.

LibGAP.jl/src/gap_to_julia.jl
fingolfin: The problem was that JuliaExperimental tests failed, which caused Travis to stop all further tests, so the LibGAP.jl tests never did run.

LibGAP.jl/src/gap_to_julia.jl
frankluebeck: (side remark: is `recusive_dict` a typo?)

fingolfin:
@frankluebeck "I get an error when trying to load the package" -> could you perhaps tell us what that error is?

frankluebeck:
I was loading "JuliaExperimental" which is maybe not expected to work with this pull request. Loading "JuliaInterface" only does work.

So, now I get:
```
gap> s:="";;
gap> for i in [0..255] do Add(s,CharInt(i));od;
gap> js:=GAPToJulia(s);
<Julia: "">
gap> sjs:=JuliaToGAP(IsString,js);
""
gap> Length(sjs);
0
gap> GAPToJulia([1,,2]);
Error, List Element: <list>[2] must have an assigned value
not in any function at *stdin*:19
you can 'return;' after assigning a value
brk> 
```

fingolfin:
1. While `String` is indeed restricted to UTF-8, `AbstractString` explicitly is meant to support arbitrary encodings. But of course we actually create `String` instances, so I guess we might need to use or introduce a custom `RawString` or `GAPString class instead. Or perhaps somebody has a better idea?

2. Automatic conversions do *not* convert bigints. For the explicit manual conversions, we do indeed provide a generic conversion between GAP large ints and Julia big ints. For non-recursive conversions, I think that's fine, as you can simply invoke another conversion function if you want, say, FLINT bigints instead. But I guess if you have a list of GAP large ints, it'd be nice if one could quickly convert that to a list of FLINT bigints, too... And in a sense, you can, using functions like `map`. But I agree this needs some more thought.

3. Right now, lists with holes are not supported, good point. We could of course reject them; but we could also create lists containing `nothing` in place of the hole, which would be nicely consistent with `julia_gap` mapping NULL pointers to `nothing`.

4. Mutability is not yet addressed at all, neither in automatic nor in manual/explicit conversions. There is likely no perfect solution for this either; e.g. Julia's `AbstractString` is *always* immutable, just like Java strings.

fingolfin:
Yes, `JuliaExperimental` is currently broken

LibGAP.jl/src/julia_to_gap.jl
sebasguts: I am not sure about the code being less optimal that way. I will try to implement the changes you propose.

LibGAP.jl/src/julia_to_gap.jl
sebasguts: I think it is sensible to convert to `infinity` here.

Remember, we are talking about a very specific corner-case here, Julia rationals with `0` as denominator. I do not think it is that important anyway.

sebasguts:
1. There is a `Cstring` class in Julia, which is composed of `Cchars`. It is basically meant to be compatible with `char*`. The other way around is a problem, though. Julia strings are by default `String`s, which are UTF-8 encoded. If we want to not have any pitfalls here, we would probably need a custom GAP object. If we want to convert easily, I think `AbstractString` is the right thing to do.

3. This is merely a problem about how we convert lists: We use GAP's `ELM_LIST` function, which errors if the element is not set. We could use a function that may return `nothing` instead. But note that this would make the conversion different. If you want to convert to an `Array{Int64}`, we would need to set the returning type to `Array{Union{Int64,Nothing}}` - So either one must know before that the list might have holes, or we have to return a different type than what we specified.

4. Indeed, Julia does not have mutable/immutable variants of all objects. Do we really want to preserve mutability all the time?

PR #163, opened 2018-11-22
Make GAP.EvalString more convenient

fingolfin:
Rename the current GAP.EvalString to GAP.EvalStringEx (for "Ex"tended), added
a new GAP.EvalString which only returns the value of the last command, and
which also automatically appends a semicolon to the input string.

### DIFF
--- a/LibGAP.jl/src/ccalls.jl
+++ b/LibGAP.jl/src/ccalls.jl
@@ -6,11 +6,16 @@ function GET_FROM_GAP(ptr::Ptr{Cvoid})::Any
     return ccall(:julia_gap,Any,(Ptr{Cvoid},),ptr)
 end
 
-function EvalString( cmd :: String )
+function EvalStringEx( cmd :: String )
     res = ccall( :GAP_EvalString, MPtr, 
                  (Ptr{UInt8},),
                  cmd );
     return res
+end
+
+function EvalString( cmd :: String )
+    res = EvalStringEx(cmd * ";")
+    return res[end][2]
 end
 
 function ValueGlobalVariable( name :: String )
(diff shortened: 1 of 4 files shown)


LibGAP.jl/src/ccalls.jl
```diff
@@ -6,13 +6,18 @@ function GET_FROM_GAP(ptr::Ptr{Cvoid})::Any
...
-function EvalString( cmd :: String )
+function EvalStringEx( cmd :: String )
     res = ccall( :GAP_EvalString, MPtr, 
                  (Ptr{UInt8},),
                  cmd );
     return res
 end
 
+function EvalString( cmd :: String )
+    res = EvalStringEx(cmd * ";")
```
sebasguts: Are you sure this is an sensible convention?

I am not sure it makes sense here, but in GAP you can end statements with `;` or `;;`. If you now put `;;` at the end of a statement, this command does not return a value. Is that intentional?

fingolfin: Yes, quite intentional:
* if you want to shoot yourself in the foot, you can always do that, and e.g. end the string with a command that returns nothing; or add `;;;;;` to the end (which amounts more or less the same); or do something else that's crazy. You can do so with the existing `EvalString` command do
* I though about doing something like "check if there is a trailing semicolon and only add one if there is not" -- but then you can continue this game, and ask "what if the user put three semicolons at the end, should I remove some" etc.
* Adding a semicolon is also what GAP's own `EvalString` does, so now the two match in their behavior.

sebasguts: Yeah, I agree that more or less all solutions have pitfalls. If GAP's EvalString method behaves like this, it is okay. Stupid question, why don't we make `EvalString` just
```
EvalString( x ) = GAP.Globals.EvalString( julia_to_gap( x ) )
```
This way we can have the *exact* same behaviour?

PR #170, opened 2018-11-23
adjusted JuliaExperimental to the new conversion rules

ThomasBreuer:
- adjusted the function calls to the new rules (from pull request #160),
  removed or simplified the relevant code
- removed `gap/record.g` since the code is now obsolete;
  moved the contents of `tst/record.tst` to `JuliaInterface/tst/convert.tst`
- changed the auxiliary dictionaries to use symbols as keys,
  in order to use the default conversions
- adjusted some Julia code which worked only in pre-1.0 versions

After these changes, the JuliaExperimental tests should work,
thus the automatic test runs can be reactivated.

(Currently I am rewriting the whole package JuliaExperimental,
a pull request for that will be available on Monday.)


JuliaExperimental/gap/gapperm.g
```diff
@@ -79,23 +100,23 @@ InstallMethod( \*, [ IsExtPerm, IsExtPerm ], function( p1, p2 )
 
 InstallMethod( \^, [ IsExtPerm, IsInt ], function( p1, n )
     return WrappedPermutationInJulia( Julia.Base.("^")(
-               p1![1], ConvertedToJulia( n ) ) );
+               p1![1], GAPToJulia( n ) ) );
     end );
 
 InstallMethod( \^, [ IsInt, IsExtPerm ], function( i, p )
-    return ConvertedFromJulia( Julia.Base.("^")( ConvertedToJulia( i ), p![1] ) );
+    return JuliaToGAP( IsInt, Julia.Base.("^")( GAPToJulia( i ), p![1] ) );
```
fingolfin: By the way (not for this PR, but perhaps a future update of this code): you can write `Base.\^`, and the resulting code will be slight more efficient, as it will directly code the right rnam, instead of creating a temporary string and then turning it into an rnam via a hash table lookup in the GAP kernel. For bigger computations, all these temporary objects can add up to a significant extra load.

ThomasBreuer: @fingolfin Thanks for this hint.
This remark should become part of the documentation,
once there is a place for it.

PR #172, opened 2018-11-23
New array uint8 conversion

sebasguts:


LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -101,8 +101,9 @@ gap_to_julia(type_obj,obj,recursion_dict) = gap_to_julia(type_obj,obj)
 
 ## Convert GAP string to Array{UInt8,1}
 function gap_to_julia( ::Type{Array{UInt8,1}}, obj :: MPtr )
+    ## convert strings to uint8 lists, if wished for
```
fingolfin:
```suggestion
    ## convert strings to uint8 lists, if requested
```
I find it a bit odd that this comes before the case that handles non-strings. Perhaps reverse the logic, i.e, `if IsStringRep ... else ` ?

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -210,7 +214,7 @@ function gap_to_julia(x::MPtr)
     elseif Globals.IsString(x)
         return gap_to_julia(AbstractString,x)
     elseif Globals.IsList(x)
-        return gap_to_julia(Array{Any,1},x)
+        return gap_to_julia(Array{Union{Any,Nothing},1},x)
```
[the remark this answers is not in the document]
sebasguts: This seemed to be slipped in, I will remove this. It should be part of #173

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -101,11 +101,15 @@ gap_to_julia(type_obj,obj,recursion_dict) = gap_to_julia(type_obj,obj)
 
 ## Convert GAP string to Array{UInt8,1}
 function gap_to_julia( ::Type{Array{UInt8,1}}, obj :: MPtr )
-    if ! Globals.IsStringRep( obj )
-        return Array{UInt8,1}(gap_to_julia(Array{UInt16,1},obj))
+    ## convert strings to uint8 lists, if requested
+    if Globals.IsStringRep( obj )
+        array = UNSAFE_CSTR_STRING( obj )
+        return deepcopy(array)
```
fingolfin: Man, this code formatting really is all over the place :/. I really wish we had something like `clang-format` for Julia code.

PR #173, opened 2018-11-23
Make holes in GAP list convert to `nothing` and vice versa.

sebasguts:


LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -121,7 +122,7 @@ function gap_to_julia( ::Type{Array{Obj,1}}, obj :: MPtr , recursion_dict = IdDi
     new_array = Array{Any,1}( undef, len_list)
     recursion_dict[obj] = new_array
     for i in 1:len_list
-        current_obj = obj[i]
+        current_obj = ElmList(obj,i)
```
fingolfin: Hm. Should we then (or even instead) also adjust `GAP.Globals.ELM_LIST` to use `ElmList` (for performance, and to also return `nothing`.

And perhaps we should also use `GAP_AssList` for `Base.setindex!` instead of `GAP.Globals.ASS_LIST`>

LibGAP.jl/test/conversion.jl
```diff
@@ -104,6 +104,15 @@
     xx = GAP.julia_to_gap( "a" )
     @test_throws MethodError GAP.gap_to_julia( Dict{Int64,Int64}, xx )
 
+    ##Test nothing
```
fingolfin:
```suggestion
    ## Test nothing
```

LibGAP.jl/test/conversion.jl
```diff
@@ -182,4 +191,8 @@ end
     xx = GAP.EvalString("[~]");
     @test xx === xx[1]
 
+    ##Test nothing
```
fingolfin:
```suggestion
    ## Test nothing
```

LibGAP.jl/src/gap_to_julia.jl
sebasguts: I am not completely sure about this. `ELM_LIST` behaves different, as it returns `fail` instead of `nothing`. We currently do not map `fail` to `nothing`, which is probably the right behaviour. I would try to mimic GAP's behaviour here, which is the behaviour of `ELM_LIST` (or even throwing an error).

fingolfin: First off, sorry, I meant to write: "... should we adjust `Base.getindex` to use `ElmLst` instead of `GAP.Globals.ELM_LIST` ..."

Secondly, I am not aware of any variant of `ELM_LIST` returning `fail`, what are you referring to exactly? I am seeing this:
```
julia> xx = GAP.EvalString("[1,,3]")
GAP: [ 1, , 3 ]

julia> GAP.Globals.ELM_LIST(xx,2)
Error, List Element: <list>[2] must have an assigned value
ERROR: Error thrown by GAP
```

fingolfin: I think it's fine to have an error here. What is less clear to me is whether we want to rely on GAP throwing that error. We could instead write a slightly more elaborate getter, which uses `ElmList`, and then if `nothing` is returned, throws an error.

Anyway, this is not a blocker for this PR.

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -143,7 +144,30 @@ function gap_to_julia( ::Type{Array{T,1}}, obj :: MPtr, recursion_dict = IdDict(
     new_array = Array{T,1}( undef, len_list)
     recursion_dict[obj] = new_array
     for i in 1:len_list
-        current_obj = obj[ i ]
+        current_obj = ElmList(obj,i)
+        if haskey(recursion_dict,current_obj)
+            new_array[ i ] = recursion_dict[current_obj]
+        else
```
fingolfin: I wonder whether this is what we want to do or not... see also PR #178... We need to write down the rules for who is responsible for setting `recursion_dict ` entries, and who for checking `recursion_dict` entries: the caller or the callee. Right now, it seems to be a mix: *Some* methods check at the start whether the given object is in `recursion_dict`, and if so, return the value it maps to. OTOH, some functions check before performing recursive conversion. 

The easiest to write is to add `if haskey(recursion_dict,obj) then return obj end` at the start of each conversion function. But then we need to do that even if we don't have a recursive conversion, and we need it even if we just convert some ints. 

So the next idea was to instead perform this check inside any conversion function for container types, like arrays/dicts resp. lists/records. But we should not do both. 

If we do that, we want to be careful, so that e.g. two identical GAP large ints still get converted into two identical Julia bigints (although arguably, this is less important than it is for container types).

I guess we could recover the first approach w/o loosing much or anything, if we just modify the fallback method
```julia
julia_to_gap(obj::Any, recursive, recursion_dict ) = julia_to_gap(obj)
```
to instead look like this:
```julia
function julia_to_gap(obj::Any, recursive, recursion_dict )
    if haskey(recursion_dict,obj)
        return recursion_dict[obj]
    end
    return julia_to_gap(obj)
end
```
But then of course any proper 3-arg method for `julia_to_gap` still needs to contain that check... 

So I am still not completely sure where placing that check makes most sense, but I *do* know that we should do it either-or, not both.

Anyway, there is another related question: If we have a list that contains itself, like the GAP list `l:=[~]`. Then it is clear how we want to handle it for recursive conversions. But what about non-recursive? In the current approach, what happens is that as we iterate over the content of `l`, we discover that its first entry was already converted, and use that result; so we end up creating a Julia list containing itself, even if recursion is off. I'd argue this is wrong, because it is inconsistent with the result for `l := [ [ ~ ] ]`: This list also contains itself, but one level deeper down. Since we don't recurse, we never get there, and thus the reference to the original GAP list is retained. It seems odd that this would work differently here.

Assuming that we do *not* want the `if haskey(recursion_dict,obj)` at the start of every conversion function, this suggests the following kind of logic for recursive conversions:
```julia
  ...
  current_obj = ...
  if Recursive
    if haskey(recursion_dict,current_obj)
      current_obj = recursion_dict[current_obj]
    else
      current_obj = recursion_dict[current_obj] = gap_to_julia(T, current_obj, recursion_dict)
    end
  end
 # ... now put current_obj into the container we are creating...
```
Thoughts?

sebasguts: I would be in favor of the variant where the caller handles the recursion. While I think both is feasible, it seems more natural to me that the caller checks the recursion.

I would guess for the non-recursive conversion, we can skip the recursion alltogether.

sebasguts: But btw, this is not really the scope of this PR, isn't it?

fingolfin: "the callers handles the conversion" is what I did for `julia_to_gap` in PR #178, so it seems we agree.

And I do think that my comments here are at least somewhat relevant to this PR, because this PR changes the recursion code for `gap_to_julia`, and I think that code is incorrect per our common understanding how recursive conversion should work. Of course it may have been incorrect before; but in any case, when reviewing this PR, I noticed the problem.

Specifically, I'd argue that if one wants to non-recursively convert the GAP self-containing list `[~]`, the result should be a Julia list containing that GAP list. But that's not the case right now... 

So, while this PR does not introduce a new issue, it promulgates an existing one; and a fix for that will necessarily conflict with this PR. So, let's merge this PR first, then work on a fix.

LibGAP.jl/test/conversion.jl
```diff
@@ -104,6 +104,15 @@
     xx = GAP.julia_to_gap( "a" )
     @test_throws MethodError GAP.gap_to_julia( Dict{Int64,Int64}, xx )
 
+    ## Test nothing
```
fingolfin: While this is a cute joke, it's not really helpful...
```suggestion
    ## Test converting GAP lists with holes in them
```

LibGAP.jl/test/conversion.jl
```diff
@@ -182,4 +191,8 @@ end
     xx = GAP.EvalString("[~]");
     @test xx === xx[1]
 
+    ## Test nothing
```
fingolfin:
```suggestion
    ## Test converting lists with 'nothing' in them -> should be converted to a hole in the list
```

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -143,7 +144,30 @@ function gap_to_julia( ::Type{Array{T,1}}, obj :: MPtr, recursion_dict = IdDict(
...
+            new_array[ i ] = gap_to_julia(T,current_obj,recursion_dict)
+            recursion_dict[ current_obj ] = new_array[ i ]
+        end
+    end
+    return new_array
+end
+
+## Special case for conversion of lists with holes
```
fingolfin:
```suggestion
## Special case for conversion of lists with holes; these are converted into 'nothing'
```

LibGAP.jl/src/gap_to_julia.jl
fingolfin:
```suggestion
        current_obj = ElmList(obj,i)  # returns 'nothing' for holes in the list
```

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -143,7 +144,30 @@ function gap_to_julia( ::Type{Array{T,1}}, obj :: MPtr, recursion_dict = IdDict(
...
+function gap_to_julia( ::Type{Array{Union{Nothing,T},1}}, obj :: MPtr, recursion_dict = IdDict() ) where T
+    if ! Globals.IsList( obj )
+        throw(ArgumentError("<obj> is not a list"))
+    end
+    if haskey(recursion_dict,obj)
+        return recursion_dict[obj]
+    end
+    len_list = length(obj)
+    new_array = Array{Union{Nothing,T},1}( undef, len_list)
```
fingolfin:
```suggestion
    new_array = Array{Union{Nothing,T},1}( undef, len_list )
```
or
```suggestion
    new_array = Array{Union{Nothing,T},1}(undef, len_list)
```

LibGAP.jl/src/gap_to_julia.jl
sebasguts: > "the callers handles the conversion" is what I did for julia_to_gap in PR #178, so it seems we agree.
Yep.
> And I do think that my comments here are at least somewhat relevant to this PR, because this PR changes the recursion code for gap_to_julia,

I do not see where I actually changed the way recursion is handled in this PR.

This is not against your comments whatsoever, they make perfect sense and are important. But I would still rather discuss it in a new PR.

sebasguts:
I rebased the PR and applied the changes. Can we merge this?

LibGAP.jl/src/gap_to_julia.jl
```diff
@@ -13,6 +13,7 @@ of converted objects and should never be given by the user.
 gap_to_julia(::Type{GAPInputType},x::GAPInputType) = x
 gap_to_julia(::Type{Any},         x::GAPInputType) = gap_to_julia(x)
 gap_to_julia(::Type{Any},         x::Any         ) = x
+gap_to_julia(::Any,               x::Nothing     ) = nothing
```
fingolfin: BTW, why is this `::Any` and now `::Type{Any}` as above? I don't understand what's going on here.

sebasguts:
Thx for merging :)

PR #174, opened 2018-11-26
redesigned parts of JuliaExperimental 

ThomasBreuer:
- introduced context objects for conversions between GAP and Nemo,
  which deal with elements and vectors and matrices of these elements;
  currently implemented for Integers, Rationals, Integers mod n,
  algebraic extensions, polynomial rings

- added code for invariant theory


fingolfin:
The tests in `zmodnz.tst` failed, starting in line 27, with
```
# Input is:
Order( m );
# Expected output:
3
# But found:
Error, MethodError: no method matching one(::Nemo.nmod_mat)
Closest candidates are:
  one(!Matched::BitArray{2}) at bitarray.jl:393
  one(!Matched::Missing) at missing.jl:79
  one(!Matched::Nemo.FlintIntegerRing) at /home/travis/.julia/packages/Nemo/XN\
\
d31/src/flint/fmpz.jl:142
  ...
```
Other than that, I wonder what the plan with this PR is -- i.e., what is planned to do before it can change from "WIP" to "please review and merge this" ? Also, is there anything we can do to help?

ThomasBreuer:
@fingolfin Thanks for the reply.
My plan is to update the pull request by adding more conversions, also to and from the Singular system,
and to shorten the code.

Concerning the mentioned test failure, apparently I had overlooked that some features
do not yet work in the GAP master branch; I am going to create a pull request for that.

JuliaExperimental/gap/singular.g
```diff
@@ -415,14 +415,25 @@ InstallOtherMethod( One, [ "IsSingularObject" ], 200,
...
 ##
+
+
+##############################################################################
+##
+#M  Degree( <pol> ) . . . . . . . . . . . . . .  for a multivariate polynomial
+##
+InstallOtherMethod( Degree,
+    [ "IsSingularPolynomial" ],
+    pol -> Julia.Singular.degree( JuliaPointer( pol ) ) );
```
fingolfin: Just in case you are not yet aware of it: If you wrapper objects were in the category `IsJuliaWrapper`, then you could replace `JuliaPointer( pol )` above by just `pol`. And the same in various other places where the wrapper object is passed as argument to a Julia function.

ThomasBreuer: Yes, @sebasguts had told me about this feature, and I will use it.

In fact, one can even install `Julia.Singular.degree` itself as a method,
there is no need to wrap it in another GAP function.
(In earlier versions, this was not possible.)

fingolfin:
What's the status of this, resp. the plan for further development? What's missing before it can be merged?

ThomasBreuer:
@fingolfin I have extended the pull request, and thought that now it could be merged.
However, the tests fail.
As far as I see, a problem is that Travis and I rely on different versions of Nemo.
In Nemo 0.10.3 (used by Travis), there are functions `Nemo.cols` and `Nemo.rows`,
whereas these functions have been renamed to `Nemo.ncols` and `Nemo.nrows` in Nemo 0.12.0 (used by me).
As far as I see, I cannot win this game.

sebasguts:
@ThomasBreuer Is this Nemo version already released?
@fingolfin Should we clean the caches?

ThomasBreuer:
@sebasguts Thank you for the hint. Apparently my Julia uses the #master version of Nemo.
I will downgrade my code to the older Nemo version.

fingolfin:
There might be no need to downgrade, I'll clear Travis' caches now and then will restart the tests

fingolfin:
Ah, wait, or was the conclusion now tha this was not yet released? Sorry, I think I mistead

ThomasBreuer:
@sebasguts / @fingolfin After downgrading to Nemo 0.10, the test results are o.k.
Sorry for the confusion.

fingolfin:
Why exactly do we have to use Nemo 0.10, though? On https://github.com/Nemocas/Nemo.jl/ the latest released version is v0.12.3, and this is also what I get in Julia. What am I missing?

PR #178, opened 2018-11-27
simplify julia_to_gap, add test for recursive record

fingolfin:


LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -103,16 +99,14 @@ function julia_to_gap(obj::Dict{T,S}, recursive::Val{Recursive}=Val(false), recu
...
-        keys[i] = x
-        entries[i] = y
-        i += 1
-    end
-    record = GAP.Globals.CreateRecFromKeyValuePairList(julia_to_gap(keys,Val(true)),julia_to_gap(entries, recursive, recursion_dict))
+
+    k = collect(T, keys(obj))
+    v = collect(S, values(obj))
+
+    k = julia_to_gap(k, Val(true))
```
sebasguts: NITPICK: No need to put the `Val(true)` here. The keys have to be `Symbol` or `String`, and both conversions ignore the recursion anyway.

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -103,16 +99,14 @@ function julia_to_gap(obj::Dict{T,S}, recursive::Val{Recursive}=Val(false), recu
...
+    v = julia_to_gap(v, recursive, recursion_dict)
+
+    record = GAP.Globals.CreateRecFromKeyValuePairList(k, v)
     ## FIXME: This is too late!
```
sebasguts: Actually, we could resolve this `FIXME` by creating an empty record at the beginning of the conversion function and then fill it with the converted stuff.

LibGAP.jl/src/julia_to_gap.jl
fingolfin: To be honest, I don't understand this line at all, but I copied verbatim form what you wrote... I'd have expected `k = julia_to_gap(k, recursive, recursion_dict)`

LibGAP.jl/src/julia_to_gap.jl
fingolfin: ... instead of using `GAP.Globals.CreateRecFromKeyValuePairList` at all. Yes, thought the same. I'll update the PR accordingly. And then I should put in a test for this, too...

LibGAP.jl/src/julia_to_gap.jl
sebasguts: In a sense, this makes sense. But the list of keys `k` (which I am not sure is the name I used) is a list of strings anyway. You can do recursion here, but it does not matter at all.

LibGAP.jl/src/julia_to_gap.jl
```diff
@@ -72,29 +72,18 @@ function julia_to_gap(obj::Array{T,1}, recursive::Val{Recursive}=Val(false), rec
...
-            if Recursive
-                current_converted = julia_to_gap(current_obj,recursive,recursion_dict)
-            else
-                current_converted = current_obj
-            end
-            recursion_dict[current_obj] = current_converted
-            ret_val[i] = current_converted
+        x = obj[i]
+        if Recursive
+            x = julia_to_gap(x, recursive, recursion_dict)
```
fingolfin: Actually, I am now unsure whether this change is quite correct. It fixed a bug, but I think it's not quite what we intend. I'll elaborate in another comment on PR #173. Please do not merge until this is resolved.

fingolfin: I have now rewritten this PR yet again, to match what I explained on PR #173

fingolfin:
From my POV this is ready to be merged.

PR #180, opened 2018-11-29
Minor changes, a few more tests

fingolfin:


JuliaInterface/gap/BindCFunction.gi
```diff
@@ -31,13 +31,13 @@ InstallGlobalFunction( JuliaBindCFunction,
 
     ## Create a call to the @cfunction macro, to get a pointer to a compiled function
     cfunction_call_string := Concatenation( "temp = @cfunction(", julia_name );
-    cfunction_call_string := Concatenation( cfunction_call_string, ",Ptr{Cvoid},(" );
+    Append( cfunction_call_string, ",Ptr{Cvoid},(" );
```
sebasguts: I just wished we had `+` or `*` for strings...

fingolfin: But then you'd still not want to use it here, unless we *also* had `*=`, because `cfunction_call_string := cfunction_call_string * "BLABLA";` still is worse than `Append( cfunction_call_string, "BLABLA" );`, at least IMHO...

PR #182, opened 2018-11-29
JuliaFunction: set name and "location" of the function

fingolfin:
The latter means that using `Display` on a GAP function produced via
`JuliaFunction` will at least give some indication about where the function
comes from. It's not perfect, but certainly better than nothing (= what we
currently get).

### DIFF
--- a/JuliaInterface/tst/utils.tst
+++ b/JuliaInterface/tst/utils.tst
@@ -101,4 +101,12 @@ gap> JuliaGetFieldOfObject(JuliaModule("Base"), "not-a-field");
 Error, type Module has no field not-a-field
 
 ##
+gap> NameFunction(Julia.Base.parse);
+"parse"
+gap> Display(Julia.Base.parse);
+function ( arg... )
+    <<kernel code>> from Julia:parse
+end
+
+##
 gap> STOP_TEST( "utils.tst", 1 );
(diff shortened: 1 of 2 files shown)


JuliaInterface/tst/utils.tst
```diff
@@ -100,5 +100,13 @@ Error, JuliaGetFieldOfObject: <field_name> must be a string
 gap> JuliaGetFieldOfObject(JuliaModule("Base"), "not-a-field");
 Error, type Module has no field not-a-field
 
+##
+gap> NameFunction(Julia.Base.parse);
+"parse"
+gap> Display(Julia.Base.parse);
+function ( arg... )
+    <<kernel code>> from Julia:parse
```
fingolfin: The `<<kernel code>> from ` bit is fixed by GAP (at least currently), and so is the "FOO:BAR" format for the final part. Nice would of course be something like: `<<Julia function Base.parse>>`. Alas, I also don't really know how to produce `Base.parse` instead of `parse`. Perhaps it's as simply as checking for the parent module, getting its name, and prepending it (with a dot), and then repeating this until the top level is reached? Alas, we can figure that out later.

sebasguts: I would say it is fine that way for now. If people start to complain we can still fix it. A sensible top-level would be `Main` I guess. Or maybe `Top`?

PR #185, opened 2018-12-14
autogen.sh in README

heiderich:
Without it `configure` fails.


sebasguts:
I am not really convinced by this tbh. If you have a download version from GAP, you don't need to ececute `autogen.sh`. Anyway, the `README.md` in the LibGAP.jl is incomplete anyway, so I will just merge this.

PR #186, opened 2019-01-02
WIP: Update .travis.yml to use their pre-install Julia 1.1

fingolfin:


sebasguts:
@fingolfin Can you have a look at this? The test claims that `etc/travis_build.sh` is missing, but the file is there. So maybe the command gets executed in the wrong directory.

fingolfin:
Obsolete

PR #188, opened 2019-01-25
Installation of all OSCAR parts for GAPJulia

frankluebeck:
This installs and fully builds on my Debian 9 installation inside the current directory
   * gap (master)
      * with pkgfull
      * and GAPJulia
   * julia (latest)
      * with AbstractAlgebra, Singular, GB, Nemo, Hecke, Polymake
   * polymake

Maybe a check could be added to test for the needed Debian/Ubuntu/Mintpackages?

### DIFF
--- a/etc/install_julia_gap.sh
+++ b/etc/install_julia_gap.sh
@@ -6,10 +6,11 @@ main_dir=${PWD}
 
 # Download and install Julia
 wget https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz
-mkdir julia 
+mkdir julia
 tar xf julia-latest-linux64.tar.gz -C julia --strip-components 1
 
-julia/bin/julia <<UNTILEND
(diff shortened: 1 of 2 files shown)


etc/install_julia_gap.sh
```diff
@@ -9,19 +9,54 @@ wget https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linu
 mkdir julia 
 tar xf julia-latest-linux64.tar.gz -C julia --strip-components 1
 
+julia/bin/julia <<UNTILEND
+using Pkg
+Pkg.add("AbstractAlgebra")
+Pkg.add("Nemo")
+#Pkg.develop("Hecke")
```
sebasguts: Could you remove this line?

sebasguts:
Okay, this is probably not going to work that way. Now GAP builds all packages and the manual, and you get unreadable travis output.

I would suggest we either maintain two scripts, one for travis testing and one for "installation", or remove the complete installation script and finally make the OSCAR install scripts repo

PR #189, opened 2019-01-29
Added JULIA_LDFLAGS and JULIA_LIBS to Makefile.in

sebasguts:
This is a follow-up to https://github.com/gap-system/gap/pull/3248


sebasguts:
Good idea.

PR #191, opened 2019-01-30
Check if JuliaInterface could be loaded

sebasguts:
when GAP is loaded into Julia


sebasguts:
I did indeed. Also, I was reluctant to create a testcase, as the most sensible way is to move the package away.

PR #194, opened 2019-02-07
JuliaInterface: don't use FEXS in function wrapper code

fingolfin:
This replaces the (ab)use of FEXS to store pointers to Julia function
by a change which uses slightly larger T_FUNCTION bags in order to make
room for the function pointer.

Necessary for compatibility with <https://github.com/gap-system/gap/pull/3265>.
Requires <https://github.com/gap-system/gap/pull/3276>.


fingolfin:
@sebasguts GAP is "ready" now, tests passed after re-running them. I'll merge.

PR #196, opened 2019-02-12
Add \. and \.\:\= adapter for {get,set}property

fingolfin:
Also rename arith.{gi,tst} to adapter.{gi,tst}.


JuliaInterface/gap/adapter.gi
```diff
@@ -50,6 +37,66 @@ Perform([[ "IsJuliaObject", "IsJuliaObject" ],
...
+    [ "IsJuliaObject", "IsPosInt and IsSmallIntRep",
+                       "IsPosInt and IsSmallIntRep", "IsObject" ],
+    function( obj, i, j, val )
+      return Julia.Base.setindex\!( obj, val, i, j );
+    end );
+
+#
+# access to fields and properties
+#
+BindGlobal("_JL_RNAM_TO_JULIA_SYMBOL_CACHE", rec());
```
sebasguts: Guess this is an optimization to safe you looking the time looking up the string which corresponds to an `rnam`?

fingolfin: Yes, this is an optimization.

PR #201, opened 2019-02-19
Compare tests up to whitespace

sebasguts:


sebasguts:
Merged as part of #174

PR #202, opened 2019-02-19
use a Julia iterator in `MinimalDegreeHard`

ThomasBreuer:


sebasguts:
The missing coverage data comes from the fact that `Primes` is missing in Travis. Will try to add it to travis

sebasguts:
@ThomasBreuer The loewy test is too long, now travis terminates. I think you could tackle this by either shorten the loewy tests or add prints.

PR #203, opened 2019-02-19
WIP: Import conversion rules into AutoDoc

fingolfin:
This is an incomplete conversion of https://github.com/oscar-system/GAPJulia/wiki/Rules-for-conversion-between-GAP-and-Julia-objects which I never finished.

Problem: AutoDoc only supports a subset of Markdown, and in particular does not support those (non-standard) Markdown tables.

So it might be better to either convert this manually into GAPDoc XML; or to import this into the "Julia documentation", however that is done, as that seems to support proper Markdown (but I don't know if it supports tables).

In any case, we might want to just discard this PR; but I thought I should at least post it here, in case it is of use to anybody.


fingolfin:
Resolved by @ThomasBreuer via PR #353

PR #207, opened 2019-02-25
Towards a Julia module

sebasguts:
This PR is going to make GAPJulia a loadable Julia module. The steps include (splitted into commits already)

- [x] Move `.jl` files which are loaded by JuliaInterface to JuliaInterface
- [x] Move GAP packages into a `pkg` subfolder
- [x] Create a julia module
- [x] Add a build.jl that fetches and builds GAP
- [x] Add Julia `__init__` function that automatically loads GAP
- [x] Adjust travis scripts


JuliaInterface/read.g
```diff
@@ -6,17 +6,17 @@
...
 
-JuliaIncludeFile( Filename( dirs, "gaptypes.jl" ) );
+JuliaIncludeFile( Filename( dirs_julia, "gaptypes.jl" ) );
 
-dirs:= DirectoriesPackageLibrary( "JuliaInterface", "../LibGAP.jl/src" );
+dirs_libgap := DirectoriesPackageLibrary( "JuliaInterface", "../LibGAP.jl/src" );
 
 if not IsBound( JULIAINTERNAL_LOADED_FROM_JULIA ) then
-    JuliaEvalString( Concatenation( "Base.include( Main,\"", Filename( dirs, "initialization.jl" ), "\")" ) );
+    JuliaEvalString( Concatenation( "Base.include( Main,\"", Filename( dirs_libgap, "initialization.jl" ), "\")" ) );
```
fingolfin: Question (kinda orthogonal to this PR, i.e., not a blocker to this PR, but looking at this code again made me wonder):  Why do we includle `gaptypes.jl` before `initialization.jl`? Bedeutet das nicht, dass deren relative Reihenfolge unterschiedlich ist, je nach dem, ob man GAP von Julia aus lädt, oder umgekehrt?

Falls es dafür einen technischen Grund gibt, würde ich den gerne hier in einem Code-Kommentar dokumentieren. Falls es keinen Grund gibt, würde ich die Reihenfolge tauschen; denn auch wenn es gerade keinen Unterschied macht, so könnte es doch eines Tags zu seltsamen Differenzen im Verhalten führen, welche dann nervig zu debuggen sind.

Thoughts?

sebasguts: No, there is no real necessity to include `gaptypes.jl` before `initialization.jl`. It is just historical in the order the files were written in. The files are orthogonal to each other. However, `gaptypes.jl` needs to be loaded before any other of the files.

sebasguts:
Another relevant question for this PR is: Do we require every commit to work on its own? I just realised that this would mean we have to touch some files several times (mostly makefiles and files with paths) or are we just fine with it working at the end?

JuliaInterface/read.g
fingolfin: OK, then let's swap it (doesn't have to be in this PR, of course)

fingolfin:
Of course it'd be nice if each commit works on its own, but if it's annoying to do, don't bother, we can always `git bisect skip` these commits.

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -0,0 +1,29 @@
...
+# Reading the implementation part of the package.
+#
+ReadPackage( "JuliaInterface", "gap/JuliaInterface.gi");
+ReadPackage( "JuliaInterface", "gap/BindCFunction.gi" );
+
+dirs_julia := DirectoriesPackageLibrary( "JuliaInterface", "julia" );
+
+JuliaIncludeFile( Filename( dirs_julia, "gaptypes.jl" ) );
+
+dirs_libgap := DirectoriesPackageLibrary( "JuliaInterface", "../../../src" );
```
fingolfin: This seems rather brittle to me. Can we ask Julia to resolve the path for us? It should "know" where `GAP.jl` is, shouldn't it?

(Also, I find the name `dirs_libgap` a little bit confusing -- to many things are referred to by libgap. Perhaps `dirs_libgap_jl` or `dirs_gap_jl`? Of course if we don't need them anymore, this becomes moot.)

sebasguts:
I disabled the draft state to test if travis works (it should, modulo typos in scripts), afterwards I will adress the comments.

sebasguts:
@fingolfin @ThomasBreuer This PR is ready for review now. How would you prefer the review process? The PR itself is not too large, but touched almost every file in the Repo.

Do you want me to rewrite the commit history first, so we can have a clearer history? I would actually prefer if you look at the changes first, and then I can redo the history, as it will save time if I only have to do it once. I think with a little bit of effort I can make basically every commit compile and run. Or we just squash the whole PR. The commits in the meantime are maybe not too useful.

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -0,0 +1,30 @@
...
+# Reading the implementation part of the package.
+#
+ReadPackage( "JuliaInterface", "gap/JuliaInterface.gi");
+ReadPackage( "JuliaInterface", "gap/BindCFunction.gi" );
+ReadPackage( "JuliaInterface", "gap/generated_path.gi" );
+
+dirs_gap_jl := Directory( _JULIAINTERFACE_JULIA_MODULE_SOURCES );
+
+if not IsBound( JULIAINTERNAL_LOADED_FROM_JULIA ) then
+    JuliaEvalString( "__IS_LOADED_FROM_GAP = true");
```
fingolfin:
```suggestion
    JuliaEvalString( "__IS_LOADED_FROM_GAP = true" );
```

deps/build.jl
```diff
@@ -0,0 +1,91 @@
...
+        run(`make bootstrap-pkg-minimal`)
+    else
+        run(`mkdir pkg`)
+    end
+end
+
+gap_executable = abspath(joinpath(gap_root,"gap"))
+
+## Compile JuliaInterface/Experimental
+cd(abspath(joinpath(@__DIR__, "..", "pkg", "GAPJulia" )))
```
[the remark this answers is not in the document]
ThomasBreuer: I think the question is whether GAP sees one meta-package GAPJulia or two independent packages.
The common `configure` fits only for the meta-package setup.
A related question is whether the name `GAPJulia` for the git repository is reasonable.

.travis.yml
```diff
@@ -3,7 +3,7 @@ env:
  global:
    - DONT_FETCH_GAP_JULIA_PKG=yes
    - JULIAROOT=$TRAVIS_BUILD_DIR/julia
-   - GAPROOT=$TRAVIS_BUILD_DIR/gap
+   - GAP_ROOT=$TRAVIS_BUILD_DIR/gap
```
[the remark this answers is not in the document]
sebasguts: Well, this was basically done because I introduced the `GAP_ROOT` in the build.jl, and then the cascade of changes piled up. I will change them all to `GAPROOT`

.travis.yml
```diff
@@ -26,9 +26,7 @@ before_script:
   - git show --pretty=fuller -s
   - mkdir -p coverage
   - etc/install_julia_gap.sh
```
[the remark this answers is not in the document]
sebasguts: Should I try to put #186 into this PR? I am not 100% sure, but it might need adjustments

Project.toml
```diff
@@ -0,0 +1,9 @@
+name = "GAP"
+uuid = "c863536a-3901-11e9-33e7-d5cd0df7b904"
+authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsche@mathematik.uni-kl.de>", "Max Horn <max@quendi.de>"]
+version = "0.1.0"
+
+[deps]
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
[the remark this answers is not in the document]
sebasguts: `Pkg` is for building (as we need `Pkg.depots1`), and `Libdl` is actually a dependency. I found it quite interesting, too, as they both should be standard packages. However, without the explicit dependency, the packages are not found when building them into a new, clean Julia depot. That is why I added them. Actually, this might be an inconsistency in other packages. I think we are doing the right thing here.

Project.toml
```diff
@@ -0,0 +1,9 @@
+name = "GAP"
+uuid = "c863536a-3901-11e9-33e7-d5cd0df7b904"
+authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsche@mathematik.uni-kl.de>", "Max Horn <max@quendi.de>"]
+version = "0.1.0"
+
+[deps]
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
```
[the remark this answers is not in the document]
sebasguts: See comment above.

.travis.yml
```diff
@@ -26,9 +26,7 @@ before_script:
...
-  - ln -s $TRAVIS_BUILD_DIR/JuliaInterface $GAPROOT/pkg/
-  - ln -s $TRAVIS_BUILD_DIR/JuliaExperimental $GAPROOT/pkg/
-  - ln -s $TRAVIS_BUILD_DIR/LibGAP.jl $GAPROOT/pkg/
+  - ${JULIAROOT}/bin/julia -e 'import Pkg ; Pkg.develop(Pkg.PackageSpec(url="$(ENV["TRAVIS_BUILD_DIR"])")) ; Pkg.build("GAP")'
```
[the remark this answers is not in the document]
sebasguts: Actually, we don't. I just was not sure if this `Pkg.build()` would work, but it seems like the right way to do it.

Project.toml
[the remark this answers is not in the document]
sebasguts: See above

deps/build.jl
```diff
@@ -0,0 +1,91 @@
...
+    run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
+    run(`make`)
+    if ! haskey(ENV,"GAP_INSTALL_PACKAGES") || ENV["GAP_INSTALL_PACKAGES"] == "yes" 
+        run(`make bootstrap-pkg-full`)
+        cd("pkg")
+        run(`../bin/BuildPackages.sh`)
+    elseif haskey(ENV,"GAP_INSTALL_PACKAGES") && ENV["GAP_INSTALL_PACKAGES"] == "minimal"
+        run(`make bootstrap-pkg-minimal`)
+    else
+        run(`mkdir pkg`)
```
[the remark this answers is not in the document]
sebasguts: It might be that people want to not download or compile any packages, as they might want to symlink a custom `pkg` folder into this GAP. Do you think this is unnecessary to create the folder?

deps/build.jl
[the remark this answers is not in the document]
sebasguts: This is exactly what I was thinking of. There is still the ongoing work on meta-packages, and I really want to make GAPJulia a meta-package at some point. It is also a natural point for the global `configure` and `make`, which I would not really like to put into the `pkg` folder. I would rather keep it structurized like this.

Regarding the Repo's name: `GAP.jl` would probably be a better name, granted. But I cannot rename a repo in a PR.

deps/build.jl
```diff
@@ -0,0 +1,91 @@
...
+end
+
+gap_executable = abspath(joinpath(gap_root,"gap"))
+
+## Compile JuliaInterface/Experimental
+cd(abspath(joinpath(@__DIR__, "..", "pkg", "GAPJulia" )))
+run(`./configure $gap_root`)
+run(`make`)
+
+new_gap_root = abspath(joinpath(@__DIR__,".."))
```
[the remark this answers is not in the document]
sebasguts: I named it `EXTRA_GAP_ROOT` now.

src/GAP.jl
```diff
@@ -90,10 +90,19 @@ run_it = function(gapdir::String)
     initialize( [ ""
                        , "-l", sysinfo["GAP_LIB_DIR"]
```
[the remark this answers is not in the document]
sebasguts: It should point to the standard GAP root as defined in `sysinfo.gap`, because it comes from the sysinfo. I am not sure it is really necessary. But it was also added by you in e56d23a769010d402f7703ab5ac1e871a3019636

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -0,0 +1,30 @@
+#
+# JuliaInterface: Test interface to julia
+#
+# Reading the implementation part of the package.
+#
+ReadPackage( "JuliaInterface", "gap/JuliaInterface.gi");
+ReadPackage( "JuliaInterface", "gap/BindCFunction.gi" );
+ReadPackage( "JuliaInterface", "gap/generated_path.gi" );
+
+dirs_gap_jl := Directory( _JULIAINTERFACE_JULIA_MODULE_SOURCES );
```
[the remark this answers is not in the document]
sebasguts: I am aware of this, and the reason why I did all of this is different: At this point of the module/package loading stage, I simply cannot translate the Julia string back to GAP. So we have to choose one of those evils here:

1. Have a C-funtion that translates a Julia string to a GAP string
2. Have a C-function that does what you described above
3. Use the generated file with the path

If you really want to use your own version of JuliaInterface, then with the current setup you would have to not only copy over the `generated_path.gi`, but also change the `deps.jl` (because otherwise the GAP called by Julia will anyway try to load their own version of GAP). So What I would suggest is actually symlinking the `GAPJulia` folder into your local GAP pkg folder. I am not sure what this means for using `git`, though, because the `.git` folder on top of the symlink might be wrong.

.travis.yml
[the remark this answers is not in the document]
sebasguts: Well, currently, I still build against the extra compiled GAP. I can of course only use the GAP installed by julia

deps/build.jl
```diff
@@ -0,0 +1,84 @@
+# Builds a GAP, and attaches JuliaInterface
+
+import Pkg
+
+julia_folder_path = abspath(joinpath(@__DIR__,".."))
```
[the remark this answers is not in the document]
sebasguts: I will reformat the `build.jl`

deps/build.jl
```diff
@@ -0,0 +1,84 @@
...
+    cd(julia_folder_path)
+    run(`git clone https://github.com/gap-system/gap`)
+    cd(gap_root)
+    run(`./autogen.sh`)
+    run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
+    run(`make`)
+    if ! haskey(ENV,"GAP_INSTALL_PACKAGES") || ENV["GAP_INSTALL_PACKAGES"] == "yes" 
+        run(`make bootstrap-pkg-full`)
+        cd("pkg")
+        run(`../bin/BuildPackages.sh`)
```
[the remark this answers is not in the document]
sebasguts: Why not?

deps/build.jl
```diff
@@ -0,0 +1,84 @@
...
+end
+
+## Find julia binary
+julia_binary = julia_binary = get(ENV, "JULIA_BINARY", Sys.BINDIR)
+
+## Install GAP
+if install_gap
+    cd(julia_folder_path)
+    run(`git clone https://github.com/gap-system/gap`)
+    cd(gap_root)
```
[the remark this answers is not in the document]
sebasguts: Okay.

deps/build.jl
```diff
@@ -0,0 +1,84 @@
...
+    install_gap = false
+end
+
+## Find julia binary
+julia_binary = julia_binary = get(ENV, "JULIA_BINARY", Sys.BINDIR)
+
+## Install GAP
+if install_gap
+    cd(julia_folder_path)
+    run(`git clone https://github.com/gap-system/gap`)
```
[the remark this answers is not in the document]
sebasguts: I added a comment.

deps/build.jl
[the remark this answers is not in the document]
sebasguts: I reformatted the `build.jl`

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -0,0 +1,30 @@
...
+
+if not IsBound( JULIAINTERNAL_LOADED_FROM_JULIA ) then
+    JuliaEvalString( "__IS_LOADED_FROM_GAP = true" );
+    JuliaEvalString( Concatenation( "Base.MainInclude.include( \"", Filename( dirs_gap_jl, "GAP.jl" ), "\")" ) );
```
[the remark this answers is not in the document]
sebasguts: Actually, even with my comment above, your suggestion seems fine. Although I am not sure we can do much error handling, except the fact that GAP might catch the Julia error. Which is some error handling :). I will try.

.travis.yml
```diff
@@ -1,11 +1,16 @@
...
+
 env:
  global:
    - DONT_FETCH_GAP_JULIA_PKG=yes
-   - JULIAROOT=$TRAVIS_BUILD_DIR/julia
+   - DONT_FETCH_JULIA=yes
    - GAPROOT=$TRAVIS_BUILD_DIR/gap
    - MAKEFLAGS="-j4"
    - JULIA_TRACK_COVERAGE=yes
+   - JULIA_PROJECT=${HOME}/.julia/environments/v1.0/
```
fingolfin: Naive question, but: does using `v1.0` also work with the 1.1 and nightly builds? I am assuming it does, else you'd not have added it and Travis would not succeed, I am simply surprised... And I checked, on my system with Julia 1.1, there is only `~/.julia/environments/v1.1/` but no `~/.julia/environments/v1.0/`. So what gives?

.travis.yml
```diff
@@ -22,23 +27,23 @@ branches:
     - master
 
 before_script:
+  - export JULIA_PROJECT=${HOME}/.julia/environments/v1.0/
```
fingolfin: Hm, couldn't you simply `unset JULIA_PROJECT`, to disable the use of a custom Julia project completely?

.travis.yml
```diff
@@ -22,23 +27,23 @@ branches:
...
   - printenv | sort
   - git show --pretty=fuller -s
   - mkdir -p coverage
   - etc/install_julia_gap.sh
-  - ln -s $TRAVIS_BUILD_DIR/JuliaInterface $GAPROOT/pkg/
-  - ln -s $TRAVIS_BUILD_DIR/JuliaExperimental $GAPROOT/pkg/
-  - ln -s $TRAVIS_BUILD_DIR/LibGAP.jl $GAPROOT/pkg/
+    ## Add the package so we can use it later
+  - julia --project="$JULIA_PROJECT" -e 'import Pkg ; Pkg.add(Pkg.PackageSpec(url="$(ENV["TRAVIS_BUILD_DIR"])"))'
```
fingolfin: Since `JULIA_PROJECT` is explicitly checked by `julia`, I don't think we need to pass it to Julia. Also, we are in the package dir... So how about this instead?
```suggestion
  - julia -e 'import Pkg ; Pkg.add(Pkg.PackageSpec(path="."))'
```

.travis.yml
```diff
@@ -22,23 +27,23 @@ branches:
...
+  - julia --project="$JULIA_PROJECT" -e 'import Pkg ; Pkg.add("Nemo")'
```
fingolfin:
```suggestion
  - julia -e 'import Pkg ; Pkg.add("Nemo")'
```

.travis.yml
```diff
@@ -22,23 +27,23 @@ branches:
...
-  - ${JULIAROOT}/bin/julia -e 'import Pkg ; Pkg.add("Nemo")'
   - etc/travis_test.sh
     # delete coverage files for packages, so that they do not get cached
   - find $HOME/.julia/packages/ -name "*.jl.*.cov" -exec rm -f {} \;
 
 after_script:
   - cd $TRAVIS_BUILD_DIR
   - $GAPROOT/gap -A --quitonbreak --norepl etc/gather_coverage.g
-  - ${JULIAROOT}/bin/julia etc/gather_coverage.jl
+  - julia --project="$JULIA_PROJECT" etc/gather_coverage.jl
```
fingolfin:
```suggestion
  - julia etc/gather_coverage.jl
```

deps/build.jl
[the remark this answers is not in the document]
fingolfin: Because what it does is hard to predict; every use will potentially get a slightly different set of packages that work, making this into a bit of a support nightmare. Of course that's a deeper issue, and one of the many reasons `BuildPackages.sh` is only a stop-gap measure, but not a complete solution for the GAP package management issue. Anyway, leave it as is, it's just one more spot we'll have to change once `BuildPackages.sh` gets replaced by something better.

etc/travis_test.sh
```diff
@@ -10,30 +10,27 @@ AnyFailures=No
 # check code formatting
 find . -name '*.c' -exec clang-format -i {} \;
 find . -name '*.h' -exec clang-format -i {} \;
-git diff --exit-code -- . # detect if there are any diffs
+git diff --exit-code -- '*.c' '*.h' # detect if there are any diffs
```
fingolfin: Why this change? My best guess is: for the sake of `generated_path.gi`. But that is in `.gitignore`, so it shouldn't matter... so...?

deps/build.jl
```diff
@@ -0,0 +1,76 @@
...
+# and exposes the right paths to julia/GAP
+
+import Pkg
+
+julia_folder_path = abspath(joinpath(@__DIR__, ".."))
+
+gap_root = abspath(joinpath(julia_folder_path, "gap"))
+install_gap = true
+
+if haskey(ENV,"GAPROOT")
```
fingolfin:
```suggestion
if haskey(ENV, "GAPROOT")
```

deps/build.jl
```diff
@@ -0,0 +1,76 @@
...
+run(`./configure $gap_root`)
+run(`make`)
+
+extra_gap_root = abspath(joinpath(@__DIR__, ".."))
+
+## Write deps.jl file containing the necessary paths
+deps_string = """
+GAPROOT = "$gap_root"
+GAP_EXECUTABLE = "$gap_executable"
+EXTRA_GAP_ROOT = "$extra_gap_root"
```
fingolfin: Perhaps EXTRA_GAP_ROOT -> EXTRA_GAPROOT for consistency?

.travis.yml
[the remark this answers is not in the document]
fingolfin: I am confused; are you saying you are downloading and building GAP (and all packages, etc.) twice now?!

etc/install_julia_gap.sh
```diff
@@ -5,15 +5,17 @@ set -e
 main_dir=${PWD}
 
 # Download and install Julia
-wget https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz
-mkdir julia 
-tar xf julia-latest-linux64.tar.gz -C julia --strip-components 1
+if [[ $DONT_FETCH_JULIA != yes ]]; then
+    wget https://julialangnightlies-s3.julialang.org/bin/linux/x64/julia-latest-linux64.tar.gz
+    mkdir julia
+    tar xf julia-latest-linux64.tar.gz -C julia --strip-components 1
```
fingolfin: Should we really download a nightly here, and not v1.1? (Of course doesn't have to be in this PR)

etc/install_julia_gap.sh
```diff
@@ -5,15 +5,17 @@ set -e
...
+fi
 
 # Download and install GAP
 git clone --depth=1 https://github.com/gap-system/gap
 cd gap
 ./autogen.sh
-./configure --with-gc=julia --with-julia=${main_dir}/julia/
+./configure --with-gc=julia --with-julia=$(which julia)
```
fingolfin:
```suggestion
./configure --with-gc=julia --with-julia
```
This will simply look for julia in `PATH`

etc/travis_test.sh
```diff
@@ -10,30 +10,27 @@ AnyFailures=No
...
-$GAPROOT/gap -A --quitonbreak --norepl --cover $TRAVIS_BUILD_DIR/coverage/JuliaExperimental tst/testall.g || AnyFailures=Yes
+${HOME}/.julia/gap -A --quitonbreak --norepl --cover ${TRAVIS_BUILD_DIR}/coverage/JuliaExperimental tst/testall.g || AnyFailures=Yes
 gcov -o .. src/*.c*
 cd ..
 
 #
-cd LibGAP.jl
 pwd
-${JULIAROOT}/bin/julia --code-coverage --inline=no test/runtests.jl || AnyFailures=Yes
+julia --project="$JULIA_PROJECT" --code-coverage --inline=no -e 'using Pkg ; Pkg.test("GAP")' || AnyFailures=Yes
```
fingolfin:
```suggestion
julia --code-coverage --inline=no -e 'using Pkg ; Pkg.test("GAP")' || AnyFailures=Yes
```

src/GAP.jl
```diff
@@ -1,5 +1,6 @@
 module GAP
 
+include(abspath(joinpath(@__DIR__,"..","deps","deps.jl")))
```
fingolfin:
```suggestion
include(abspath(joinpath(@__DIR__, "..", "deps", "deps.jl")))
```

src/GAP.jl
```diff
@@ -90,10 +90,19 @@ run_it = function(gapdir::String)
...
                        , "-T", "-r", "-A", "--nointeract"
+                       , "-l", "$(EXTRA_GAP_ROOT);"
 #                      , "-m", "512m" ], [""] )
-                       , "-m", "1000m" ], [""] )
+                       , "-m", "1000m" ], [""], error_handler_func )
     gap_is_initialized = true
 end
 
+function __init__()
+    error_handler_func = @cfunction(error_handler,Cvoid,())
```
fingolfin:
```suggestion
    error_handler_func = @cfunction(error_handler, Cvoid, ())
```

.travis.yml
sebasguts: Actually, we could choose `/any/dir/that/is/not/${TRAVIS_BUILD_DIR}` here, because at the beginning (before the first `Pkg.add`) it just needs to be empty. I have choosen this one, but I can do a more custom one if you prefer.

.travis.yml
sebasguts: Yep, would also be possible.

.travis.yml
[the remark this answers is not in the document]
sebasguts: No. We just do not download the GAP via Julia, but use the GAP installed by our script.

This way, it is easier to install `profiling` and `io` afterwards.

deps/build.jl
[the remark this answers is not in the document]
sebasguts: Okay, then let's change this once we have the package manager.

etc/install_julia_gap.sh
sebasguts: I will do something completely different. I will delete this script (as it is not super useful anyway anymore) by just the GAP install script and rename it to `travis_install_gap.sh`

.travis.yml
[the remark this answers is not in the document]
fingolfin: I don't understand. Why is it easier to install profiling and io this way, resp. what's the problem with installing them in the GAP installed by Julia? Because you don't know the directory? Well, seeing that we already have the `GAP_INSTALL_PACKAGES`, couldn't it also have a "debug" mode where the `build.jl` installs io and profiling?

Also, does this mean building the Julia package is not actually tested? It seems to me that we should have one Travis config using our own GAP, and one using the GAP installed via Julia, to make sure both variants keep working... That said, it is fine to implement this *after* this PR has been merged; I am mostly bringing it up here to understand what the issues are you are working around here.

etc/travis_test.sh
sebasguts: This is basically a leftover. When I tried to just do `Pkg.build()` with the changed travis `JULIA_PROJECT`, Nemo was added to the `Project.toml`. I can remove this again.

deps/build.jl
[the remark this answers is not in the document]
fingolfin: Personally I am skeptical whether on the long term, having a GAPJulia meta package is the right approach or even needed: While right now it of course makes perfect sense to develop JuliaInterface and JuliaExperimental in lock step, I thought that in the future, we'd strive to have a stable JuliaInterface, providing a stable API and ABI; then JuliaExperimental can move out here; or rather, the various separate packages I'd imagine might replace it: NemoInterface, HeckeInterface, "SingularInterface" (hm, name clash). Of course those then could be in a meta package "OSCAR" or "OSCARInterface"... 

Anyway, while I am not thrilled by the current structure, I think it is perfectly fine to use it as you put it in this PR for now, and then discuss whether to change it further or not again at some point in the future; I think changing this aspect would be much simpler than the big change in this PR, so it's not worthwhile to delay this big PR for it.

.travis.yml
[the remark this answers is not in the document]
sebasguts: Exactly the fact that we do not know in which directory the GAP that Julia compiles is compiled to. I guess the easiest is the debug flag you propose. Should we do that?

pkg/GAPJulia/JuliaInterface/read.g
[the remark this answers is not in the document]
fingolfin: I dislike the `generated_path.gi` for various reasons; one is that it makes it harder to relocate the files. Which matters if one wants to package these things into a binary installer for users.

So I'd really prefer approach 1 or 2. I think 1 would be better, as it requires less and simpler C code, and might be handy for other things, too.

deps/build.jl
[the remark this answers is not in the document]
sebasguts: Okay, then let's leave this structure as it is now. I think both structures have pro's and con's, so it is really not a decision we have to make right now.

pkg/GAPJulia/JuliaInterface/read.g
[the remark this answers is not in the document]
fingolfin: OK, but I just see that `dirs_gap_jl` is not used at all anymore, so the even simpler solution is to remove all of this :)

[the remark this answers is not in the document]
sebasguts: Yep, I already removed the `generated_path.gi` completely.

.travis.yml
```diff
@@ -22,23 +24,23 @@ branches:
     - master
 
 before_script:
+  - unset JULIA_PROJECT
```
fingolfin: Do we perhaps also need this `unset` inside of `script:` and/or `after_script`? (Since tests on Travis seem to have passed: apparently not?!)

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -6,23 +6,22 @@
...
 
-dirs:= DirectoriesPackageLibrary( "JuliaInterface", "julia" );
-
-JuliaIncludeFile( Filename( dirs, "gaptypes.jl" ) );
-
-dirs:= DirectoriesPackageLibrary( "JuliaInterface", "../LibGAP.jl/src" );
-
 if not IsBound( JULIAINTERNAL_LOADED_FROM_JULIA ) then
-    JuliaEvalString( Concatenation( "Base.include( Main,\"", Filename( dirs, "initialization.jl" ), "\")" ) );
+    JuliaEvalString( "__IS_LOADED_FROM_GAP = true" );
```
fingolfin: Perhaps rename `__IS_LOADED_FROM_GAP` to say `__JULIAINTERNAL_LOADED_FROM_GAP` to match `JULIAINTERNAL_LOADED_FROM_JULIA` (and then perhaps also rename the latter to `__JULIAINTERNAL_LOADED_FROM_JULIA` ?

Also, while in GAP it is perfectly normal to pollute the global namespace with variables (sadly, and due to the lack of proper namespaces), I suspect this is less well received on the Julia side. So I wonder if we can pull this off by some other means? E.g. could we set this value inside the `__JULIAGAPMODULE`? (But this is very important, and can easily be done (or not) after this PR is merged.

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -6,23 +6,22 @@
...
-JuliaEvalString( Concatenation( "Base.include( Main.GAP,\"", Filename( dirs, "libgap.jl" ), "\")" ) );
+dirs_julia := DirectoriesPackageLibrary( "JuliaInterface", "julia" );
+
+JuliaIncludeFile( Filename( dirs_julia, "gaptypes.jl" ) );
+
+JuliaEvalString( Concatenation( "__JULIAGAPMODULE.include( \"", Filename( dirs_julia, "libgap.jl" ), "\")" ) );
 
 _JULIAINTERFACE_INTERNAL_INIT();
 
+Julia!.storage.GAP := _WrapJuliaModule( "GAP", _JuliaGetGlobalVariable( "__JULIAGAPMODULE" ) );
```
fingolfin: Could you explain what this line is about? Why do we need it, and why does it expose `__JULIAGAPMODULE` under the name "GAP", instead of putting the module "GAP" there?
Perhaps a comment could then be added to the file with that explanation?

.travis.yml
sebasguts: This seems to be valid for all parts of the tests. I only had one additional export before, too, also in `before_script`, so this seems to be valid for all parts of the test.

pkg/GAPJulia/JuliaInterface/read.g
sebasguts: Nope, this has to stay in `Main`, because it is older than the creation of the `GAP` module, as we read `GAP.jl` only afterwards.

`I will rename it to `__GAPINTERNAL_LOADED_FROM_GAP`, as opposed to `__JULIAINTERNAL_LOADED_FROM_JULIA`, to also emphasize in which runtime those vars exist.

sebasguts: As I think it is weird to have a `__JULIAINTERNAL` variable in julia itself.

pkg/GAPJulia/JuliaInterface/read.g
sebasguts: I added a comment.

sebasguts:
I addressed all your comments for now. We can just squash all the commits when merging it, as this will be easier to rewrite the history.

sebasguts:
We should tag this commit, btw, as it will be quite hard to bisect with this commit in the middle.

pkg/GAPJulia/JuliaInterface/read.g
fingolfin: Indeed, you even added two comments. That helped a lot, thanks!

pkg/GAPJulia/JuliaInterface/read.g
```diff
@@ -0,0 +1,29 @@
...
+dirs_julia := DirectoriesPackageLibrary( "JuliaInterface", "julia" );
+
+JuliaIncludeFile( Filename( dirs_julia, "gaptypes.jl" ) );
+
+JuliaEvalString( Concatenation( "__JULIAGAPMODULE.include( \"", Filename( dirs_julia, "libgap.jl" ), "\")" ) );
+
+_JULIAINTERFACE_INTERNAL_INIT();
+
+## The GAP module is also bound to the variable __JULIAGAPMODULE,
+## to prevent Name clashes when accessing it before it is completely initialized.
```
fingolfin:
```suggestion
## to prevent name clashes when accessing it before it is completely initialized.
```

PR #211, opened 2019-03-08
Test normal and precompiled gap

sebasguts:


fingolfin:
Please rebase so that CI runs again and we get proper coverage.

sebasguts:
Done.

.travis.yml
```diff
@@ -4,10 +4,18 @@ julia:
...
- global:
-   - GAPROOT=$TRAVIS_BUILD_DIR/gap
-   - MAKEFLAGS="-j4"
-   - JULIA_TRACK_COVERAGE=yes
+  - GAP=$HOME/.julia/gap.sh
+    GAP_INSTALL_PACKAGES=debug
+    MAKEFLAGS="-j4"
+    JULIA_TRACK_COVERAGE=yes
+    CFLAGS+=--coverage
+    LDFLAGS+=--coverage
```
fingolfin: Huh, are you sure += is allowed here?!? I kinda doubt it

fingolfin: Looking at the build logs, it definitely doesn't work.

Also, why not keep the common env vars in an "env: global:" ? That's what it's for, no?

sebasguts: I will try. It did not work for me the last time I tried. However, it should also work without the `+`.

sebasguts: Lets see how it works now.

.travis.yml
```diff
@@ -28,11 +34,11 @@ before_script:
   - printenv | sort
   - git show --pretty=fuller -s
   - mkdir -p coverage
-  - etc/travis_install_gap.sh
+  - if [[ -v GAPROOT ]]; then etc/travis_install_gap.sh; fi
```
fingolfin: More idiomatic would be
```suggestion
  - [[ -v GAPROOT ]] && etc/travis_install_gap.sh
```

.travis.yml
```diff
@@ -28,11 +34,11 @@ before_script:
...
     ## Add the package so we can use it later
   - julia -e 'import Pkg ; Pkg.add(Pkg.PackageSpec(url=pwd()))'
   - julia -e 'import Pkg ; Pkg.add("Nemo")'
-  - etc/travis_build.sh
+  - if [[ -v GAPROOT ]]; then etc/travis_build.sh; fi
```
fingolfin:
```suggestion
  - [[ -v GAPROOT ]] && etc/travis_build.sh
```

sebasguts:
Argh, now I merged it, but we still have incorrect displayed data, right?

fingolfin:
Well, we'll just need to make another PR with coverage fixes :-). Besides missing some .c file, this is also missing all files from GAP.jl itself, i.e., from the /src/ dir.

sebasguts:
Oha, yeah, I also have some suspicion where that comes from. I will make another PR.

PR #213, opened 2019-03-10
Repair coverage (as Pkg gets installed to $HOME/.julia)

sebasguts:


etc/gather_coverage.jl
```diff
@@ -6,9 +6,7 @@ using Coverage
 import JSON
 
 # gather coverage
-data = Codecov.process_folder("pkg/GAPJulia/JuliaInterface")
-append!(data, Codecov.process_folder("pkg/GAPJulia/JuliaExperimental"))
-append!(data, Codecov.process_folder())
+data = Codecov.process_folder("$(ENV["HOME"])")
```
fingolfin: That string quoting looks wrong

sebasguts: Nope, that string quoting is correct in Julia.

sebasguts:
I will restart travis, as there was a hickup in downloading the GAP required packages. But one of the Jobs was successful, and the coverage looks much better now.

PR #214, opened 2019-03-11
Delete coverage files AFTER coverage is collected

sebasguts:


fingolfin:
For some reason, Travis and AppVeyor are not building this PR at all?

fingolfin:
OK, I figured out why: recent change suggested by me (to use `[[` without `if`) broke our `.appveyor.yml`, it was no longer valid YAML. This is fixed in PR #215, but I can also make a standalone fix if desired.

sebasguts:
Obsolete with #219

PR #219, opened 2019-03-11
Scan less dirs for Julia codecoverage

sebasguts:
and do not delete coverage files before collecting


sebasguts:
Now part of #218

PR #224, opened 2019-04-09
Add a default `gap_to_julia` conversion for Julia objects

sebasguts:


pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -9,11 +9,20 @@ The parameter `recursion_dict` is meant to preseve egality
 of converted objects and should never be given by the user.
 """
 
+## Default for conversion:
+## Do not convert Julia types, as long as type matches
```
fingolfin: It would be nice if the comment also explained why this is useful, e.g. by describing an example use case.

sebasguts:
Sorry, forgot to push my changes yesterday. Test case and comments are there now . . .

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
fingolfin: I still find this confusing... We *do* convert them, just to themselves. Or, for yet another perspective: we do not convert them, but still let them through, while other objects are explicitly rejected.
```suggestion
## Base case fro conversion (least specialized method): Allow converting any
## Julia object x to type T, provided that the type of x is a subtype of T;
## otherwise, explicitly reject the conversion.
```

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -9,11 +9,23 @@ The parameter `recursion_dict` is meant to preseve egality
...
+## If you have a GAP list of Julia Tuples, this function
+## enables conversion of that list to a list of type
+## Array{Tuple,1}.
```
fingolfin: How about this instead:
```suggestion
## As an example why this is useful, suppose you have a GAP list x (i.e., an
## object of type MPtr) containing a bunch of Julia tuples. Then this method
## enables conversion of that list to a Julia array of type Array{Tuple,1},
## like this:
##    gap_to_julia(Array{Tuple{Int64},1},xx)
## This works because first the gap_to_julia method with signature
## (::Type{Array{T,1}}, :: MPtr) is invoked, with T = Tuple{Int64}; this then
## invokes gap_to_julia recursively with signature (::Tuple{Int64},::Any),
## which ends up selecting the method below.
```

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -9,11 +9,23 @@ The parameter `recursion_dict` is meant to preseve egality
...
+function gap_to_julia(t::T, x::Any) where T <: Type
+    if ! (typeof(x) <: t)
+        throw(MethodError("Wrong type: expected " * string(t) * ","))
+    end
+    return x
+end
+
 ## Default
 gap_to_julia(::Type{GAPInputType},x::GAPInputType) = x
```
fingolfin: Is this method still needed now? It is a special case of the above, isn't it? (Might still be needed due to its slightly higher method priority, I guess?)

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -9,11 +9,23 @@ The parameter `recursion_dict` is meant to preseve egality
...
 gap_to_julia(::Type{Any},         x::GAPInputType) = gap_to_julia(x)
 gap_to_julia(::Type{Any},         x::Any         ) = x
-gap_to_julia(::Any,               x::Nothing     ) = nothing
+gap_to_julia(::T,                 x::Nothing     ) where T <: Type = nothing
 gap_to_julia(::Type{Any},         x::Nothing     ) = nothing
```
fingolfin: Are the above three methods still needed now?They are special cases of the new rule, aren't they? (Might still be needed due to its slightly higher method priority, I guess?)

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -9,11 +9,23 @@ The parameter `recursion_dict` is meant to preseve egality
...
+function gap_to_julia(t::T, x::Any) where T <: Type
+    if ! (typeof(x) <: t)
+        throw(MethodError("Wrong type: expected " * string(t) * ","))
+    end
+    return x
+end
+
 ## Default
 gap_to_julia(::Type{GAPInputType},x::GAPInputType) = x
 gap_to_julia(::Type{Any},         x::GAPInputType) = gap_to_julia(x)
```
fingolfin: I was confused a bit by this rule; it really is different from the rules above and below it, as it performs some kind of "default conversion" for objects of type `GAPInputType`; I think it should be separated and a comment added. Of course it doesn't have to be in this PR, but I am pointing this out now as I had a fresh look at this conversion code, and think we really need to document it a bit better.

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
sebasguts: Yes, there are still needed.
Removing any one or two of them would either change the behaviour, or make the choice ambigious.
 
In principle, they are subsets of each other, but Julia wants its choice of methods to be unique.

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
sebasguts: Indeed, that one is not necessary anymore :).

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
sebasguts: I added that suggestion

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
sebasguts: Yeah, it indeed adds a default conversion (specially on `MPtr`) if the target type is `Any`. We need indeed to document it better, though.

sebasguts:
Thank you :)

PR #225, opened 2019-04-09
Added GAP REPL mode - WIP

sebasguts:


fingolfin:
An alternate implementation (using `}` instead of `$` as start key, but that's obviously easy to tweak):
```julia
using REPLMaker, GAP
initrepl(GAP.evalstr, prompt_text="gap> ", prompt_color = :blue, start_key='}', mode_name="GAP_mode")
```
Of course this is also quite limited and differs in numerous ways from an actual GAP prompt. And in the meantime, we have `GAP.prompt()`, which gives us a genuine GAP prompt (but not a true Julia REPL mode, which also would have advantages...).

I have an old incomplete GAP PR which I ought to dig out that would make it possible to implement a "proper" REPL mode (and also would allow to refactor the GAP REPL). I'll link it here once I submitted it.

PR #227, opened 2019-04-12
Add Julia 1.2 to travis script

sebasguts:


sebasguts:
@fingolfin Yeah, I unfortunately saw this myself. I will see if we can make it work somehow. I guess compiling Julia on Travis is not an option . . .

PR #228, opened 2019-04-16
Added conversion for two-dimensional arrays

sebasguts:


pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -202,6 +202,56 @@ function gap_to_julia( ::Type{Array{Union{Nothing,T},1}}, obj :: MPtr, recursion
...
+    end
+    if haskey(recursion_dict, obj)
+        return recursion_dict[obj]
+    end
+    len_list_outer = length(obj)
+    len_list_inner = len_list_outer == 0 ? 0 : length(obj[1])
+    new_array = Array{T,2}( undef, len_list_outer, len_list_inner )
+    recursion_dict[obj] = new_array
+    for i in 1:len_list_outer
+        for j in 1:len_list_outer
```
ThomasBreuer: `1:len_list_inner`

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -202,6 +202,56 @@ function gap_to_julia( ::Type{Array{Union{Nothing,T},1}}, obj :: MPtr, recursion
...
+    end
+    if haskey(recursion_dict, obj)
+        return recursion_dict[obj]
+    end
+    len_list_outer = length(obj)
+    len_list_inner = len_list_outer == 0 ? 0 : length(obj[1])
+    new_array = Array{Union{Nothing,T},2}( undef, len_list_outer, len_list_inner )
+    recursion_dict[obj] = new_array
+    for i in 1:len_list_outer
+        for j in 1:len_list_outer
```
ThomasBreuer: `1:len_list_inner`

fingolfin:
> as far as I see, equal subobjects on the source side are turned into identical objects on the target side.

What makes you say that? This would be a bug in general. Do you have some specific examples for this?

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
sebasguts: thank you for spotting this, fixed now

sebasguts:
I think this is a misunderstanding. Identical objects on the source side are turned into identical objects (on purpose). This is done by using `IdDict()`, which compares by `===`, not `==`, for caching already converted objects.

ThomasBreuer:
Thanks for the explanation.

PR #234, opened 2019-05-08
First version of scoping

sebasguts:
Which leaks memory and is generally a bad choice. We should just use "import/export/using" instead:

```
julia> using GAP
Adding path /home/sebastian/Software/GAPJulia/gap/.libs to DL_LOAD_PATH
 ┌───────┐   GAP 4.dev of today
 └───────┘   Architecture: x86_64-pc-linux-gnu-julia64-kv6
 Configuration:  gmp 6.1.2, Julia GC, Julia 1.1.0, readline
 Packages:   GAPDoc 1.6.2.dev, PrimGrp 3.3.1, SmallGrp 1.3, TransGrp 2.0.4

julia> import Polymake
polymake version 3.4
Copyright (c) 1997-2019
Ewgenij Gawrilow, Michael Joswig (TU Berlin)
https://polymake.org

This is free software licensed under GPL; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.


julia> GAP.@gap_scope Polymake begin
           x = Polytope.cube(4)
       end

julia> x
type: Polytope<Rational>
description: cube of dimension 4
```

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/julia/gap.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/gap.jl
@@ -154,3 +154,38 @@ end
 
 export LoadPackageAndExposeGlobals
 
+macro gap_scope(tuple...)
+    if length(tuple) != 2
+        throw(ArgumentError("must have two arguments"))
+    end
+
+    rand_module = Symbol( "mockmodule"*string( rand( UInt64 ) ) )
+    
+    module_name = tuple[1]
+    eval_expr = tuple[2]
+    
+    Base.MainInclude.eval(:(module $(rand_module) using Main.$(module_name) end))
+    mockmodule = Base.MainInclude.eval(:($(rand_module)))
+
+    real_module = Base.MainInclude.eval(:($module_name))
+
+    for i in names(real_module,all=true)
+        if ( !isdefined(mockmodule,i)) && isdefined(real_module, i)
+            Base.invokelatest( mockmodule.eval, :($(i) = Main.$(module_name).$(i) ) ) 
+        end
+    end
+
+    nameslist = names(mockmodule,all=true)
+
+    Base.invokelatest( mockmodule.eval, tuple[2] )
+
+    for i in names(mockmodule,all=true)
+        if (! (i in nameslist)) && isdefined(mockmodule, i)
+            Base.MainInclude.eval(:($(i) = $(rand_module).$(i)))
+            Base.invokelatest( mockmodule.eval, :( $(i) = nothing ) )
+        end
+    end
+
+end
+
+export @gap_scope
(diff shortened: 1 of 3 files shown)


sebasguts:
@mohamed-barakat please have a look at this PR. Is that the scoping you wanted? Here is an example:
```
julia> LoadPackageAndExposeGlobals( "CAP","CAP" )

julia> @gap_scope CAP begin
           x = CapCat
       end

julia> x
GAP: Cat
```

However, I would suggest just exporting the names from `CAP` into `Main`, which works like this:

```
julia> LoadPackageAndExposeGlobals( "CAP", Main )

julia> CapCat
GAP: Cat
```

This however should only be used in interactive computations. For programming, you could also do explicit imports:

```
julia> LoadPackageAndExposeGlobals( "CAP", "CAP" )

julia> import Main.CAP: CapCat

julia> CapCat
GAP: Cat
```

mohamed-barakat:
Yes, this is exactly what I wanted 👍 Is it stable enough to use it in Berlin tomorrow? 😉

sebasguts:
This should work completely stable, yes.

However, for tomorrow we do not need this. Writing CAP is also pretty lightweight.

sebasguts:
After thinking about it, this is done in a bad way anyway. I will provide a new version of it.

sebasguts:
I thought about this functionality for quite some time now, and overall, it is a bad idea, as it cannot made in any sense consistent. The main idea is to provide a macro that lets use use functions that are in a gap package module created via `LoadPackageAndExposeGlobals` without prefixing them with the module name for some time.

```
julia> LoadPackageAndExposeGlobals( "CAP","CAP" )

julia> @gap_scope CAP begin
           x = CapCat
       end

julia> x
GAP: Cat
```

However, to achieve this, one has to do the execution in a module other than `Main`, either the package module `CAP` or in a mock one specifically created for this purpose. The latter produces a major overhead, so I dropped this idea.

For the second possibility, i.e., to execute directly in the `CAP` module above, one can do two things: Remove the newly created identifiers when handing them back to `Main`, or leave them in the CAP module. The first one has the disadvantage that for the next time this `@gap_scope` macro is used, the identifiers created in the first run are not avialiable anymore, i.e.
```
julia> @gap_scope CAP begin
           y = x
       end
```
would not work. What would work is
julia> @gap_scope CAP begin
           y = Main.x
       end
```
which I consider even worse.

For the second option, the `CAP` module produces a memory leak, as the objects in there, even if their identifier in Main got newly assigned, will not be deleted. So this is also not an option.

The best (and already possible option) is to simly use
```
LoadPackageAndExposeGlobals( "CAP", Main )
```
which just exports all identifiers from the CAP package into the main module, so they can be used unqualified. However, this should probably just be used in interactive usage. I need to think about a way to make this possible in "real" julia modules.

PR #238, opened 2019-05-16
Added options to GAP functions as keyword arguments

sebasguts:
It would be cool to have an example. Can you point me to a GAP function that does something with options?

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
@@ -125,10 +125,30 @@ end
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-(func::MPtr)(args...) = call_gap_func(func,args...)
+(func::MPtr)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
 
-function call_gap_func(func::MPtr,args...)
-    return ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+function call_gap_func(func::MPtr, args...; kwargs...)
+    global Globals
+    options = false
+    if length(kwargs) > 0
+        kwargs_dict = Dict(kwargs)
+        kwargs_rec = julia_to_gap(kwargs_dict)
+        Globals.PushOptions(kwargs_rec)
+        options = true
+    end
+    result = nothing
+    try
+        result = ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+    catch e
+        if options
+            Globals.ResetOptionsStack()
+        end
+        rethrow(e)
+    end
+    if options
+        Globals.PopOptions()
+    end
+    return result
 end
 
 struct GlobalsType
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -50,4 +50,13 @@ end
 
     @test GAP.Globals.IdFunc(2^62) isa Int64
 
+    x = GAP.Globals.Indeterminate(GAP.Globals.Rationals)
+    f = x^4+1
+    i1 = GAP.Globals.IdealDecompositionsOfPolynomial(f)
+    i2 = GAP.Globals.IdealDecompositionsOfPolynomial(f, onlyone = true)
+    @test i1 != i2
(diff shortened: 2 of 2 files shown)


ThomasBreuer:
Here is an example where options change the output.

```
gap> x:= Indeterminate( Rationals );;
gap> f:= x^4 + 1;;
gap> IdealDecompositionsOfPolynomial( f );
[ [ x_1^2-2, x_1^3-x_1 ], [ x_1^2+1, x_1^2 ], [ x_1^2+2, x_1^3+x_1 ] ]
gap> IdealDecompositionsOfPolynomial( f : onlyone );
[ [ x_1^2-2, x_1^3-x_1 ] ]
```

sebasguts:
Thank you very much. This works.

test/basics.jl
```diff
@@ -50,4 +50,10 @@ end
 
     @test GAP.Globals.IdFunc(2^62) isa Int64
 
+    x = GAP.Globals.Indeterminate( GAP.Globals.Rationals )
+    f = x^4+1
+    i1 = GAP.Globals.IdealDecompositionsOfPolynomial( f )
+    i2 = GAP.Globals.IdealDecompositionsOfPolynomial( f , onlyone = true)
```
fingolfin:
```suggestion
    i2 = GAP.Globals.IdealDecompositionsOfPolynomial( f , onlyone = true )
```
... makes this change locally consistent. But of course it is inconsistent with the surrounding code, which does not put spaces around parenthesis like that...

I wonder whether there is a usable code formatter for Julia? Last time I looked, I couldn't find one.

pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
```diff
@@ -125,10 +125,22 @@ end
...
+    options = false
+    if length(kwargs) > 0
+        kwargs_dict = Dict(kwargs)
+        kwargs_rec = julia_to_gap(kwargs_dict)
+        Globals.PushOptions(kwargs_rec)
+        options = true
+    end
+    result = ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+    if options
+        Globals.PopOptions()
```
fingolfin: Perhaps there should also be a test involving options which runs into a GAP error, to see if the options stack is correctly cleared in this case, too?

test/basics.jl
fingolfin: Perhaps add a check that `GAP.Globals.OptionsStack` is empty at this point?

sebasguts: Will add that.
I checked for a formatter, and was not able to find one, unfortunately . . .

pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
sebasguts: I added one, and of course discovered that the stack was not reset. Now, I reset the stack.

pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
```diff
@@ -125,10 +125,30 @@ end
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-(func::MPtr)(args...) = call_gap_func(func,args...)
-
-function call_gap_func(func::MPtr,args...)
-    return ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+(func::MPtr)(args...; kwargs...) = call_gap_func(func,args...; kwargs...)
```
fingolfin:
```suggestion
(func::MPtr)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
```

pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
```diff
@@ -125,10 +125,30 @@ end
...
+
+function call_gap_func(func::MPtr,args...; kwargs...)
```
fingolfin:
```suggestion
function call_gap_func(func::MPtr, args...; kwargs...)
```

sebasguts:
Thx. I applied them :)

PR #239, opened 2019-05-16
Added g_str macro to allow creating GAP strings more conveniently

sebasguts:


pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -170,3 +170,9 @@ function Display(x::MPtr)
         error("variable was not correctly evaluated")
     end
 end
+
+macro g_str(str)
```
fingolfin: Can you add a documentation comment?

PR #243, opened 2019-05-20
added more conversions for homalg

mohamed-barakat:
- 0 - x = Julia(0) - GAP(x) -> GAP(0) - GAP(x)
- R * "x" = GAP(R) * Julia("x") -> GAP(R) * GAP("x")
- R["x"] = GAP(R)[Julia("x")] -> GAP(R)[GAP("x")]

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
@@ -125,10 +125,22 @@ end
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-(func::MPtr)(args...) = call_gap_func(func,args...)
+(func::MPtr)(args...; kwargs...) = call_gap_func(func,args...; kwargs...)
 
-function call_gap_func(func::MPtr,args...)
-    return ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+function call_gap_func(func::MPtr,args...; kwargs...)
+    global Globals
+    options = false
+    if length(kwargs) > 0
+        kwargs_dict = Dict(kwargs)
+        kwargs_rec = julia_to_gap(kwargs_dict)
+        Globals.PushOptions(kwargs_rec)
+        options = true
+    end
+    result = ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+    if options
+        Globals.PopOptions()
+    end
+    return result
 end
 
 struct GlobalsType
--- a/pkg/GAPJulia/JuliaInterface/julia/gap.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/gap.jl
@@ -44,6 +44,9 @@ Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(
 Base.zero(x::GAPInputType_internal) = Globals.ZERO(x)
 Base.one(x::GAPInputType_internal) = Globals.ONE(x)
 Base.:-(x::GAPInputType_internal) = Globals.AINV(x)
+Base.:-(x::Int64,y::GAP.MPtr) = GAP.julia_to_gap(x)-y
+Base.:*(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
+Base.getindex(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
 
 #
 typecombinations = ((:GAPInputType_internal,:GAPInputType_internal),


pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -44,6 +44,9 @@ Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(
 Base.zero(x::GAPInputType_internal) = Globals.ZERO(x)
 Base.one(x::GAPInputType_internal) = Globals.ONE(x)
 Base.:-(x::GAPInputType_internal) = Globals.AINV(x)
+Base.:-(x::Int64,y::GAP.MPtr) = GAP.julia_to_gap(x)-y
+Base.:*(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
+Base.getindex(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
```
sebasguts: This `getindex` function is not general, cause it will prevent us to use `[]` with Julia strings. You should remove it fromthis  and implement `[]` on the GAP side.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -44,6 +44,9 @@ Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(
 Base.zero(x::GAPInputType_internal) = Globals.ZERO(x)
 Base.one(x::GAPInputType_internal) = Globals.ONE(x)
 Base.:-(x::GAPInputType_internal) = Globals.AINV(x)
+Base.:-(x::Int64,y::GAP.MPtr) = GAP.julia_to_gap(x)-y
```
sebasguts: I am actually puzzled this is necessary. Arithmetic between `Int64` and `MPtr` should already be defined. If not, we should add it there, not as a special case.
Moreover, `julia_to_gap` is a no-op on all integers smaller than 2^60, so are you sure this is necessary?

pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
```diff
@@ -125,10 +125,22 @@ end
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-(func::MPtr)(args...) = call_gap_func(func,args...)
-
-function call_gap_func(func::MPtr,args...)
-    return ccall(:call_gap_func, Any, (MPtr, Any), func, args)
+(func::MPtr)(args...; kwargs...) = call_gap_func(func,args...; kwargs...)
```
sebasguts: Please remove my other commit from this PR.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -44,6 +44,9 @@ Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(
...
+Base.:*(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
```
sebasguts: Maybe we should try to implement arithmetic by automatically trying to call `julia_to_gap`? Otherwise, this is a very special case, too.

pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
mohamed-barakat: OK.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
mohamed-barakat: OK.

mohamed-barakat: I have this on the GAP side

https://github.com/homalg-project/homalg_project/blob/master/MatricesForHomalg/gap/Julia.gi#L15

but it is apparently not enough. But I see, that the RHS of the assignment is far too special.

sebasguts: Maybe the solution is adding a function like this one: https://github.com/oscar-system/GAP.jl/blob/master/pkg/GAPJulia/JuliaInterface/julia/gap.jl#L23 but which takes arbitrary second arguments, or `String`s for example

pkg/GAPJulia/JuliaInterface/julia/gap.jl
sebasguts: https://github.com/oscar-system/GAP.jl/blob/master/pkg/GAPJulia/JuliaInterface/julia/gap.jl#L51 This line plus the `-` below in the loop should actually be sufficient to implement this function. Do you call it on integers larger than 2^60?

pkg/GAPJulia/JuliaInterface/julia/gap.jl
sebasguts: https://github.com/oscar-system/GAP.jl/blob/master/pkg/GAPJulia/JuliaInterface/julia/gap.jl#L51 This line plus the `-` below in the loop should actually be sufficient to implement this function. Do you call it on integers larger than 2^60?

mohamed-barakat: My code seems to be working without this now, I don't remember why I needed this conversion.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
mohamed-barakat:
```
Base.getindex(x::GAP.MPtr,y::String) = Globals.ELM_LIST(x,GAP.julia_to_gap(y))
```
If it is OK for you I can use it.

mohamed-barakat:
Sorry, I deleted the branch by accident.

mohamed-barakat:
Reopened as new PR #244.

PR #244, opened 2019-05-20
added more conversions for homalg

mohamed-barakat:
- R * "x" = GAP(R) * Julia("x") -> GAP(R) * GAP("x")
- R["x"] = GAP(R)[Julia("x")] -> GAP(R)[GAP("x")]

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/julia/gap.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/gap.jl
@@ -44,6 +44,8 @@ Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(
 Base.zero(x::GAPInputType_internal) = Globals.ZERO(x)
 Base.one(x::GAPInputType_internal) = Globals.ONE(x)
 Base.:-(x::GAPInputType_internal) = Globals.AINV(x)
+Base.:*(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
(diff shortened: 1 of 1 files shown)


pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -44,6 +44,8 @@ Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(
 Base.zero(x::GAPInputType_internal) = Globals.ZERO(x)
 Base.one(x::GAPInputType_internal) = Globals.ONE(x)
 Base.:-(x::GAPInputType_internal) = Globals.AINV(x)
+Base.:*(x::GAP.MPtr,y::String) = x*GAP.julia_to_gap(y)
+Base.getindex(x::GAP.MPtr,y::String) = Globals.ELM_LIST(x,GAP.julia_to_gap(y))
```
fingolfin: Hmm... That's quite specific, and would break our rule of avoiding implicit conversion unless they are free, or absolutely required.

All in all, wouldn't it make more sense to put these conversions into a dedicated `Homalg.jl`?

mohamed-barakat: Yes, maybe. I have no objection in starting a `homalg.jl` :)

mohamed-barakat:
I have now created:

https://github.com/homalg-project/homalg_project.jl

mohamed-barakat:
As discusses I renamed/made a new package:

https://github.com/homalg-project/HomalgProject.jl

PR #245, opened 2019-05-21
Some small examples to start with in README.md

sebasguts:


README.md
```diff
@@ -15,6 +15,31 @@ To install this module in Julia, use
 ] add https://github.com/oscar-system/GAP.jl
 ```
 
+## Basic usage
+
+After
```
fingolfin:
```suggestion
After entering the following in Julia,
```
This way it is clear that we are talking about Julia code, and it applies both to interactive and non-interactive usage.

README.md
```diff
@@ -15,6 +15,31 @@ To install this module in Julia, use
...
+```julia
+using GAP
+```
+one can call all GAP functions via prefixing them with `GAP.Globals.`:
```
fingolfin:
```suggestion
one may call any GAP function by prefixing its name with `GAP.Globals.`. For example:
```

README.md
```diff
@@ -15,6 +15,31 @@ To install this module in Julia, use
...
+```julia
+julia> GAP.Globals.SymmetricGroup( 3 )
+GAP: SymmetricGroup( [ 1 .. 3 ] )
+```
+The Julia types `Int64` and `Bool` are automatically converted to GAP, several others can be converted using `GAP.julia_to_gap`:
```
fingolfin:
```suggestion
The Julia types `Int64` and `Bool` are automatically converted to GAP objects.
Several others types of objects can be converted using `GAP.julia_to_gap`:
```

README.md
```diff
@@ -15,6 +15,31 @@ To install this module in Julia, use
...
+```julia
+julia> x = GAP.julia_to_gap([1,2,3])
+GAP: [ 1, 2, 3 ]
+```
+Converting back to Julia is done using `GAP.gap_to_julia`, adding the type of the resulting object
```
fingolfin:
```suggestion
Converting back to Julia is done using `GAP.gap_to_julia`. However, for this one needs to specify the desired type of the resulting object. For example, to convert the GAP list of integers we just defined back to Julia, we might do this:
```

fingolfin:
(Of course I am not asking you to write that manual here -- the change request is purely for some phrasing in you PR).

sebasguts:
I see. I applied your suggests :)

README.md
```diff
@@ -15,6 +15,32 @@ To install this module in Julia, use
...
+julia> GAP.Globals.SymmetricGroup( 3 )
+GAP: SymmetricGroup( [ 1 .. 3 ] )
+```
+The Julia types `Int64` and `Bool` are automatically converted to GAP objects.
+Several others types of objects can be converted using `GAP.julia_to_gap`:
+```julia
+julia> x = GAP.julia_to_gap([1,2,3])
+GAP: [ 1, 2, 3 ]
+```
+Converting back to Julia is done using `GAP.gap_to_julia`. However, for this one needs to specify the desired type of the resulting object. For example, to convert the GAP list of integers we just defined back to Julia, we might do this:
```
fingolfin: Why is this one long line, while lines 29/30 is wrapped?

README.md
```diff
@@ -15,6 +15,32 @@ To install this module in Julia, use
...
 
+## Basic usage
+
+After entering the following in Julia,
+```julia
+using GAP
+```
+one may call any GAP function by prefixing its name with `GAP.Globals.`. For example:
+```julia
+julia> GAP.Globals.SymmetricGroup( 3 )
```
fingolfin: I am not going to request a change for this, but I note that once more, this is formatted differently in terms of whitespace than the other example inputs :-).

sebasguts: I think because you made this line longer because of your suggestions. Anyway, fixed it.

README.md
sebasguts: I think because you made this line longer because of your suggestions ;). Anyway, fixed it.

README.md
sebasguts: Still fixed it.

PR #248, opened 2019-06-07
in show, use StringView not String

ThomasBreuer:
I think that the `show` method for GAP objects fits better to GAP's `ViewObj` than to `String`.
Ideally, `ViewString` should be called but this is not (yet) supported for enough GAP objects.
Thus `StringView` is a reasonable temporary alternative.
(Cf. https://github.com/gap-packages/JupyterKernel/pull/104.)

### DIFF
--- a/src/gap2.jl
+++ b/src/gap2.jl
@@ -1,7 +1,7 @@
 import Base: convert, getindex, setindex!, length, show
 
 function Base.show( io::IO, obj::Union{GapObj,FFE} )
-    str = Globals.String( obj )
+    str = Globals.StringView( obj )
     stri = CSTR_STRING( str )
     print(io,"GAP: $stri")
 end


ThomasBreuer:
One of the Travis tests failed, due to a change of a type name in AbstractAlgebra:
`MatSpaceElem` instead of `Mat`.
The other tests succeeded, they seem to use an older version of AbstractAlgebra.
At the moment, I see no way to get a consistent output.
(Should I better avoid showing type names in tests?)

mohamed-barakat:
> (Should I better avoid showing type names in tests?)

@ThomasBreuer: Yes, maybe this is a good idea for the moment.

ThomasBreuer:
@mohamed-barakat 
Note that @fingolfin has already disabled the tests in question, as a part of pull request #253.

For the future, I would still like to have *some* check whether a Julia object created in a specific conversion is reasonable.
If the `typeof` string is not reliable then a compromise would be to check for a supertype, using `isa`, and to hope that the names of these supertypes are stable.
(This has nothing to do with type stability, just type name stability.)

For example, a matrix `m` in AbstractAlgebra and Nemo can be detected with `isa( m, MatElem )`.
There seem to be no general functions such as `ismatrix`.

mohamed-barakat:
> I think that the `show` method for GAP objects fits better to GAP's `ViewObj` than to `String`.
> Ideally, `ViewString` should be called but this is not (yet) supported for enough GAP objects.
> Thus `StringView` is a reasonable temporary alternative.
> (Cf. [gap-packages/JupyterKernel#104](https://github.com/gap-packages/JupyterKernel/pull/104).)

I was unaware of `StringView` and for me it seems like the correct choice anyway.

mohamed-barakat:
I rebased this PR locally and it works perfectly for me.

mohamed-barakat:
May I resolve by rebasing and force-pushing?

PR #249, opened 2019-06-12
Make GAP.jl usable as a Julia module with precompilation

### LINKED ISSUE #265: Make GAPTypes.jl a dependency in the Project.toml
Otherwise you get
```julia
julia> using GAP
Adding path /home/adsd/.julia/packages/GAP/aMcJN/gap/.libs to DL_LOAD_PATH
Panic in src/julia_gc.c:825: could not import GAPTypes module into Julia
```

sebasguts:
* Move *.jl files from JuliaInterface to module itself
* Add GAPTypes.jl package as dependency
* Replaced MPtr by GapObj
* Removed unnecessary libgap.jl
* Moved MPtr call definition into init, eval into Main  to not eval into a closed module

Resolves #265


src/GAP.jl
```diff
@@ -105,8 +122,21 @@ function __init__()
     gap_module = @__MODULE__
     Base.MainInclude.eval(:(__JULIAGAPMODULE = $gap_module))
     if ! isdefined(Main, :__GAPINTERNAL_LOADED_FROM_GAP) || Main.__GAPINTERNAL_LOADED_FROM_GAP != true
-        run_it(GAPROOT, error_handler_func)
+        run_gap(GAPROOT, error_handler_func)
     end
+
+    ## FIXME: Hack because abstract types cannot be endowed with methods in 1.1
+    ## At 1.2, this should be possible and this hack can be replaced with the
```
fingolfin:
```suggestion
    ## With Julia 1.2, this will be possible and this hack could then be replaced with the
```

src/GAP.jl
```diff
@@ -105,8 +122,21 @@ function __init__()
...
+    ## correspponding line in ccalls.jl
```
fingolfin:
```suggestion
    ## corresponding line in ccalls.jl
```
Though... what *is* the "corresponding line"? Couldn't this be made explicit here, by just saying what the replacement should be...

src/GAP.jl
```diff
@@ -105,8 +122,21 @@ function __init__()
...
+    MPtr = Base.MainInclude.eval(:(ForeignGAP.MPtr))
+    Base.MainInclude.eval(:(
+        (func::$MPtr)(args...; kwargs...) = GAP.call_gap_func(func, args...; kwargs...)
+    ))
 end
 
+include( "ccalls.jl" )
+include( "gap2.jl" )
```
fingolfin: On the long run, I suggest we split up `gap2.jl` into multiple files with clearly defined content. Like `adapter.jl` containing `getindex`, `length` etc. methods (matching `adapter.gi`);  perhaps a `types.jl` with the definitions of `Obj`, `GAPInputType`, etc.

src/ccalls.jl
```diff
@@ -7,7 +7,7 @@ function GET_FROM_GAP(ptr::Ptr{Cvoid})::Any
 end
 
 function EvalStringEx( cmd :: String )
-    res = ccall( :GAP_EvalString, MPtr, 
+    res = ccall( :GAP_EvalString, Any,
```
fingolfin: Could you explain why `Any` is used here instead of `GapObj`? Perhaps to deal with cases where an FFE is returned? In that case, perhaps `GAPInputType_internal` would be more appropriate?

src/ccalls.jl
```diff
@@ -29,42 +29,42 @@ function CanAssignGlobalVariable(name::String)
              (Ptr{UInt8},), name)
 end
 
-function AssignGlobalVariable(name::String, value::MPtr)
+function AssignGlobalVariable(name::String, value::GapObj)
```
fingolfin: Shouldn't this be 
```suggestion
function AssignGlobalVariable(name::String, value::Obj)
```
so that we can also assign bools, FFEs and immediate integers? (That is of course an existing problem, not new in this PR).

Well, and then I guess we should actually test all this variants. And also similarly for a bunch of other functions in this file...

src/ccalls.jl
```diff
@@ -80,54 +80,56 @@ end
...
-    (func::MPtr)(args...)
+    (func::GapObj)(args...)
 
-> This function makes it possible to call MPtr objects as functions.
+> This function makes it possible to call GapObjs as functions.
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-(func::MPtr)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
+# This needs to be instantiated for `MPtr` and is therefore moved to the init function
```
fingolfin: Which init function is that? Could you clarify?

src/ccalls.jl
```diff
@@ -183,4 +185,4 @@ end
 
 # For backwards compatibility
 # TODO: remove this again
-GAPFuncs = Globals
+const GAPFuncs = Globals
```
fingolfin: Is there any code out there still using `GAPFuncs`? I guess @CarloSircana uses it in `CarloLibGAPTest`, but adjusting that is a simple matter of search and replace, no?

test/basics.jl
```diff
@@ -15,9 +15,6 @@
     @test x[1] == 1
     @test length(x) == 1
 
-    @test GAP.True
-    @test ! GAP.False
```
fingolfin: Why are these removed?

src/julia_to_gap.jl
```diff
@@ -150,3 +149,6 @@ julia_to_gap(func::Function) = NewJuliaFunc(func)
 
 
 ## TODO: BitArray <-> blist; ranges; ...
+
+# Overload Base.convert
+Base.convert(::Type{T}, x) where T <: GapObj = julia_to_gap(x)
```
fingolfin: This seems to go beyond what this PR claims to do, no? There are also no tests at all for this, are there?

src/GAP.jl
```diff
@@ -17,12 +17,24 @@ import Base: length, convert, finalize
 
 import Libdl
 
+using GAPTypes
+
+export GapObj
+
+#
+# This is a very hacky prototype calling libgap from julia
```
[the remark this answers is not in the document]
sebasguts: I think this is fairly old. We should probably just remove this comment at all

src/GAP.jl
sebasguts: Will do so.

src/GAP.jl
sebasguts: Definitely, yes. I am also happy about a better name suggestion right now.

src/ccalls.jl
sebasguts: Using `GapObj` there would actually do the same as `Any`, and so would object. They might do more type checking (I am not sure, I will have to check), but they all expect the C function to return a valid `jl_value_t*`, so I decided to not do it.

src/ccalls.jl
sebasguts: Yes, indeed.

src/ccalls.jl
sebasguts: Well, the one below the comment, but I can make this more specific

src/ccalls.jl
sebasguts: Well, I am not sure. We can just remove it and see if anybody complains ;)

test/basics.jl
sebasguts: Because the constants were removed, too.

They were not really needed anyway, and there is no need to even have GAPs `true` in Julia, as `true` and `false` are always converted.

The reason why they "needed" to be removed with the update was that calling GAP functions is only possible after executing the `init` function now, and the constants were defined before.

src/julia_to_gap.jl
sebasguts: No tests atm, no. I was to lazy to put the second commit in a single PR. Will do this now.

fingolfin: That separate PR with `convert` in it has not yet happened, has it? (Just wondering as I don't see it in the list of open PRs)

sebasguts:
Sorry, I was afk for the weekend. Will continue now.

deps/build.jl
```diff
@@ -16,6 +16,12 @@ println("gap_root = ", gap_root)
 println("extra_gap_root = ", extra_gap_root)
 println("install_gap = ", install_gap)
 
+println("DEPOT = ", DEPOT_PATH)
+
+println("................ENV.............")
+println(ENV)
```
fingolfin: Is this for debugging or is it meant to say here?

deps/build.jl
```diff
@@ -26,11 +32,19 @@ if install_gap
...
-    run(`git clone https://github.com/gap-system/gap`)
+    run(`git clone --depth=1 https://github.com/gap-system/gap`)
     cd("gap")
     run(`./autogen.sh`)
     run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
     run(`make -j$(Sys.CPU_THREADS)`)
+    ## Hack to make GAPTypes available in the global package environment:
+    ## For a moment, we set the Julia load path to the actual depot, i.e.,
+    ## the env GAP.jl is installed to, to install GAPTypes.jl globally
+    julia_load_path_save = ENV["JULIA_LOAD_PATH"]
```
fingolfin: Is `ENV["JULIA_LOAD_PATH"]` guaranteed (by documentation / specification) to be set at this point? If it is not, then this line could throw an exception.

deps/build.jl
```diff
@@ -26,11 +32,19 @@ if install_gap
...
+    pop!( ENV, "JULIA_LOAD_PATH" )
+    julia_executable = joinpath(julia_binary, "julia")
+    run(`$(julia_executable) -e 'import Pkg; Pkg.add("GAPTypes")'`)
```
fingolfin: Wow, that *is* a gross hack. I am also not quite sure why it is needed? Surely, GAP's `configure` also takes care of it? OK, perhaps the problem is that it inherits the `JULIA_LOAD_PATH`? But then at the very least, the comment above should explain so.

But I also wonder (very naively, so there is probably an easy answer why this won't work): couldn't we make the code in GAP more flexible, and ask it to first check if `GAP.GAPTypes` is available, then use that; if not, try to use `GAPTypes` directly?

src/GAP.jl
```diff
@@ -17,6 +17,12 @@ import Base: length, convert, finalize
 
 import Libdl
 
+using GAPTypes
```
fingolfin: Perhaps this would be more appropriate?
```suggestion
import GAPTypes: GapObj
```

deps/build.jl
sebasguts: Yeah, I am painfully aware that this is gross, and should not be done this way. On the other hand, the Julia package manager sets this environment variable, too, which is also not very nice of it.

Actually, patching GAP this way would be possible, although not for the `4.10.2` release I guess. But then we can at least remove that hack afterwards.

deps/build.jl
sebasguts: Since this build script is (or better should, nobody is keeping you from shooting yourself in the foot) only executed by the package manager, yes, I am sure this is always set.

src/GAP.jl
sebasguts: Actually, this is exactly the same, but I will change it.

deps/build.jl
sebasguts: I will leave one more comment. The problem is that for the configure script, this somehow failed to give me the right responses when we asked for flags. I have no real idea why yet. This solution is merely the result of a lot of try-and-error.

mohamed-barakat:
I get the following error:

```Julia
julia> using GAP
[ Info: Recompiling stale cache file ~/.julia/compiled/v1.1/GAP/YqHod.ji for GAP [c863536a-3901-11e9-33e7-d5cd0df7b904]
Adding path /Users/mo/.julia/dev/GAP/gap/.libs to DL_LOAD_PATH
 ┌───────┐   GAP 4.10dev-2052-gf30c1e4 of today
 └───────┘   Architecture: x86_64-apple-darwin18.2.0-julia64-kv6
 Configuration:  gmp 6.1.2, Julia GC, Julia 1.1.0, readline
 Packages:   GAPDoc 1.6.2, PrimGrp 3.3.2, SmallGrp 1.3, TransGrp 2.0.4
Error, MethodError: Cannot `convert` an object of type Main.ForeignGAP.MPtr to an object of type GAPTypes.GapObj
Closest candidates are:
  convert(::Type{T}, !Matched::T) where T at essentials.jl:154
ERROR: InitError: Error thrown by GAP
Stacktrace:
 [2] error_handler() at /Users/mo/.julia/dev/GAP/src/GAP.jl:47
 [3] initialize(::Array{String,1}, ::Array{String,1}, ::Ptr{Nothing}) at /Users/mo/.julia/dev/GAP/src/GAP.jl:67
 [4] (::getfield(GAP, Symbol("##3#4")))(::String, ::Ptr{Nothing}) at /Users/mo/.julia/dev/GAP/src/GAP.jl:97
 [5] __init__() at /Users/mo/.julia/dev/GAP/src/GAP.jl:115
 [6] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:633
 [7] _require_from_serialized(::String) at ./loading.jl:684
 [8] _require(::Base.PkgId) at ./loading.jl:967
 [9] require(::Base.PkgId) at ./loading.jl:858
 [10] require(::Module, ::Symbol) at ./loading.jl:853
during initialization of module GAP
```

sebasguts:
Please do not put bug reports in closed PR's, but create an issue instead. Anyway, will investigate.

mohamed-barakat:
I was about to do that. This is just for reference.

sebasguts:
I have to admit, I cannot reproduce this. Did you update everything? Including GAP itself?

mohamed-barakat:
As you suggest, I deleted `GAP.jl/gap` for
```Julia
pkg> build GAP
```
to work. Now `using GAP` works. Thanks.

PR #251, opened 2019-06-13
show GAP help in Julia with 'GAP.show_GAP_help' and via `?`

ThomasBreuer:
This is a first approximation.

The functions in the file `helpstring.g` arose from those
in GAP's library file `helpbase.gi`.
The idea is to extract the requested information
from the text format files of the GAP manuals,
and to collect it in a string,
also in the case where several matches are found.

Currently the following problems are known.

- When the last subsection of a section is requested
  then also the part of the next section (in the same chapter)
  until the start of its first subsection is shown;
  apparently the sections themselves have no entry numbers
  in the sense of `HELP_BOOK_HANDLER.(book.handler).HelpData`.

  (An example is `SizeScreen`.)

- Some pieces of GAP help are very long,
  an example is the entry for `Size` in the HAP package.

On the other hand, duplicate entries are filtered out,
which should perhaps also done in GAP's help system.


sebasguts:
I updated this PR and added some functionality to display the GAP help when the Julia GAP help is called.

sebasguts:
@ThomasBreuer @mohamed-barakat @fingolfin Please have a look.

I am going to rebase this PR once all the file-reordering is done.

mohamed-barakat:
```Julia
julia> GAP.show_GAP_help("HomalgMatrix")
```
works.

In the notebook it looks like this:

https://gist.github.com/mohamed-barakat/f0335ef22e4a8618dd195cca0b6ac9b4

ThomasBreuer:
@mohamed-barakat Thanks for testing this feature.
I had not intended this way of displaying GAP's help for the notebook situation,
opening the HTML help (similar to what GAP's JupyterKernel package does)
would be a better solution.

sebasguts:
I think even in the notebook it looks kinda reasonable.
With this PR, if you trigger the notebook help by pressing `shift+tab` either once, twice, or trice, it even gets displayed with colors. I would think this is reasonable.

ThomasBreuer:
@sebasguts The first lines of help output in the notebook are broken into lines in a strange way.
I will fix this.

sebasguts:
Well, just when you use that printing function. If you use `shift + tab`, it actually prints it reasonable in Jupyter (which is also the prefered way to display help in Jupyter notebooks)

mohamed-barakat:
> @mohamed-barakat Thanks for testing this feature.
> I had not intended this way of displaying GAP's help for the notebook situation,
> opening the HTML help (similar to what GAP's JupyterKernel package does)
> would be a better solution.

@ThomasBreuer: Actually I also like to have this "inline" option for presentations.

mohamed-barakat:
> @ThomasBreuer @mohamed-barakat @fingolfin Please have a look.
> 
> I am going to rebase this PR once all the file-reordering is done.

@sebasguts: The updated link

https://gist.github.com/mohamed-barakat/f0335ef22e4a8618dd195cca0b6ac9b4

shows that also ?HomalgMatrix works.

mohamed-barakat:
> @ThomasBreuer @mohamed-barakat @fingolfin Please have a look.
> 
> I am going to rebase this PR once all the file-reordering is done.

@sebasguts: The updated link

https://gist.github.com/mohamed-barakat/f0335ef22e4a8618dd195cca0b6ac9b4

shows that also ?HomalgMatrix works.

mohamed-barakat:
> Well, just when you use that printing function. If you use `shift + tab`, it actually prints it reasonable in Jupyter (which is also the prefered way to display help in Jupyter notebooks)

Also this works fine for me.

mohamed-barakat:
One last comment. I get colors in the Jupyter notebook which are not visible in the gist link.

sebasguts:
@mohamed-barakat which colors do you mean? The ones in the ouput of `show_GAP_help`?

mohamed-barakat:
> @mohamed-barakat which colors do you mean? The ones in the ouput of `show_GAP_help`?

@sebasguts: yes, although the output of `show_GAP_help` in the gist-link has no colors.

sebasguts:
Indeed. But you can see them on [nbviewer](https://nbviewer.jupyter.org/gist/mohamed-barakat/f0335ef22e4a8618dd195cca0b6ac9b4)

mohamed-barakat:
Could you please resolve the conflicts in this PR? I would like to merge it.

fingolfin:
@mohamed-barakat we'll resolve them once PR #249 is merged, not before. Also, I'd like to review this PR once, too

pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -184,3 +184,12 @@ macro g_str(str)
 end
 
 export @g_str
+
+function GAP_help_string( topic::String, onlyexact::Bool=false )
+    return GAP.gap_to_julia( GAP.Globals.HelpString(
+           GAP.julia_to_gap( topic ), onlyexact ) )
+end
+
+function show_GAP_help( topic::String, onlyexact::Bool=false )
```
fingolfin: Perhaps run VS Code autoformatter on this code section? I'd also move it into the new `help.jl`.

mohamed-barakat:
> Indeed. But you can see them on [nbviewer](https://nbviewer.jupyter.org/gist/mohamed-barakat/f0335ef22e4a8618dd195cca0b6ac9b4)

Excellent.

sebasguts:
@mohamed-barakat I am going to wait untill #249 is merged, because otherwise I would resolve conflicts twice.

sebasguts:
I will move the code over into a separate file. I would guess the comments about the GAP code are optional @fingolfin? Because I really do not want to wait with this until we have refactored the GAP help system.

sebasguts:
I addressed the comments and updated this PR

sebasguts:
I rebased the PR

fingolfin:
@sebasguts the GAP refactoring remarks as "not even optional": they were ramblings about the distant future. Don't worry about them here :-)

fingolfin:
OK, looking at coverage made me realize that we don't have any tests for this. Could we please add some? Possible both on the GAP and the Julia side?

ThomasBreuer:
@fingolfin The only tests I can think of are tests whether calling the functions in question return some strings, regardless of the contents. If this is sufficient then I can add such tests.

ThomasBreuer:
I think that the reported code coverage is wrong.

For example, the function `HELP_DESC_MATCH` is called in the tests,
but if I understand the output correctly then https://codecov.io/gh/oscar-system/GAP.jl/compare/adcfc6dd37646408e31e7c88a8583113738b0036...b7d69210cee9dac1d99bc66af609236d40ef5e7b/diff claims that the lines of the function are counted as missed.

(I have already tried to rebase the pull request, what else should I do?)

fingolfin:
@ThomasBreuer the coverage data looks correct to me. I don't see a direct call to `HELP_DESC_MATCH` in the tests either. I assume you have verified locally that it actually gets called by the tests (e.g. by inserting an `Error` or `Print` into it). My guess would be that the problem is that we do not compile the GAP manual on Travis?

ThomasBreuer:
@fingolfin thanks for your hint.
Yes, the problem is that the `.txt` and `manual.six` files of the manuals are not available.

In this situation, requests for help just reply that nothing reasonable was found,
thus I do not see a way to provide sensible tests for this pull request.

(The current code can run into errors when the manuals are missing/incomplete,
for example if a `manual.six` file is available but a `.txt` file is missing.
I am going to update the pull request.)

fingolfin:
@ThomasBreuer on `master`, we now build the manuals of `JuliaInterface` and `JuliaExperimental`. So if you rebase this, perhaps coverage will improve.

fingolfin:
As an experiment, I cherry-picked commit 6466e3de1dcb53c4007d39d169b5e6be0bde8cc7 onto this PR, to see if it has any effect on the coverage.

PR #252, opened 2019-06-14
Remove GAPInputType

fingolfin:
As a next step, we might want to rename `GAPInputType_internal` or even remove it as well (e.g. by using `Union{MPtr,FFE}` everywhere instead).

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/julia/gap.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/gap.jl
@@ -3,10 +3,7 @@ import Base: convert, getindex, setindex!, length, show
 const True = Globals.ReturnTrue()
 const False = Globals.ReturnFalse()
 
-const GAPInputType_internal = Union{MPtr,FFE}
-const GAPInputType = Union{GAPInputType_internal,Int64,Bool}
-
-const Obj = Union{GAPInputType,Nothing}
+const Obj = Union{MPtr,FFE,Int64,Bool,Nothing}
 
 function Base.show( io::IO, obj::Union{MPtr,FFE} )
     str = Globals.String( obj )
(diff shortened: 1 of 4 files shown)


sebasguts:
The patches look good to me. But is there a specific reason why we remove this definition, and then duplicate code? To me, this looks more like a complication than a simplification. Could you please elaborate your motivation for it?

We could even do the function definition of `julia_to_gap` with `Union` types, and not have the same thing 5 times.

fingolfin:
The motivation is to make the code simpler: Right now, to understand any place using `GAPInputType`, one has to remember what that is, i.e., look it up. And make sure to not confuse it with `GAPInputType_internal`, or `Obj` (and the latter in turn should not be confused with `GapObj`... as `GAP.Obj` and `GAPTypes.GapObj` are indeed two completely different types, despite the confusingly similar names).

I personally find the code with `GAPInputType` removed easier to grok; and in fact, I am tempted to remove `GAPInputType_internal` for the same reason -- note that `Union{MPtr,FFE}` is actually *shorter* than `GAPInputType_internal`. So in fact, I have now done so in a second commit (though these changes could be squashed as well).

pkg/GAPJulia/JuliaInterface/julia/julia_to_gap.jl
```diff
@@ -12,12 +12,14 @@ of input data, by converting egal data to identical objects in GAP.
...
-julia_to_gap(x::GAPInputType) = x
+julia_to_gap(x::MPtr) = x
+julia_to_gap(x::FFE) = x
+julia_to_gap(x::Bool) = x
 
 
 ## Integers
 julia_to_gap(x::Int128) = MakeObjInt(BigInt(x)) # FIXME: inefficient hack
-#julia_to_gap(x::Int64)  = x
+julia_to_gap(x::Int64)  = x
```
fingolfin: Also note how we had to comment out this case to avoid a clash; it seems much ore natural to handle it this way.

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
```diff
@@ -30,10 +30,13 @@ function gap_to_julia(t::T, x::Any) where T <: Type
...
 
 ## Default
-gap_to_julia(::Type{Any},         x::GAPInputType) = gap_to_julia(x)
-gap_to_julia(::Type{Any},         x::Any         ) = x
-gap_to_julia(::T,                 x::Nothing     ) where T <: Type = nothing
-gap_to_julia(::Type{Any},         x::Nothing     ) = nothing
+gap_to_julia(::Type{Any},         x::MPtr   ) = gap_to_julia(x)
+gap_to_julia(::Type{Any},         x::FFE    ) = x
+gap_to_julia(::Type{Any},         x::Int64  ) = x
+gap_to_julia(::Type{Any},         x::Bool   ) = x
```
fingolfin: Note that here, we also avoid the overhead of dispatching to `gap_to_julia`, and directly can write `x`. (Though I'd hope the Julia compiler would have optimized that call away anyway; but this way, also anybody reading the code, e.g. me, understands what's going on with one glance).

pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -41,16 +38,16 @@ Base.setproperty!(x::MPtr, f::Symbol, v) = Globals.ASS_REC(x, RNamObj(f), v)
 Base.setproperty!(x::MPtr, f::Union{AbstractString,Int64}, v) = Globals.ASS_REC(x, RNamObj(f), v)
 
 #
-Base.zero(x::GAPInputType_internal) = Globals.ZERO(x)
-Base.one(x::GAPInputType_internal) = Globals.ONE(x)
-Base.:-(x::GAPInputType_internal) = Globals.AINV(x)
+Base.zero(x::Union{MPtr,FFE}) = Globals.ZERO(x)
+Base.one(x::Union{MPtr,FFE}) = Globals.ONE(x)
+Base.:-(x::Union{MPtr,FFE}) = Globals.AINV(x)
```
fingolfin: It's immediately clear now that these functions only apply to MPtr and FFE, nothing else.

test/conversion.jl
```diff
@@ -1,7 +1,8 @@
 @testset "conversion from GAP" begin
 
     ## Defaults
-    @test GAP.gap_to_julia(GAP.GAPInputType, true )
```
fingolfin: Using this internal, non-exported type in a test also seems strange -- no user should ever have used `GAPInputType` like that.

pkg/GAPJulia/JuliaInterface/julia/gap_to_julia.jl
sebasguts: Agreed. The last thing I could add here is this
```suggestion
gap_to_julia(::Type{Any},         x::Union{Bool,FFE,Int64}   ) = x
```
instead of the last three lines

fingolfin: Fair enough. I'll update the PR.

fingolfin: Actually, it turns out that this rule simply was redundant, so I removed it. Please take a look. If you agree, then this PR is ready for merging.

PR #253, opened 2019-06-16
Add REPL as package dependency

fingolfin:
That should fix the Travis build failures with Julia nightlies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
   global:
     - MAKEFLAGS="-j4"
     - JULIA_TRACK_COVERAGE=yes
-    - GAPJL_DIR=${HOME}/.julia/packages/GAP
   matrix:
     - GAP=$HOME/.julia/gap.sh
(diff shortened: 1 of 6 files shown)


mohamed-barakat:
The commit looks good, but Travis seems to have failed again?

fingolfin:
Yes, more changes are needed. And in fact, I am not so sure the change in this PR is needed / necessary either. It does not really help, in any case. My next best guess comes from https://github.com/JuliaLang/PackageCompiler.jl/issues/194 where a similar error appears, and there, it was worked around by setting `JULIA_LOAD_PATH`. So I wonder if perhaps in Julia nightlies, when loading Julia from within GAP, it somehow fails to "find itself", i.e., the julia directory with the `stdlib/REPL/` subdir in it...

fingolfin:
It seems that for the nightly builds, a completely wrong version of GAP.jl is being tested, so any change I make here has no effect there. I am fiddling around in the hopes of finding a way to resolve that.

PR #257, opened 2019-06-16
JuliaInterface: move macros to macros.jl

fingolfin:
This is a step towards breaking up `gap.jl`. It'll conflict with PR #249, but I am prepared to resolve that.


pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -74,31 +74,6 @@ for (left, right) in typecombinations
...
-    GAP: SymmetricGroup( [ 1 .. 3 ] )
-    julia> @gap(SymmetricGroup)(3)
-    GAP: SymmetricGroup( [ 1 .. 3 ] )
-
-Note that the last two examples have a slight syntactical, and therefore also
-a semantical difference. The first one executes the string `SymmetricGroup(3)`
-directly inside GAP. The second example returns the function `SymmetricGroup`
-via `@gap(SymmetricGroup)`, then calls that function with the argument `3`.
-"""
-macro gap(str)
```
sebasguts: Honestly, I am not even sure we should keep this macro.

Its functionality is very limited, and it is not exactly clear to the user what works and what does not work. So it might be easier to just remove it.

Although, if you have some experience, it leads to nicer and shorter code, specially in presentations. So I am not completely sure what to do with it . . .

PR #255, opened 2019-06-16
Misc changes, preparation for PR #249

fingolfin:
In trying to understand what's going on in PR #249, I factored out some of the changes there. A few more will come later, in a separate PR.

I will now also rebase @sebasguts's PR #249 on this.

PR #258, opened 2019-06-16
GAP.AssignGlobalVariable: allow non-mptr arguments

fingolfin:
It turns out that just changing the argument type to `Obj` is not quite enough...

This PR is based on PR #255. Once #255 is merged, it can be rebased.

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
+++ b/pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
@@ -2,7 +2,7 @@
 
 import Base: getproperty, propertynames
 
-function GET_FROM_GAP(ptr::Ptr{Cvoid})::Any
+function RAW_GAP_TO_JULIA(ptr::Ptr{Cvoid})::Any
     return ccall(:julia_gap,Any,(Ptr{Cvoid},),ptr)
 end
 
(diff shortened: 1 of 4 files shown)


pkg/GAPJulia/JuliaInterface/julia/ccalls.jl
```diff
@@ -2,12 +2,16 @@
 
 import Base: getproperty, propertynames
 
-function GET_FROM_GAP(ptr::Ptr{Cvoid})::Any
+function GAP_TO_JULIA(ptr::Ptr{Cvoid})::Any
```
sebasguts: I don't know if this name is much better, as it is easy to confuse with `gap_to_julia`, even though `GAP_TO_JULIA` is merely a preparation for `gap_to_julia`.

Maybe something like `GAP_PTR_TO_JULIA`, to make even more clear what it does?

fingolfin: I simply wanted something that is less confusing, and symmetric (so I can add JULIA_TO_GAP). How about `RAW_GAP_TO_JULIA` or `GAP_TO_JULIA_INTERN` ? Also note that I would not export this function. Can somehow make it "internal" to the module to prevent outsiders from calling it? Does Julia have naming conventions to mark "internal" functions?

sebasguts: I am happy with `RAW`. Afaik, internal functions start with an `_`. But since you have exported and non-exported functions, private functions are more or less the not exported ones.

PR #259, opened 2019-06-17
Fix Travis; disable Travis caches; etc.

fingolfin:
This is the source of many troubles, and should no longer be needed,
now that Nemo has a binary builder


mohamed-barakat:
Makes a lot sense. Thanks.

fingolfin:
@mohamed-barakat right now everything is broken regarding Travis; we are sorting this out now

PR #261, opened 2019-06-17
Use ccache on Travis

fingolfin:
I am not sure whether this is actually useful, though, as we spend very little time compiling stuff (mostly GAP), and far more with downloading GAP, and running the tests.

It's a bit difficult to measure the benefit of this without merging it, though, as of course it only helps once the cache has been filled. As it is, I am not even fully sure this works as intended...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,10 @@ branches:
 before_script:
   - unset JULIA_PROJECT
   - printenv | sort
+  - which gcc
+  - which g++
+  - which cc
+  - which c++
   - git show --pretty=fuller -s
   - mkdir -p coverage
   - export CFLAGS=--coverage; export LDFLAGS=--coverage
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,10 +3,13 @@ julia:
   - 1.1
   - nightly
 
+cache: ccache
+
 env:
   global:
     - MAKEFLAGS="-j4"
     - JULIA_TRACK_COVERAGE=yes
+    - PATH="/usr/lib/ccache:$PATH"
```
sebasguts: This is not necessary, travis adds `/usr/lib/ccache` automatically to the path when ccache is enabled.
```suggestion
    - PATH="/usr/lib/ccache:$PATH"
```

fingolfin: As I understood it, it only does so with `language: c` or `language: c++`. In any case, I don't see `/usr/lib/ccache` twice in the PATH... But sure, I can try what happens if I remove that line.

sebasguts: Ah yeah, that might be a restriction. No, it is fine for me as it is.

PR #262, opened 2019-06-17
Travis: tweak how coverage is collected

fingolfin:


etc/gather_coverage.jl
```diff
@@ -6,8 +6,9 @@ using Coverage
 import JSON
 
 # gather coverage
-data = Codecov.process_folder(pwd())
-data_dict = Codecov.to_json(data)
+data = process_folder("src")
+data = append!(data, process_folder("pkg"))
```
fingolfin: Hrm, I must be missing something -- for some reason, this collects data in `pkg`, but fails to collect it in `src`...

sebasguts: I am not sure what you mean. In this branch, `src` only has `GAP.jl`, which gets collected. Also, the `.c` files get collected. So I guess it is fine.

sebasguts:
Feel free to merge it once you agree with my comment.

etc/gather_coverage.jl
fingolfin: Err, no, `src/GAP.jl` does *not* get collected, as the codecov report clearly showed, and as now also https://codecov.io/gh/oscar-system/GAP.jl shows. So merging this unfortunately was incorrect. I'll revert it.

fingolfin:
@mohamed-barakat I am afraid that merge was a bit premature.

mohamed-barakat:
I am sorry, I oversaw the file. Please feel free to revert.

sebasguts:
Yeah, I don't know what I checked . . .

PR #264, opened 2019-06-17
Added optional parameter to `LoadPackageAndExposeGlobals`

sebasguts:
which allows to choose if already defined names in the module
the variables get exported to will be overwritten. Default is `false`.

### DIFF
--- a/src/gap2.jl
+++ b/src/gap2.jl
@@ -71,7 +71,7 @@ end
 
 """
     LoadPackageAndExposeGlobals(package::String, mod::String; all_globals::Bool = false)
-    LoadPackageAndExposeGlobals(package::String, mod::Module = Main; all_globals::Bool = false)
+    LoadPackageAndExposeGlobals(package::String, mod::Module = Main; all_globals::Bool = false, overwrite::Bool = false)
 
 `LoadPackageAndExposeGlobals` loads `package` into GAP via `LoadPackage`,
 and stores all newly defined GAP globals as globals in the module `mod`. If `mod` is
@@ -85,6 +85,9 @@ If you load the package `CAP` via
 you can use CAP commands via
 
     CAP.PreCompose( a, b )
+
+If `overwrite` is true, Symbols already in the `Main` module will be overloaded.
+Be aware that this flag only works in `Main`.
 
 """
 function LoadPackageAndExposeGlobals(package::String, mod::String; all_globals::Bool = false)
@@ -103,7 +106,7 @@ function LoadPackageAndExposeGlobals(package::String, mod::String; all_globals::
     Base.invokelatest(LoadPackageAndExposeGlobals, package, mod_mod; all_globals = all_globals)
 end
 
-function LoadPackageAndExposeGlobals(package::String, mod::Module; all_globals::Bool = false)
+function LoadPackageAndExposeGlobals(package::String, mod::Module; all_globals::Bool = false, overwrite::Bool = false)
     current_gvar_list = nothing
     if !all_globals
         current_gvar_list = Globals.ShallowCopy(Globals.NamesGVars())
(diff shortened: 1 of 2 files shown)


sebasguts:
@mohamed-barakat This PR allows you to install your own `HomalgMatrix` method in `HomalgProject.jl`

mohamed-barakat:
Excellent, will test and report.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -103,7 +103,7 @@ function LoadPackageAndExposeGlobals(package::String, mod::String; all_globals::
     Base.invokelatest(LoadPackageAndExposeGlobals, package, mod_mod; all_globals = all_globals)
 end
 
-function LoadPackageAndExposeGlobals(package::String, mod::Module; all_globals::Bool = false)
+function LoadPackageAndExposeGlobals(package::String, mod::Module; all_globals::Bool = false, overwrite::Bool = false)
```
fingolfin: Can we please have documentation for this function? It's getting more and more complicated.

Also, I think it should be moved to a separate file, afterwards I think we could rename `gap.jl` to `adapter.jl` (to match `adapter.gi`). But of course that can happen after this PR resp. after PR #249, I just wanted to mention it.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
```diff
@@ -120,11 +120,13 @@ function LoadPackageAndExposeGlobals(package::String, mod::Module; all_globals::
     end
     new_symbols = gap_to_julia(Array{Symbol,1},new_gvar_list)
     for sym in new_symbols
-        try
-            mod.eval(:(
-                $(sym)=GAP.Globals.$(sym)
-            ))
-        catch
+        if overwrite || ! isdefined(mod,sym)
```
fingolfin: So the default behaviour is changed by this PR, right? As you default `overwrite` to `false`, but the old behavior was equivalent to `overwrite = true`. So any existing code using `LoadPackageAndExposeGlobals` should be updated, right?

sebasguts: Yeah, indeed. Although I would guess the code is only @mohamed-barakat s code, who specifically requested this change.

sebasguts:
I will provide some tests.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
sebasguts: I added tests.

pkg/GAPJulia/JuliaInterface/julia/gap.jl
sebasguts: The documentation is above. I personally think the function should be used at all, but it is handy to have it for using GAP from Julia, if you only want to use GAP.

test/convenience.jl
```diff
@@ -85,3 +85,21 @@ end
     @test sym5.:1 === GAP.Globals.GeneratorsOfGroup(sym5)[1]
     @test sym5.:1 === sym5."1"
 end
+
+## Need to be created outsite of test
```
fingolfin:
```suggestion
## Need to be created outside of test
```

test/convenience.jl
```diff
@@ -85,3 +85,21 @@ end
...
+    SymmetricGroup = 5
+end
+
+@testset "LoadPackageAndExposeGlobals" begin
+
+    LoadPackageAndExposeGlobals("GAPDoc", "test1")
+    @test isdefined(test1,:SymmetricGroup) == false
+
+    LoadPackageAndExposeGlobals("GAPDoc", "test2", all_globals = true)
+    @test isdefined(test2,:SymmetricGroup) == true
```
fingolfin:
```suggestion
    @test isdefined(test2, :SymmetricGroup) == true
```

test/convenience.jl
```diff
@@ -85,3 +85,21 @@ end
...
+
+## Need to be created outsite of test
+module test3 
+    SymmetricGroup = 5
+end
+
+@testset "LoadPackageAndExposeGlobals" begin
+
+    LoadPackageAndExposeGlobals("GAPDoc", "test1")
+    @test isdefined(test1,:SymmetricGroup) == false
```
fingolfin:
```suggestion
    @test isdefined(test1, :SymmetricGroup) == false
```

src/gap2.jl
```diff
@@ -86,6 +86,9 @@ you can use CAP commands via
 
     CAP.PreCompose( a, b )
 
+If `overwrite` is true, Symbols already in the `Main` module will be overloaded.
```
fingolfin:
```suggestion
If `overwrite` is true, symbols already in the `Main` module will be overloaded.
```

sebasguts:
I added your suggestions. Will merge if tests pass.

PR #267, opened 2019-06-20
In build.jl: Only clone new GAP if gap subfolder not present

sebasguts:


deps/build.jl
```diff
@@ -23,11 +23,18 @@ julia_binary = get(ENV, "JULIA_BINARY", Sys.BINDIR)
...
-    ## TODO: We currently use the GAP master branch.
-    ##       Once all issues of using GAP with the julia
-    ##       GC are resolved, we switch to a stable version.
-    run(`git clone --depth=1 https://github.com/gap-system/gap`)
-    cd("gap")
+    ## check if gap already exists
+    if isdir("gap")
+        ### Update the GAP clone
+        cd("gap")
+        run(`git pull --depth=1 origin master`)
```
fingolfin: So, what if we ever switch to building GAP point releases? Then `gap` won't be a git dir, right? So perhaps also check for the presence of `.git`?

Hm, IIRC, in the past, pulling in a shallow repo was problematic for various reasons (though this may have improved in recent git versions -- but of course we don't know if our users all have recent git versions, do we?). Also `git pull --depth=N` had its issues. Also note that this is slightly risky if files got renamed, possibly clashing with files we made.

I would therefore recommend that we also run something like `git clean -fx` or `git clean -fdx` (depending on whether we want to reuse the `pkg` dir or not... same for the downloaded packages...

Thinking further along this line, I wonder if we shouldn't do an out-of-tree build for GAP anyway?

PR #269, opened 2019-06-21
Place `call_gap_func` in the main module to ensure it is accessible

sebasguts:
even when the GAP module is not loaded into main (e.g., if loaded
from another module)

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -120,7 +120,7 @@ function __init__()
     ## corresponding function call in in ccalls.jl (func::GapObj)(...)
     MPtr = Base.MainInclude.eval(:(ForeignGAP.MPtr))
     Base.MainInclude.eval(:(
-        (func::$MPtr)(args...; kwargs...) = GAP.call_gap_func(func, args...; kwargs...)
+        (func::$MPtr)(args...; kwargs...) = $(GAP.call_gap_func)(func, args...; kwargs...)
     ))
 end
 


src/GAP.jl
```diff
@@ -119,8 +119,11 @@ function __init__()
     ## With Julia 1.2, this will be possible and this hack could then be replaced with the
     ## corresponding function call in in ccalls.jl (func::GapObj)(...)
     MPtr = Base.MainInclude.eval(:(ForeignGAP.MPtr))
+    global call_gap_func
+    call_gap_func_temp = call_gap_func
     Base.MainInclude.eval(:(
-        (func::$MPtr)(args...; kwargs...) = GAP.call_gap_func(func, args...; kwargs...)
+        __call_gap_func_temp = $(call_gap_func_temp);
+        (func::$MPtr)(args...; kwargs...) = __call_gap_func_temp(func, args...; kwargs...)
```
fingolfin: Huh, why isn't this sufficient (as a single line patch)?
```suggestion
        (func::$MPtr)(args...; kwargs...) = $(GAP.call_gap_func)(func, args...; kwargs...)
```

sebasguts:
Done.

src/GAP.jl
```diff
@@ -119,8 +119,9 @@ function __init__()
     ## With Julia 1.2, this will be possible and this hack could then be replaced with the
     ## corresponding function call in in ccalls.jl (func::GapObj)(...)
     MPtr = Base.MainInclude.eval(:(ForeignGAP.MPtr))
+    global call_gap_func
     Base.MainInclude.eval(:(
-        (func::$MPtr)(args...; kwargs...) = GAP.call_gap_func(func, args...; kwargs...)
+        (func::$MPtr)(args...; kwargs...) = $(call_gap_func)(func, args...; kwargs...)
```
fingolfin: So it doesn't work like this, without the `global` above?
```suggestion
        (func::$MPtr)(args...; kwargs...) = $(GAP.call_gap_func)(func, args...; kwargs...)
```
I just tested that locally, and it seemed to work fine? If there is no reason, let's use mine. If there is a reason, let's add a comment explaining it?

fingolfin: @sebasguts am I missing something?

sebasguts: @fingolfin I am not completely sure if this also works when GAP is loaded as a dependency of another module. Will test, and then change it accordingly (either comment or your version).

sebasguts: Tested, and your approach is valid. I updated the PR.

PR #272, opened 2019-06-30
Travis: build GAP package manuals

fingolfin:
This is a good idea in general, so that we notice any syntax errors we introduce there. It might also be helpful for @ThomasBreuer's PR #251

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -62,6 +62,7 @@ if install_gap
         cd("pkg")
         run(`git clone https://github.com/gap-packages/io`)
         run(`git clone https://github.com/gap-packages/profiling`)
+        run(`git clone https://github.com/gap-packages/AutoDoc`)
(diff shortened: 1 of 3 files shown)


sebasguts:
I think GAP master requires AutoDoc now, but it is not in the `bootstrap-pkg-minimal`. We could add it to our debug setup when installing GAP, what do you think?

fingolfin:
Well, yeah, for this to work, AutoDoc needs to be installed, but when using GAP master, that is not the case.

What we *really* should be doing is using Docker images with GAP, Singular, etc. already set up, so that we don't wast time on that...

In the meantime, we could either switch to using GAP 4.10.2;  or stay on `master` but then indeed install AutoDoc. For this PR, I'll just install AutoDoc, as that is by far the easiest change to make for now.

sebasguts:
Agreed. I think we should start with just installing AutoDoc. Its only dependency is GAPDoc, so that should work.

PR #279, opened 2019-07-15
fix access to matrix entries for the GAP master branch

ThomasBreuer:
GAP 4.10 calls the operation `\[\]` with 3 arguments,
the master branch calls the new operation `\[\,\]`.

Note that the default delegation from `\[\,\]` to `\[\]`
does not apply in the given situation because the GAP objects
are just Julia objects, without type information about `IsMatrixObj`.

(After this change, the tests that were broken after merging #276 should pass again.)

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/gap/adapter.gi
+++ b/pkg/GAPJulia/JuliaInterface/gap/adapter.gi
@@ -55,19 +55,39 @@ InstallOtherMethod( \[\]\:\=,
 #
 # "matrices" / lists of lists
 #
+#TODO:  Remove the installations for '\[\]' and '\[\]\:\='
+#       as soon as a released GAP version supports '\[\,\]' and '\[\,\]\:\='.
+if IsBound( \[\,\] ) then
+InstallOtherMethod( \[\,\],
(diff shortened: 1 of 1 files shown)


pkg/GAPJulia/JuliaInterface/gap/adapter.gi
```diff
@@ -55,19 +55,39 @@ InstallOtherMethod( \[\]\:\=,
 #
 # "matrices" / lists of lists
 #
+#TODO:  Remove the installations for '\[\]' and '\[\]\:\='
+#       as soon as a released GAP version supports '\[\,\]' and '\[\,\]\:\='.
+if IsBound( \[\,\] ) then
+InstallOtherMethod( \[\,\],
```
fingolfin: Can't we just install methods for MatElm and SetMatElm? At least that was the theory, that this way one could write the code so that it works with 4.9, 4.10 and 4.11

fingolfin: Argh, no of course not, and you already explained the reason in your PR description. Sorry about that

PR #281, opened 2019-07-17
Pin object in gap_julia before encapsulating in GAP object

### LINKED ISSUE #280: memory problem with the Julia integration of GAP and Polymake
Currently Klaus Lux is using our setup for certain computations, which means that one starts GAP (the master branch, using Julia's garbage collector, plus some unpublished GAP packages that consist only of interpreted GAP code) and loads its JuliaInterface package, which loads Polymake.jl into Julia and then uses the GAP-Julia interface to call (frequently) some Polymake functions. This works fine in most of the cases, but occasionally we run into the following error.
```
Error, Exception occured at Polymake side:
C++ object was deleted in
  Julia.Base.getproperty( obj, _JL_RNAM_TO_JULIA_SYMBOL( rnam )
 ) at /users/oscar/gap-master/gap/pkg/GAPJulia/JuliaInterface/gap/adapter.gi called from
res := lp.MILP.MINIMAL_VALUE; at /users/klux/pkg/moc/gap/mocjulia.g:52 called from
MinimalIntegralSolution( z, A, b ) at /users/klux/pkg/moc/gap/mocjulia.g:134 called from
SolveViaPolymake( z, A, bb ) at /users/klux/pkg/moc/gap/polymake.gi:338 called from
MOC_MinimalMultiplicityMFS( scp, PinPS, proj, multfrees[depth], nonpims, M, k
 ) at /users/klux/pkg/moc/gap/polymake.gi:583 called from
SubtractAndStrip_tree( proj - maxes[multfrees[depth]][1] * PinPS[PS[multfree[depth]]], maxes,
 depth + 1, k ); at /users/klux/pkg/moc/gap/polymake.gi:586 called from
...  at *stdin*:27
```
We have not been able to reproduce this problem.
It occurs from time to time after many successful calls,
and if we repeat such a failed GAP session then we do not run into this error at the same place.

(Background:
The object 'lp' is a polytope object in which 'MILP' has been set via a Julia function, and GAP calls 'lp.MILP.MINIMAL_VALUE'.
The '.' access in GAP is executed by a call to Julia's 'getproperty'.
Here I cannot decide which of the two '.' calls fails.)

sebasguts:
@ThomasBreuer Could you please check if this resolves #280 ?

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
@@ -145,8 +145,10 @@ Obj NewJuliaObj(jl_value_t * v)
 {
     if (is_gapobj(v))
         return (Obj)v;
+    JL_GC_PUSH1(&v);
     Obj o = NewBag(T_JULIA_OBJ, 1 * sizeof(Obj));
     SET_JULIA_OBJ(o, v);
+    JL_GC_POP();
     return o;
 }
 


ThomasBreuer:
@sebasguts Thanks. I have installed this branch of GAP.jl, and will run the computations (several times) which have shown the error. This will take some time.

PR #289, opened 2019-09-19
Set GAPTypes and GapObj global variables in init, before loading GAP

sebasguts:
Fixes [this problem](https://github.com/homalg-project/HomalgProject.jl/issues/2)

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -17,6 +17,7 @@ import Base: length, convert, finalize
 
 import Libdl
 
+import GAPTypes
 import GAPTypes: GapObj
 
 export GapObj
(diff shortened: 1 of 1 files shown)


src/GAP.jl
```diff
@@ -111,6 +112,8 @@ function __init__()
     ## the module is loaded.
     gap_module = @__MODULE__
     Base.MainInclude.eval(:(__JULIAGAPMODULE = $gap_module))
+    Base.MainInclude.eval(:(__GapObj_TYPE_VAR = $(gap_module.GAPTypes.GapObj)))
+    Base.MainInclude.eval(:(__GapObj_MODULE_VAR = $(gap_module.GAPTypes)))
```
fingolfin: Could you please elaborate what's going on here? Why do these mysterious globals __GapObj_TYPE_VAR and __GapObj_MODULE_VAR have to be set (nothing else references them?)

sebasguts: Sorry for the missing link. The variables are used [here](https://github.com/gap-system/gap/pull/3669)

fingolfin: But then why is this PR here needed at all? We already have `__JULIAGAPMODULE` aka `gap_module ` on the GAP side, it should be no problem to extract its members / submodules `gap_module.GAPTypes` and `gap_module.GAPTypes.GapObj` from that, no?

sebasguts:
The corresponding GAP pr does not need these changes any more, therefore they can be ignored.

PR #300, opened 2019-11-27
Add upper bound to GAPTypes dependency

### LINKED ISSUE #299: Add upper bound to GAPTypes dependency
On https://github.com/JuliaRegistries/General/pull/5422 we were told:

> The following dependencies do not have a compat entry that has an upper bound: GAPTypes

fingolfin:
Resolves #299

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,6 +11,7 @@ REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 
 [compat]
 julia = "1.1"
+GAPTypes = "1"
 
 [extras]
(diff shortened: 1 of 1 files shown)


fingolfin:
There were two failed Travis jobs, both of them flukes. One was with Julia nightly, which I resolved by re-running the job. The remaining one is due to Julia 1.3.0-rc5 (still) being used by some Travis builders, while others already use Julia 1.3.0 (released last night), which fixes that regression. Restarting the job so far didn't help (I got 1.3.0-rc5 each time), I'll keep trying. Possibly also have to wait a day or two to make sure all Travis build servers switched to the new Julia version.

PR #302, opened 2019-12-09
Make GAP.jl threadsafe.

rbehrends:
Currently, this still misses error handling, for which we require hooks in the GAP kernel. The functionality offered by `RegisterSyLongjmpObserver()` is insufficient, as it doesn't allow us to save state.

### DIFF
--- a/pkg/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/JuliaInterface/src/JuliaInterface.c
@@ -38,6 +38,7 @@ void handle_jl_exception(void)
     string_object = jl_call1(JULIA_FUNC_String_constructor, string_object);
     BEGIN_GAP_SYNC();
     ErrorMayQuit(jl_string_data(string_object), 0, 0);
+    END_GAP_SYNC();
 }
 
 static jl_module_t * get_module(const char * name)
@@ -50,6 +51,7 @@ static jl_module_t * get_module(const char * name)
     if (!jl_is_module(module_value)) {
         BEGIN_GAP_SYNC();
         ErrorQuit("Not a module", 0, 0);
+        END_GAP_SYNC();
     }
     return (jl_module_t *)module_value;
 }
@@ -64,6 +66,7 @@ Obj Func_JULIAINTERFACE_INTERNAL_INIT(Obj self)
     if (!JULIA_GAPFFE_type) {
         BEGIN_GAP_SYNC();
         ErrorMayQuit("Could not locate the GAP.FFE datatype", 0, 0);
+        END_GAP_SYNC();
     }
     return NULL;
 }
--- a/pkg/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/JuliaInterface/src/JuliaInterface.c
@@ -6,6 +6,7 @@
 
 #include "calls.h"
 #include "convert.h"
+#include "sync.h"
 
 #include <src/julia_gc.h>
 
@@ -35,6 +36,7 @@ void handle_jl_exception(void)
     jl_value_t * string_object =
         jl_call1(JULIA_FUNC_take_inplace, JULIA_ERROR_IOBuffer);
     string_object = jl_call1(JULIA_FUNC_String_constructor, string_object);
+    BEGIN_GAP_SYNC();
     ErrorMayQuit(jl_string_data(string_object), 0, 0);
 }
 
@@ -45,8 +47,10 @@ static jl_module_t * get_module(const char * name)
     // JL_TRY/JL_CATCH.
     jl_value_t * module_value = jl_eval_string(name);
     JULIAINTERFACE_EXCEPTION_HANDLER
-    if (!jl_is_module(module_value))
+    if (!jl_is_module(module_value)) {
+        BEGIN_GAP_SYNC();
         ErrorQuit("Not a module", 0, 0);
+    }
     return (jl_module_t *)module_value;
 }
 
@@ -57,8 +61,10 @@ Obj Func_JULIAINTERFACE_INTERNAL_INIT(Obj self)
     jl_module_t * gap_module = get_module("__JULIAGAPMODULE");
     JULIA_GAPFFE_type =
         (jl_datatype_t *)jl_get_global(gap_module, jl_symbol("FFE"));
-    if (!JULIA_GAPFFE_type)
+    if (!JULIA_GAPFFE_type) {
+        BEGIN_GAP_SYNC();
         ErrorMayQuit("Could not locate the GAP.FFE datatype", 0, 0);
+    }
     return NULL;
 }
 
@@ -89,6 +95,7 @@ int is_gapobj(jl_value_t * v)
 
 static Obj Func_NewJuliaCFunc(Obj self, Obj julia_function_ptr, Obj args)
 {
+    BEGIN_GAP_SYNC();
     if (!IS_JULIA_OBJ(julia_function_ptr)) {
         ErrorMayQuit("NewJuliaCFunc: <ptr> must be a Julia object", 0, 0);
     }
@@ -96,6 +103,7 @@ static Obj Func_NewJuliaCFunc(Obj self, Obj julia_function_ptr, Obj args)
 
     jl_value_t * func_ptr = GET_JULIA_OBJ(julia_function_ptr);
     void *       ptr = jl_unbox_voidpointer(func_ptr);
+    END_GAP_SYNC();
     return NewJuliaCFunc(ptr, args);
 }
 
@@ -148,6 +156,7 @@ Obj NewJuliaObj(jl_value_t * v)
 jl_function_t * get_function_from_obj_or_string(Obj func)
 {
     jl_function_t * f = NULL;
+    BEGIN_GAP_SYNC();
     if (IS_JULIA_OBJ(func)) {
         f = (jl_function_t *)GET_JULIA_OBJ(func);
     }
@@ -158,10 +167,10 @@ jl_function_t * get_function_from_obj_or_string(Obj func)
         if (f == 0) {
             ErrorMayQuit("Function is not defined in julia", 0, 0);
         }
-        return f;
     }
     else
         ErrorMayQuit("argument is not a julia object or string", 0, 0);
+    END_GAP_SYNC();
     return f;
 }
 
@@ -183,6 +192,7 @@ static Obj Func_JuliaFunction(Obj self, Obj func)
  */
 static Obj Func_JuliaFunctionByModule(Obj self, Obj funcName, Obj moduleName)
 {
+    BEGIN_GAP_SYNC();
     RequireStringRep("_JuliaFunctionByModule", funcName);
     RequireStringRep("_JuliaFunctionByModule", moduleName);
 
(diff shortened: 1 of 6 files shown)


pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
```diff
@@ -165,8 +178,10 @@ jl_function_t * get_function_from_obj_or_string(Obj func)
         if (function == 0) {
             ErrorMayQuit("Function is not defined in julia", 0, 0);
         }
+        END_GAP_SYNC();
         return function;
     }
+    END_GAP_SYNC();
     ErrorMayQuit("argument is not a julia object or string", 0, 0);
     return 0;
```
fingolfin: We could change this to use a single `END_GAP_SYNC` by initializing `result` to 0; then changing `if (IsStringConv(func))` to `else if`; and replacing the variable `function` by `result`. I think the result would be clearer.

pkg/GAPJulia/JuliaInterface/src/convert.c
```diff
@@ -16,10 +18,16 @@ jl_value_t * julia_gap(Obj obj)
         return gap_box_gapffe(obj);
     }
     if (IS_JULIA_OBJ(obj)) {
-        return GET_JULIA_OBJ(obj);
+        BEGIN_GAP_SYNC();
+        result = GET_JULIA_OBJ(obj);
+        END_GAP_SYNC();
```
fingolfin: Why do we need to wrap calls to `GET_JULIA_OBJ` into SYNC sections? We only read a single value from the JuliaObj GAP object here; and these objects are effectively read-only, i.e., they are never modified after being created.

pkg/GAPJulia/JuliaInterface/src/convert.c
```diff
@@ -16,10 +18,16 @@ jl_value_t * julia_gap(Obj obj)
...
+        return result;
     }
     if (IS_JULIA_FUNC(obj)) {
-        return GET_JULIA_FUNC(obj);
+        BEGIN_GAP_SYNC();
+        result = GET_JULIA_FUNC(obj);
```
fingolfin: Same remark as above, except that a few more things are read, but all from immutable/read-only objects of type `T_FUNCTION` and `T_JULIA_OBJ`

pkg/GAPJulia/JuliaInterface/src/sync.c
```diff
@@ -0,0 +1,32 @@
...
+void EndGapSync()
+{
+  if (LockLevel > 1 && pthread_equal(pthread_self(), LockOwner))
+    LockLevel--;
+  else if (LockLevel == 1 && pthread_equal(pthread_self(), LockOwner)) {
+    LockLevel = 0;
+    pthread_mutex_unlock(&GapLock);
+  } else {
+    abort(); // cannot happen
+  }
```
fingolfin: Please `clang-format` this file.

rbehrends:
> What (if anything) is missing for now?

I'm still getting some crashes with multi-threading if I stress concurrency hard enough. Looking at what causes that.

fingolfin:
@rbehrends OK, but if you address my minor remarks above, perhaps we could already merge these bits?

pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
```diff
@@ -46,7 +49,7 @@ jl_module_t * get_module_from_string(char * name)
     jl_value_t * module_value = jl_eval_string(name);
     JULIAINTERFACE_EXCEPTION_HANDLER
     if (!jl_is_module(module_value))
-        ErrorQuit("Not a module", 0, 0);
+        SYNC_GAP { ErrorQuit("Not a module", 0, 0); }
```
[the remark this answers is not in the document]
rbehrends: My rationale was that I wanted it less likely that in such cases the `END_GAP_SYNC()` got missing by accident. The reason is that a forgotten `END_GAP_SYNC()` might result in a deadlock down the road and is really hard to track down.

However, in most cases of interest, we seem to need `BEGIN_GAP_SYNC()` and `END_GAP_SYNC()` anyway, it appears, so I'm not sure now myself if it's worth it.

pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
rbehrends: Sounds good. Done.

pkg/GAPJulia/JuliaInterface/src/convert.c
rbehrends: Ah, I hadn't realized that they were immutable, I think. This simplifies things.

fingolfin:
@rbehrends I took the liberty of rebasing this PR

fingolfin:
@rbehrends the code in here now has diverged again from `master`. This will keep happening. I'd like to once again suggest that we at least merge parts of this. My suggestion for this would be to simply have `BEGIN_GAP_SYNC` *do nothing* in the merged version; the important bit is simply to insert them into the code. Then future code refactoring will not cause lots of conflicts within this PR.

(I'd probably do `#define BEGIN_GAP_SYNC do {`  and `#defined END_GAP_SYNC  } while(0)` to at least capture basic mistakes when moving them around in the code).

pkg/JuliaInterface/src/JuliaInterface.c
```diff
@@ -90,10 +95,14 @@ int is_gapobj(jl_value_t * v)
...
 {
     if (!IS_JULIA_OBJ(julia_function_ptr)) {
-        ErrorMayQuit("NewJuliaCFunc: <ptr> must be a Julia object", 0, 0);
+        SYNC_GAP {
+            ErrorMayQuit("NewJuliaCFunc: <ptr> must be a Julia object", 0, 0);
+        }
     }
     if (!IS_PLIST(arg_names)) {
-        ErrorMayQuit("NewJuliaCFunc: <arg_names> must be plain list", 0, 0);
+        SYNC_GAP {
```
fingolfin, line 103: oh, and let's get rid of this `SYNC_GAP` and only use `BEGIN_GAP_SYNC()` (or vice versa) -- that way, I don't have to remember the meaning of three different macros, just two (or one).

rbehrends:
I've updated them; the operations default to no-ops for now and need to be enabled with a `#define`.

I haven't defined `BEGIN_GAP_SYNC()` and `END_GAP_SYNC()` as `do {` and `} while (0)` as suggested, because they don't always match; sometimes, we need an `END_GAP_SYNC()` before a return and I've omitted the ones after `ErrorQuit()` and `ErrorMayQuit()`, where they only served to document structure. But if you prefer it, I can easily restructure the changes.

I've fixed a couple of race conditions with threads enabled in the process, but I can still crash it with aggressively parallel tests sometimes.

pkg/JuliaInterface/src/convert.c
```diff
@@ -27,22 +28,27 @@ jl_value_t * julia_gap(Obj obj)
...
+            END_GAP_SYNC();
             return GET_JULIA_OBJ(obj);
         }
         // to handle immediate integers, booleans, FFEs, IS_JULIA_FUNC, we
         // use recursion - but only for internal types; this prevents an
         // infinite recursion, e.g. if an object's JuliaPointer points to the
         // object itself
-        if (TNUM_OBJ(obj) < FIRST_EXTERNAL_TNUM)
+        if (TNUM_OBJ(obj) < FIRST_EXTERNAL_TNUM) {
+            END_GAP_SYNC();
```
fingolfin, line 44: I really think I'd prefer if we use `BEGIN_GAP_SYNC` / `END_GAP_SYNC` symmetrically.

Alternatively, we could convert all this code to C++, and use RAII to take care of beginning and ending the GAP "sync". That would then also avoid bugs where somebody accidentally returns w/o calling END_GAP_SYNC.

fingolfin: In the meantime, in this particular function, we could easily restructure the code to have a single `return` at the end, though, no?

pkg/JuliaInterface/src/JuliaInterface.c
```diff
@@ -57,8 +61,10 @@ Obj Func_JULIAINTERFACE_INTERNAL_INIT(Obj self)
     jl_module_t * gap_module = get_module("__JULIAGAPMODULE");
     JULIA_GAPFFE_type =
         (jl_datatype_t *)jl_get_global(gap_module, jl_symbol("FFE"));
-    if (!JULIA_GAPFFE_type)
+    if (!JULIA_GAPFFE_type) {
+        BEGIN_GAP_SYNC();
         ErrorMayQuit("Could not locate the GAP.FFE datatype", 0, 0);
+    }
```
fingolfin, line 67: Huh... I wind this really irritating to read, and I think I'd prefer to have the `END_GAP_SYNC` in there. Or have a `ErrorWithSync` helper function which contains the call to `BEGIN_GAP_SYNC`, and then call that here and in similar places (I think it's fine to always use `ErrorMayQuit`, so no need for multiple such helpers).

pkg/JuliaInterface/src/convert.c
rbehrends: Yes, I've done that. I've also considered whether we should just use `GAP_SYNC` as the sole primitive, similar to the `GAP_THROW` rework in GAP proper, if we're going to always match them. But (1) that may produce scoping issues (see below for more) and (2) we may sometimes have to split the lock/unlock parts across functions (undesirable from a code cleanliness perspective, but it may become necessary).

pkg/JuliaInterface/src/JuliaInterface.c
rbehrends: Right now, I've simply put the `END_GAP_SYNC()` back in, even though the code is never reached. It still serves as documentation.

rbehrends:
I've also looked at using `do {` and `} while (0)` as part of the macros in order to ensure that they get matched up. However, sometimes we declare variables within that scope that are used afterwards, so I haven't done that yet.

pkg/JuliaInterface/src/convert.c
```diff
@@ -27,23 +28,28 @@ jl_value_t * julia_gap(Obj obj)
     if (obj == False) {
         return jl_false;
     }
+    jl_value_t * result = (jl_value_t *) obj;
```
fingolfin, line 31:
```suggestion
    jl_value_t * result = (jl_value_t *)obj;
```
otherwise fails the clang-format check

pkg/JuliaInterface/src/convert.c
```diff
@@ -27,23 +28,28 @@ jl_value_t * julia_gap(Obj obj)
...
         // infinite recursion, e.g. if an object's JuliaPointer points to the
         // object itself
-        if (TNUM_OBJ(obj) < FIRST_EXTERNAL_TNUM)
-            return julia_gap(obj);
-        ErrorMayQuit("JuliaPointer must be a Julia object or an internal GAP "
-                     "object (not a %s)",
-                     (Int)TNAM_OBJ(obj), 0);
+        else if (TNUM_OBJ(obj) < FIRST_EXTERNAL_TNUM) {
+            result = julia_gap(obj);
+        } else {
```
fingolfin, line 45:
```suggestion
        } else
        {
```
otherwise fails the clang-format check

pkg/JuliaInterface/src/convert.c
rbehrends: Fixed. I had run clang-format over it for the previous commit, but forgot it this time.

pkg/JuliaInterface/src/convert.c
rbehrends: Also fixed.

PR #309, opened 2020-01-17
Require Julia >= 1.3 and drop some hacks

fingolfin:
This gets rid of an ugly, ugly hack, but requires Julia 1.3. Once we are happy to switch all of OSCAR to Julia >= 1.3, this can be merged.

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -122,9 +122,7 @@ end
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-# # The (func::GapObj) function (commented out below) needs to be instantiated for `MPtr` 
-# # and is therefore moved to the init function
-# (func::GapObj)(args...; kwargs...) where T <: GapObj = call_gap_func(func, args...; kwargs...)
+(func::GapObj)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
 
 
 function call_gap_func(func::GapObj, args...; kwargs...)
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: julia
 julia:
-  - 1.1
-  - 1.2
   - 1.3
   - 1.4
   - nightly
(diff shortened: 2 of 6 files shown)


src/ccalls.jl
```diff
@@ -130,9 +130,7 @@ end
 > There is no argument number checking here, all checks on the arguments
 > are done by GAP itself.
 """
-# # The (func::GapObj) function (commented out below) needs to be instantiated for `MPtr` 
-# # and is therefore moved to the init function
-# (func::GapObj)(args...; kwargs...) where T <: GapObj = call_gap_func(func, args...; kwargs...)
+(func::GapObj)(args...; kwargs...) where T <: GapObj = call_gap_func(func, args...; kwargs...)
```
fingolfin: @kalmarek in our Skype just now, I think you were trying to point out some issue with the above code to me (in terms of efficiency perhaps?), but I didn't quite follow; if you could explain here, that'd be much appreciated :-)

kalmarek: I actually thought of
```julia
(func::T)(args...; kwargs...) where T <: GAP.GapObj = call_gap_func(func, args...; kwargs...)
```
but it just doesn't work :P with `ERROR: function type in method definition is not a type`.
Sorry for bothering you ;)

kalmarek: if we stick with `GAPTypes` `GapObj` we need to wait for 1.3 to define
```julia
(func::GapObj)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
```

We could also try 
```julia
struct GapObj{T}
   mptr{T}
end
```
so that `MPtr` doesn't need a supertype and we can write
```julia
(func::GapObj)(args...;kwargs...) = call_gap_func(func.mptr, args..., kwargs...)
```

fingolfin:
Based on https://github.com/oscar-system/Oscar.jl/issues/80 we could merge this now. @ThomasBreuer agreed?

ThomasBreuer:
@fingolfin Agreed, except that the two Travis tests with Julia 1.3 fail.
As far as I understand, Julia does not find the kernel function `GAP_register_GapObj` in these cases.

fingolfin:
Indeed, I'll take care of that ASAP

Issue #306, opened 2020-01-16
use shallow clone and `get(ENV, "GAP_INSTALL_PACKAGES", "no")` 


kalmarek:
when building GAP, just that we don't end up downloading the full git tree and 2.6GB of pkgs
```
~/.julia/packages/GAP/W20PN/gap/pkg on  stable-4.11 ⌚ 17:58:29
$ du -h --max-depth=1 ./ | sort -h -r
2.6G    ./
661M    ./NormalizInterface-1.1.0
465M    ./semigroups-3.2.1
247M    ./simpcomp
111M    ./CaratInterface
90M     ./transgrp
80M     ./SmallGrp-1.4.1
74M     ./sglppow-2.1
73M     ./irredsol-1.4
53M     ./tomlib-1.2.9
51M     ./YangBaxter-0.9.0
49M     ./hap-1.24
49M     ./ctbllib
48M     ./rcwa-4.6.4
47M     ./ferret-1.0.2
41M     ./primgrp-3.4.0
40M     ./PolymakeInterface-2019.09.02
40M     ./difsets-2.3.1
29M     ./kbmag-1.5.9
26M     ./digraphs-1.0.3
24M     ./smallsemi-0.6.12
19M     ./unitlib-4.0.0
18M     ./sonata-2.9.1
17M     ./FactInt-1.6.3
14M     ./anupq-3.2.1
13M     ./guava-3.15
13M     ./fining
12M     ./profiling-2.2.1
12M     ./corelg-1.53
```
also #285 would be greatly appreciated

fingolfin:
There seem to be two requests mixed in one issue here, let's see:

1. "Use shallow clone":
We used to use a shallow clone, but that makes it really awkward to track GAP fixes as they come in. The plan is to drop the clone and instead download a specific GAP release -- but this is waiting for GAP 4.11 to be released. Which was meant to happen in September... I've been pressing for this, but there is a problem with personal here. So I've been co-organizing a GAP workshop in March which is meant (among other things) to revise how GAP releases are made, so that delays with them won't happen again (well, at least not due to pure infrastructure reasons, like "there is just one person who has the knowledge and the SSH keys to do this, and they don't have time, because the release process takes several days")

[...]

PR #313, opened 2020-01-17
Reduce overhead in build.jl a bit

fingolfin:
- go back to using a shallow clone of git; but now don't try to reuse
  existing clones, just delete them and start over
- don't compile NormalizInterface, semigroups, simpcomp, and in fact,
  delete them, to save some space and reduce the overall build time

See issue #306

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -24,17 +24,9 @@ if install_gap
     println("Installing GAP ...")
     gap_branch = "stable-4.11"
     cd(extra_gap_root)
-    ## check if gap already exists
-    if isdir("gap")
-        ### Update the GAP clone
-        cd("gap")
-        run(`git fetch origin`)
-        run(`git checkout $(gap_branch)`)
-        run(`git pull origin $(gap_branch)`)
-    else
-        run(`git clone -b $(gap_branch) https://github.com/gap-system/gap`)
-        cd("gap")
-    end
+    run(`rm -rf gap`)
+    run(`git clone --depth=1 -b $(gap_branch) https://github.com/gap-system/gap`)
+    cd("gap")
     run(`./autogen.sh`)
     run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
     run(`make -j$(Sys.CPU_THREADS)`)
@@ -54,6 +46,10 @@ if install_gap
     if gap_install_packages == "yes"
         run(`make bootstrap-pkg-full`)
         cd("pkg")
+        # eliminate a few big packages that take long to compile
+        pkgs = Base.Filesystem.readdir()
+        pkgs = Base.filter(x -> startswith(x, r"Normaliz|semigroups|simpcomp"), pkgs)
+        run(`rm -rf $pkgs`)
         run(`../bin/BuildPackages.sh`)
     elseif gap_install_packages == "minimal"
         run(`make bootstrap-pkg-minimal`)


kalmarek:
wow, you're moving pretty fast! ;-)

fingolfin:
Well, I managed to quickly produce garbage here, because it doesn't do what I want (shell glob are not interpolated by `run`, it turns out. Ah well, shouldn't be hard to fix (this time maybe with some more local testing ;-)

deps/build.jl
```diff
@@ -24,17 +24,9 @@ if install_gap
...
-        ### Update the GAP clone
-        cd("gap")
-        run(`git fetch origin`)
-        run(`git checkout $(gap_branch)`)
-        run(`git pull origin $(gap_branch)`)
-    else
-        run(`git clone -b $(gap_branch) https://github.com/gap-system/gap`)
-        cd("gap")
-    end
+    run(`rm -rf gap`)
```
kalmarek: I'm no git guru ;) but maybe this will be better:
```julia
try
    if isdir("gap")
        cd("gap")
        run(`git checkout $(gap_branch)`)
        run(`git pull origin $(gap_branch)`)
    else
        mkdir("gap")
        cd("gap")
        run(`git remote add origin https://github.com/gap-system/gap`)
        run(`git fetch --depth 100 origin $(gap_branch))
        run(`git checkout $(gap_branch)`)
    end
catch e
    @error("Error occured while updating gap repository at $(pwd())
    rethrow(e)
finally
    cd(orig_dir)
end
```
in general whenever I need to `cd` I try to `try-catch-finally`

fingolfin: No that won't work for multiple reasons
- `gap` may be an existing dir but not a git repository (indeed in the future, this is intended to be the default; it already would be, if the GAP 4.11 release hadn't been delayed so badly)
- even if it is a `.git` repository, it'll be a shallow clone; while un-shallowing supposedly works pretty well these days, I am wary of it due to issues I had with it in the past
- even if it is a `.git` repository, it's not guaranteed that `origin` is right (but that could be fixed by specifying the remote URL explicitly, I guess)
- in the `else` branch, that code won't work at all (to what repository are you adding that `origin` remote? Well, to the enclosing git repository... which is not the empty `gap` dir you just created, but rather the dir of the Julia package; which probably already has a remote `origin`, so it'll just run into an error). The fix is to use `git clone --depth=...` -- which is precisely what my PR does :-)

Good point about that try/catch, I'll add that.

fingolfin: Hmm, I guess I'd best wrap all of `build.jl` into a try/catch then... but that makes me wonder: why doesn't `Pkg.jl` do that (save & restore PWD) for you? Or perhaps it does (hmm, nah, at least on a quick look through its code base, it doesn't seem to; ah well)

kalmarek: I think I forgot `git init`, but looking at the documentation of `git clone --depth` I guess it does in one step what I did in three ;-)
```julia
try
           if isdir("gap")
               cd("gap")
               run(`git checkout $(gap_branch)`)
               run(`git pull origin $(gap_branch)`)
           else
               mkdir("gap")
               cd("gap")
               run(`git clone --depth 100 -b $(gap_branch) https://github.com/gap-system/gap` )
           end
       catch e
           @error("Error occured while updating gap repository at $(pwd())")
           rethrow(e)
       finally
           cd(orig_dir)
       end
```
otherwise it works (just checked);
 * I guess the future is BinaryGAP, so this becomes a no-issue
 * there is no need to unshallow it
 * if it is a user supplied directory with git-tree it's also user responsibility to point origin to the right remote
 * all errors are handled gracefully and the user is informed what happened, e.g. without `git init`:
```
error: pathspec 'stable-4.11' did not match any file(s) known to git
┌ Error: Error occured while updating gap repository at /home/kalmar/.julia/dev/GAP/gap
└ @ Main REPL[14]:15
ERROR: failed process: Process(`git checkout stable-4.11`, ProcessExited(1)) [1]
```

fingolfin: Perhaps it can be made to work, but what's the usecase? On the downside, it means people can end up in weird unsupported states. If you are requesting us to build gap for you, let us do it, by starting from a clean, known-to-be-good slate; if you want a custom GAP, just don't ask us to build it for you, and supply your own.

kalmarek: the usecase is that we don't download full git tree (accomplished) and gap tarball (still downloaded) for packages every time `build` is run. But I guess we need to live with this for the time

PR #320, opened 2020-01-20
process the main manuals in the installation of GAP

ThomasBreuer:


deps/build.jl
```diff
@@ -38,6 +38,9 @@ if install_gap
     run(`./autogen.sh`)
     run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
     run(`make -j$(Sys.CPU_THREADS)`)
+    cd("doc")
+    run(`make_doc nopdf`)
```
fingolfin: This won't work, you'd need this:
```suggestion
    run(`./make_doc nopdf`)
```
But even better is to invoke `make html` (I've submitted PR #321 doing that)

PR #321, opened 2020-01-21
Build the GAP manual if we install GAP

### LINKED ISSUE #316: Some GAP docstrings are missing?
```julia
help?> GAP.Globals.ListPerm
Automata (not loaded):
  2.5-7 ListPermutedAutomata
  
  ‣ ListPermutedAutomata( aut ) ────────────────────────────────────────────────── function
  
  Given an automaton aut, returns a list of automata with permuted states
  
  ───────────────────────────────────────  Example  ───────────────────────────────────────
    gap> x:=Automaton("det",3,2,[ [ 0, 2, 3 ], [ 1, 2, 3 ] ],[ 1 ],[ 2, 3 ]);;
    gap> ListPermutedAutomata(x);
    [ < deterministic automaton on 2 letters with 3 states >, 
      < deterministic automaton on 2 letters with 3 states >, 
      < deterministic automaton on 2 letters with 3 states >, 
      < deterministic automaton on 2 letters with 3 states >, 
      < deterministic automaton on 2 letters with 3 states >, 
      < deterministic automaton on 2 letters with 3 states > ]
  ─────────────────────────────────────────────────────────────────────────────────────────
  

─────────────────────────────────────────────────────────────────────────────────────────────
```
misses the `Reference` docstring;

on the other hand
```julia
help?> GAP.Globals.PermList
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 2nd choice method found for `+' on 2 arguments
ERROR: Error thrown by GAP
Stacktrace:
 [2] error_handler() at /home/kalmar/.julia/dev/GAP/src/GAP.jl:48
 [3] #call_gap_func#5(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(GAP.call_gap_func), ::Main.ForeignGAP.MPtr, ::Main.ForeignGAP.MPtr, ::Vararg{Any,N} where N) at /home/kalmar/.julia/dev/GAP/src/ccalls.jl:149
 [4] call_gap_func(::Main.ForeignGAP.MPtr, ::Main.ForeignGAP.MPtr, ::Vararg{Any,N} where N) at /home/kalmar/.julia/dev/GAP/src/ccalls.jl:139
 [5] #_#5(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Main.ForeignGAP.MPtr, ::Main.ForeignGAP.MPtr, ::Vararg{Any,N} where N) at /home/kalmar/.julia/dev/GAP/src/GAP.jl:124
 [6] (::Main.ForeignGAP.MPtr)(::Main.ForeignGAP.MPtr, ::Vararg{Any,N} where N) at /home/kalmar/.julia/dev/GAP/src/GAP.jl:124
 [7] GAP_help_string(::String, ::Bool) at /home/kalmar/.julia/dev/GAP/src/help.jl:4 (repeats 2 times)
 [8] doc(::GAP.GAPHelpType, ::Type) at /home/kalmar/.julia/dev/GAP/src/help.jl:21 (repeats 2 times)
 [9] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/REPL/src/docview.jl:355
```

(I have to admit GAP docstrings are pretty cool ;-)

fingolfin:
Closes #320
Fixes #316

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -51,16 +51,25 @@ if install_gap
         pkgs = Base.filter(x -> startswith(x, r"Normaliz|semigroups|simpcomp"), pkgs)
         run(`rm -rf $pkgs`)
         run(`../bin/BuildPackages.sh`)
+        cd("..")
     elseif gap_install_packages == "minimal"
         run(`make bootstrap-pkg-minimal`)
+        cd("pkg")
+        run(`git clone https://github.com/frankluebeck/GAPDoc`) # HACK to workaround outdated packages-required-stable-4.11.tar.gz
+        cd("..")
     elseif gap_install_packages == "debug"
         run(`make bootstrap-pkg-minimal`)
         cd("pkg")
+        run(`git clone https://github.com/frankluebeck/GAPDoc`) # HACK to workaround outdated packages-required-stable-4.11.tar.gz
         run(`git clone https://github.com/gap-packages/io`)
         run(`git clone https://github.com/gap-packages/profiling`)
         run(`git clone https://github.com/gap-packages/AutoDoc`)
         run(`../bin/BuildPackages.sh io profiling`)
+        cd("..")
     end
+    # build the GAP manual so that the help system works (must be done
+    # after GAP packages are installed)
+    run(`make html`)
 end
 
 gap_executable = abspath(joinpath(gap_root, "gap"))


fingolfin:
@ThomasBreuer @kalmarek Unfortunately this change increases the CI times massively, from 10 minutes to 45 minutes per job. I find this rather baffling, but have not yet had time to look closer at  what's going on. On suspicion would be that to build the manual, GAPDoc has to load and parse tons of files, and I think disk I/O in the Travis VMs is rather slow. But still, 35 minutes is a *lot* (on my laptop, `make html` takes 2.5 minutes -- which is slower than I'd want it to be, but more than a magnitude faster than what we see on Travis)

Unless we can figure out a way to speed this up, I am afraid this PR is a no-go for now. I guess I could hack it to only build the manual once (at the cost of not getting cross refs right), but that'd still add something like 17-20 minutes...

Of course once we can switch to a released version of GAP, this problem just goes away, as that comes with a fully generated set of documentation.

kalmarek:
I understand, since we've just decreased the build time there is no point in increasing it again. 

Well, that's yet another reason why Binary Builder should have shipped yesterday, but that's life ;)

ThomasBreuer:
@fingolfin Julia needs only the '.txt'  files of the manuals, thus it would make sense to omit not only the PDF version but also the HTML version; this reduces the runtime on my laptop by 30 %. For that, another change in GAPDoc would be necessary. Since the problem will eventually just go away, it is perhaps not worth to proceed like this.

ThomasBreuer:
@fingolfin Until the problem goes away, we could disable the `make html` in the situation where `build.jl` is processed by Travis, for example by setting a Julia variable before calling `Pkg.build` in `.travis.yml`, and checking for this variable in `build.jl`.
This way, user installations of GAP via GAP.jl would be complete in the sense that they have the main GAP manuals.

PR #322, opened 2020-01-23
Get rid of GAPTypes.jl

### LINKED ISSUE #305: GAPTypes are added to the main environment
`build.jl` should not add packages to the global environment as it does now

### LINKED ISSUE #310: `] dev --local GAP` results in a GAP.jl installation that cannot start due to missing GAPTypes.jl
```julia
julia> ENV["GAP_INSTALL_PACKAGES"] = "no";

(gap) pkg> dev --local GAP
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/oscar-system/GAP.jl.git`
 Resolving package versions...
  Updating `/tmp/gap/Project.toml`
  [c863536a] + GAP v0.2.1 [`dev/GAP`]
  Updating `/tmp/gap/Manifest.toml`
  [c863536a] + GAP v0.2.1 [`dev/GAP`]
  [cd4130c4] + GAPTypes v1.0.0
  [2a0f44e3] + Base64 
  [ade2ca70] + Dates 
  [b77e0a4c] + InteractiveUtils 
  [76f85450] + LibGit2 
  [8f399da3] + Libdl 
  [56ddb016] + Logging 
  [d6f4376e] + Markdown 
  [44cfe95a] + Pkg 
  [de0858da] + Printf 
  [3fa0cd96] + REPL 
  [9a3f8284] + Random 
  [ea8e919c] + SHA 
  [9e88b42a] + Serialization 
  [6462fe0b] + Sockets 
  [cf7118a7] + UUIDs 
  [4ec0a83e] + Unicode 
  Building GAP → `/tmp/gap/dev/GAP/deps/build.log`

julia> using GAP
Adding path /tmp/gap/dev/GAP/gap/.libs to DL_LOAD_PATH
Panic in src/julia_gc.c:776: GAPTypes.jl is not available
```
this also happens with `ENV["GAP_INSTALL_PACKAGES"] = "yes";`

fingolfin:
This requires a custom GAP branch for now, but see also https://github.com/gap-system/gap/pull/3867

Fixes #305
Fixes #310

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,13 +4,11 @@ authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsch
 version = "0.2.4"
 
 [deps]
-GAPTypes = "cd4130c4-8d21-11e9-2a29-61e008680c2d"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 
 [compat]
-GAPTypes = "1"
 julia = "1.1"
 
 [extras]
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -30,18 +30,7 @@ if install_gap
     run(`./autogen.sh`)
     run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
     run(`make -j$(Sys.CPU_THREADS)`)
-    ## FIXME: Hack to make GAPTypes available in the global package environment:
-    ## For a moment, we set the Julia load path to the actual depot, i.e.,
-    ## the env GAP.jl is installed to, to install GAPTypes.jl globally.
-    ## Since the package manager usually sets the load paths, 
-    ## JULIA_LOAD_PATH should always be set when this script is
-    ## executed. However, just setting this environment variable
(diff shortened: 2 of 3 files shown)


fingolfin:
Unfortunately the new GAP.jl in this PR requires the new GAP; mixing either of them with an old version of their counterpart doesn't work (some cases work, but not the general thing).
I see two options to go forward:
1. We just accept that, and try to merge both changes about the same time, and then immediately make a new GAP.jl release so as to not annoy users too much
2. I try to come up with a way to split this into separate updates which always retain compatibility to one version prior:
  - this PR would get split into at least two parts: one adds the call to `GAP_register_GapObj` but basically nothing else (in particular, doesn't drop GAPTypes)
  - we make a release of `GAP.jl` immediately afterwards
  - next, GAP is update by merging https://github.com/gap-system/gap/pull/3867 into `master` and `stable-4.11`; this will work with the latest released `GAP.jl`; it drops use of GAPTypes from GAP
  - finally, we update GAP.jl once more, removing references to GAPTypes here, too.

The second approach is more work but should be safer, and requires less perfect timing. So I'll try and see if I can do it this way.

kalmarek:
I'd go for the first version, it's less work and `GAP.jl` is still living on the bleeding edge ;-)

fingolfin:
@kalmarek not sure what you refer to with "first version"? Anyway, this PR is already merged

PR #323, opened 2020-01-23
Enh/refactor gap to julia

kalmarek:
But after @fingolfin's comments about `GAPInt ↔ Int/BigInt/MPtr` I think all of this is moot as long as the low level interface doesn't give us type stability in some (maybe inconvenient way)

### DIFF
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -15,8 +15,22 @@ function Base.string( obj::Union{GapObj,FFE} )
 end
 
 ## implement indexing interface
-Base.getindex(x::GapObj, i::Int64) = Globals.ELM_LIST(x, i)
-Base.setindex!(x::GapObj, v::Any, i::Int64 ) = Globals.ASS_LIST( x, i, v )
+Base.getindex(x::GapObj, i::Integer) = Globals.ELM_LIST(x, Int(i))
+Base.setindex!(x::GapObj, v, i::Integer) = Globals.ASS_LIST( x, Int(i), v)
+
+function Base.iterate(x::GapObj)
+    Globals.IsList(x) || throw(ArgumentError("GAP object is not iterable:\n $x"))
+    len = length(x)
+    iszero(len) && return nothing
+    return x[1], (1, len)
+end
+function Base.iterate(x::GapObj, state)
+    s, len = state
+    s == len && return nothing
+    return x[s+1], (s+1, len)
+end
+
+Base.eltype(x::GapObj) = Any # guess_eltype(x)
 Base.length(x::GapObj) = Globals.Length(x)
 Base.firstindex(x::GapObj) = 1
 Base.lastindex(x::GapObj) = Globals.Length(x)
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -64,7 +64,7 @@ function Base.BigInt(x::GapObj)
 end
 
 function Base.big(obj::GapObj) # very much type unstable
-    Globals.IsFloat(obj) && return BigFloat(Float64(obj))
+    Globals.IsIEEE754FloatRep(obj) && return BigFloat(Float64(obj))
     Globals.IsInt(obj) && return BigInt(obj)
     Globals.IsRat(obj) && return Rational{BigInt}(obj)
     throw(ConversionError(obj, "any of BigFloat, BigInt or Rational{BigInt}"))
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -37,7 +37,7 @@ function Base.Int64(obj::GapObj)
 end
 
 function Base.Float64(obj::GapObj)
-    Globals.IsFloat(obj) && return ValueMacFloat(obj)::Float64
+    Globals.IsIEEE754FloatRep(obj) && return ValueMacFloat(obj)::Float64
     throw(ConversionError(obj, Float64))
 end
 
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -22,233 +22,205 @@ of converted objects and should never be given by the user.
 ## (::Type{Array{T,1}}, :: GapObj) is invoked, with T = Tuple{Int64}; this then
 ## invokes gap_to_julia recursively with signature (::Tuple{Int64},::Any),
 ## which ends up selecting the method below.
-function gap_to_julia(t::T, x::Any) where T <: Type
-    if ! (typeof(x) <: t)
-        throw(MethodError("Wrong type: expected " * string(t) * ","))
-    end
-    return x
+
+struct ConversionError <: Base.Exception
+    obj
+    jl_type
 end
 
-## Default
-gap_to_julia(::Type{Any}, x::GapObj) = gap_to_julia(x)
-gap_to_julia(::Type{Any}, x::Any) = x
-gap_to_julia(::T,         x::Nothing) where T <: Type = nothing
-gap_to_julia(::Type{Any}, x::Nothing) = nothing
+Base.showerror(io::IO, e::ConversionError) =
+    print(io, "failed to convert GapObj to $(e.jl_type):\n $(e.obj)")
 
-## Integers
-gap_to_julia(::Type{Int128} ,x::Int64) = trunc(Int128 ,x)
-gap_to_julia(::Type{Int64}  ,x::Int64) = x
-gap_to_julia(::Type{Int32}  ,x::Int64) = trunc(Int32  ,x)
-gap_to_julia(::Type{Int16}  ,x::Int64) = trunc(Int16  ,x)
-gap_to_julia(::Type{Int8}   ,x::Int64) = trunc(Int8   ,x)
+function Base.Int64(obj::GapObj)
+    Globals.IsInt(obj) && return Int64(BigInt(obj))
+    throw(ConversionError(obj, Int))
+end
 
-## Unsigned Integers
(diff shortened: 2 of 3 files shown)


src/gap_to_julia.jl
```diff
@@ -22,233 +22,205 @@ of converted objects and should never be given by the user.
...
+
+    if !haskey(recursion_dict, obj)
+        new_array = Vector{T}(undef, length(obj))
+        for i in 1:length(obj)
+            current_obj = ElmList(obj, i)
+            # should return 'missing' for holes in the list
+            new_array[i] = get!(recursion_dict, current_obj,
+                _gap_to_julia(T, current_obj, recursion_dict))
         end
+        recursion_dict[obj] = new_array
```
ThomasBreuer: Why shall the assignment in `recursion_dict` be moved behind the loop?
If `obj` is self-referential then the changed variant runs into an error, whereas the old code had no problem.
Here is an example.
```
l = GAP.Globals.EvalString( GAP.julia_to_gap( "[ 1, ~, 3 ]" ) );
GAP.gap_to_julia( l )
```

src/adapter.jl
```diff
@@ -15,8 +15,22 @@ function Base.string( obj::Union{GapObj,FFE} )
...
+    Globals.IsList(x) || throw(ArgumentError("GAP object is not iterable:\n $x"))
+    len = length(x)
+    iszero(len) && return nothing
+    return x[1], (1, len)
+end
+function Base.iterate(x::GapObj, state)
+    s, len = state
+    s == len && return nothing
+    return x[s+1], (s+1, len)
+end
```
ThomasBreuer: What about the following extension:
```
function Base.iterate(x::GapObj)
    if Globals.IsList(x)
      len = length(x)
      iszero(len) && return nothing
      return x[1], (1, len)
    elseif Globals.IsDomain(x)
      iter = Globals.Iterator( x )
      Globals.IsDoneIterator( iter ) && return nothing
      return Globals.NextIterator( iter ), iter
    else
      throw(ArgumentError("GAP object is not iterable:\n $x"))
    end
end
function Base.iterate(x::GapObj, state::Tuple{Integer,Integer})
    s, len = state
    s == len && return nothing
    return x[s+1], (s+1, len)
end
function Base.iterate(x::GapObj, state::GapObj)
    Globals.IsDoneIterator( state ) && return nothing
    return Globals.NextIterator( state ), state
end
```
Then one can do the following.
```
g = GAP.Globals.SymmetricGroup( 3 )
for i in GAP.Globals.Elements( g ) print( i, ", " ) end
for i in g print( i, ", " ) end
```

src/gap_to_julia.jl
kalmarek: yep, I didn't even consider self-referring lists
btw.
```julia
julia> s = GAP.julia_to_gap( "[ 1, ~, 3 ]" )
GAP: "[ 1, ~, 3 ]"

julia> GAP.Globals.EvalString(s)
Error, Panic: tried to print a path of unsupported type 'component object'
Error showing value of type Main.ForeignGAP.MPtr:
ERROR: Error thrown by GAP
(...)
```
crashes

src/adapter.jl
kalmarek: this was just need to write concise code for constructors;

it does work, but indeed this iterate(Gapobj) is a julia anti-pattern, even if we can infer eltype of list and group. There is little point in writing more type unstable code, even if it works. If we know that a `GapObj` represents a group, lets wrap it in a `GAP(Perm)group` type and iterate over it in a type-stable form.

@fingolfin I just had an idea (since I believe your long-term plan is to reduce the amount of magic conversions in GAP) to explicitly set a `no-change-to-representation-please`-bitflag on objects coming from julia. This is a way to 'catch-them-all' (I mean all functions that try to do so) and eventually replace/rewrite them (they will fail on such objects). But maybe my understanding is too simplistic anyway ;)

fingolfin: I like what @ThomasBreuer wrote.

@kalmarek could you elaborate in how far this is a Julia anti-pattern?

And yeah, that `no-change-to-representation-please` object flag is precisely what I had in mind during our Skype call (I'd call it "locked representation object flag", which reflects existing GAP terminology). But I don't think it's very important right now, we have more pressing issues :-)

src/gap_to_julia.jl
fingolfin: As discussed on issue #324, the issue is in printing. But one can just suppress that:
```julia
julia> x = @gap [1,~,3] ;

julia> GAP.gap_to_julia(x)
3-element Array{Any,1}:
 1
  Any[1, Any[#= circular reference @-1 =#], 3]
 3

```

src/gap_to_julia.jl
```diff
@@ -22,233 +22,205 @@ of converted objects and should never be given by the user.
...
+Base.showerror(io::IO, e::ConversionError) =
+    print(io, "failed to convert GapObj to $(e.jl_type):\n $(e.obj)")
+
+function Base.Int64(obj::GapObj)
+    Globals.IsInt(obj) && return Int64(BigInt(obj))
+    throw(ConversionError(obj, Int))
+end
+
+function Base.Float64(obj::GapObj)
+    Globals.IsFloat(obj) && return ValueMacFloat(obj)::Float64
```
fingolfin: There are many different float implementations in GAP. `ValueMacFloat` is only defined for `MacFloats`, i..e., machine floats.
```suggestion
    Globals.IsIEEE754FloatRep(obj) && return ValueMacFloat(obj)::Float64
```

src/gap_to_julia.jl
```diff
@@ -22,233 +22,205 @@ of converted objects and should never be given by the user.
...
-        return gap_to_julia(T,x) // T(1)
-    end
-    if ! Globals.IsRat(x)
-        throw(ArgumentError("obj is not a rational"))
-    end
-    numer = Globals.NumeratorRat(x)
-    denom = Globals.DenominatorRat(x)
-    return gap_to_julia(T,numer) // gap_to_julia(T,denom)
+function Base.big(obj::GapObj) # very much type unstable
+    Globals.IsFloat(obj) && return BigFloat(Float64(obj))
```
fingolfin:
```suggestion
    Globals.IsIEEE754FloatRep(obj) && return BigFloat(Float64(obj))
```

src/gap_to_julia.jl
```diff
@@ -22,233 +22,205 @@ of converted objects and should never be given by the user.
...
 ## Generic conversions
 
-gap_to_julia(x::Any)  = x
-
 function gap_to_julia(x::GapObj)
-    if Globals.IsInt(x)
-        return gap_to_julia(BigInt,x)
-    elseif Globals.IsRat(x)
-        return gap_to_julia(Rational{BigInt},x)
-    elseif Globals.IsFloat(x)
```
fingolfin: Ah, the existing code was buggy already :-)

src/adapter.jl
kalmarek: its an anti-pattern since if You write `for g in G ...`, `g` will be of type `Union{Nothing, Int, BigInt, GapObj, Julia wrapped thing, Mptr...}` which essentially means `Any`.
If You create `PList{T, ElT}(::GapObj{T}) where {T, ElT}` first, this introduces a "function barrier" (where type instability is contained) and then iterate on `PList` in a type stable way. Whenever you have a switch statement to decide what to return (that cannot be decided at compile time), you're venturing into the world of `Any`.

kalmarek: in the sense I see no point in adding more code like this, until a design decision (on how to deal with Gap types on the julia side) is arrived at

ThomasBreuer:
@kalmarek I would like to understand your point of view, concerning the example of "iterable GAP objects".
You had proposed a function `iterate` that takes a `GapObj` and decides that this object is regarded as iterable if and only if `GAP.Globals.IsList` returns `true`.
In this case, Julia's `iterate` then shall remember the current position and shall use GAP's `[]` operation to fetch the next entry if requested.
The code piece that makes also GAP's domains (that is, structured sets such as groups) iterable is completely analogous, it just uses GAP's `NextIterator` instead of `[]`.
In both cases, Julia does not know anything about the GAP objects which will be returned by `iterate`.
(Can `eltype` for a GAP list ever return something different from `Any`, except in very special cases like Boolean lists and perhaps ranges containing only small integers?)

One interpretation of your statements is that one should better not invite people to write Julia code that cannot be fast. But this raises the question what one can expect at all from an interface to GAP, looking from the Julia side: Most GAP functions are interpreted and thus can be regarded as not fast.

kalmarek:
@ThomasBreuer of course you can implement that however you wish, and it can even work for now.

My point is that it is not really sensible to define high-level functionality (iteration) on `Mptr`s (or `GapObj`s). These should be exposed to user only as a last resort with the sole purpose of passing them back to GAP (as is `PropertyValue` in Polymake). If we want to work with the underlying data in Julia, we need to add more structure to it. Throwing bunches of `Any`s (or GapObj`s) will result in interpreted code which is an order of magnitude (or probably more) slower than similar code in GAP or python (since GAP/python have been optimized for running in the interpreted setting, while julia has not, and probably will never be). 

If you want to follow this design patter then you should go back to the design table and ask what objects of GAP and how can be made accessible in Julia. Current wrapper understands Ints, Rationals, floats, chars, strings, lists and records; not too much to change. I added iteration only as a stub for terser syntax in constructors, therefore I'm not really willing to turn it into a separate pull, but feel free to do so if You wish.

fingolfin:
Most (all?) of this has been taken care of now in one way or another.

PR #326, opened 2020-01-23
HACK: test to see if old GAP.jl can work with my latest GAP modifications

fingolfin:
... as submitted in https://github.com/gap-system/gap/pull/3867


fingolfin:
OK, so starting `julia` and loading GAP from that still works and passes tests; the other way around is broken, but that's to be expected as my GAP PR removes the GAPTypes hackery.

PR #327, opened 2020-01-23
Prepare to get rid of GAPTypes.jl

fingolfin:
This requires changes in GAP to work, specifically a new function
`GAP_register_GapObj` that we call in __init__. For now, we wrap
this call into try/catch so that it works with older GAP.jl, too.

Extracted from PR #322 (see explanations there for the reasons).

First iteration of this PR has a HACK commit which verifies everything works with the https://github.com/gap-system/gap/pull/3867 ; once it does, I'll drop that commit to make sure it works with GAP stable-4.11, too.

And once this PR has been merged, we should immediately make a release of GAP.jl

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -79,10 +79,19 @@ function initialize( argv::Array{String,1}, env::Array{String,1}, error_handler_
 end
 
 function finalize( )
-    ccall( (:GAP_finalize, "libgap")
-           , Cvoid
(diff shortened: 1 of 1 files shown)


fingolfin:
See also https://discourse.julialang.org/t/swizzling-the-super-type-of-a-foreign-julia-type-or-how-evil-must-i-be/33733

PR #328, opened 2020-01-24
Reduce deps/deps.jl to only contain the GAPROOT

fingolfin:
Also add a comment and remove unused const pkgdir.

I think we could get completely rid of `deps.jl` (and thus a little bit of complexity and a potential pitfall) if we just insisted that `${PACKAGE_DIR}/gap` must be a gap directory resp. a symlink to one. I think that's not unreasonable, is it?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -80,9 +80,6 @@ println("Generating deps.jl ...")
 
 deps_string = """
 GAPROOT = "$gap_root"
-GAP_EXECUTABLE = "$gap_executable"
-EXTRA_GAPROOT = "$extra_gap_root"
-
 """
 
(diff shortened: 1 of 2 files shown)


fingolfin:
No, the idea is this: We still check `ENV["GAPROOT"]` at build time. But then instead of creating a `deps/deps.jl` file which stores this path, we put a symlink `gap`  into the `GAP.jl` directory. Or, if the user did not provide a custom GAPROOT, it is an actually director containing `gap` (as it is right now).

ThomasBreuer:
Now I understand the idea. Yes, this way one can get rid of `deps/deps.jl`.

PR #331, opened 2020-01-27
Travis: add Julia 1.4 to test matrix

fingolfin:


fingolfin:
... not sure whether we really want to / need to test all Julia versions; perhaps it's enough to test the minimal (so 1.1 here) and the latest stable release branch (as of today: 1.3)

PR #347, opened 2020-02-27
fixed and extended GAP <-> Julia conversions

ThomasBreuer:
- fixed the GAP -> Julia conversion for strings
- added the GAP <-> Julia conversion for ranges
- added the GAP -> Julia conversion for Blists (to BitArrays);
  note that the Julia -> GAP conversion to Blists happens automatically
- changed the generic GAP -> Julia conversions for GAP strings, ranges, blists
- added tests for all this

(Why is Cuchar and not Char the default type for the conversion of GAP characters to Julia?)

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -103,6 +103,13 @@ gap_to_julia( ::Type{Float16}, obj::GapObj)  = Float16(gap_to_julia(Float64,obj)
 gap_to_julia( ::Type{BigFloat}, obj::GapObj) = BigFloat(gap_to_julia(Float64,obj))
 
 ## Chars
+function gap_to_julia( ::Type{Char}, obj::GapObj)
+    if ! Globals.IsChar( obj )
+        throw(ArgumentError("argument is not a character object"))
+    end
+    return Char( Globals.INT_CHAR(obj ) )
+end
+
 function gap_to_julia( ::Type{Cuchar}, obj::GapObj)
     if ! Globals.IsChar( obj )
         throw(ArgumentError("argument is not a character object"))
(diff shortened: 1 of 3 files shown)


fingolfin:
Good question. I think the reasoning was that Julia's `Char` is a 32bit type and meant to store unicode code point. But in GAP, we don't specify any encoding, a gap char really is the same as a byte, w/o any particular encoding. Not sure why we mapped this to `Cuchar` instead of `Cchar`, but that doesn't matter for your question, of course...

For most "regular" text, this doesn't matter, because it is ASCII. It would matter for strings that represent binary data read from a file; but that's somewhat rare in GAP.

All in all, I am open to changing this. Pure GAP chars are rarely used anyway.

src/gap_to_julia.jl
```diff
@@ -243,9 +301,18 @@ function gap_to_julia(x::GapObj)
     elseif Globals.IsFloat(x)
         return gap_to_julia(Float64,x)
     elseif Globals.IsChar(x)
+#T why Cuchar not Char?
```
fingolfin: I am fine with switching this to `Char`

test/conversion.jl
```diff
@@ -192,13 +227,32 @@ end
...
+    @test GAP.julia_to_gap( 1:0 ) == r
+    @test GAP.julia_to_gap( 1:1:0 ) == r
+    r = GAP.EvalString( "[ 1 ]" )
+    @test GAP.julia_to_gap( 1:1 ) == r
+    @test GAP.julia_to_gap( 1:1:1 ) == r
+    r = GAP.EvalString( "[ 4 .. 13 ]" )
+    @test GAP.julia_to_gap( 4:13 ) == r
+    @test GAP.julia_to_gap( 4:1:13 ) == r
+    r = GAP.EvalString( "[ 1, 4 .. 10 ]" )
+    @test GAP.julia_to_gap( 1:3:10 ) == r
```
fingolfin: I assume trying to convert `1:2^62` will throw an error? Perhaps this should also be tested for?

ThomasBreuer:
@fingolfin I have updated the pull request according to your suggestions

PR #354, opened 2020-03-04
catch GAP error messages in a Julia session

ThomasBreuer:
Up to now, GAP (running in a Julia session) printed its error messages
to the screen.
Now GAP's `ERROR_OUTPUT` gets redirected to a separate stream,
which is read and reset by the `error_handler` function in Julia.

(One effect is that the GAP error messages disappear when Julia tests for these errors are processed.)


ThomasBreuer:
My attempt to add a small change ended in a desaster.
I close this pull request and open a new one.
I hate GIT.

PR #357, opened 2020-03-05
Change the way we do Int64 conversion to/from GAP

fingolfin:
Previously, any Int64 too big to fit into a GAP immediate integer was put into
a T_JULIA wrapper when passed to GAP, and unwrapped later. At the same time,
`julia_to_gap` relied completely (and incorrectly) on the C kernel functions
`julia_gap` and `gap_julia` to take care of *all* Int64 values.

The result was that passing such an integer as argument to a GAP function did
not work, not even if one explicitly used `julia_to_gap`.

To fix this we teach the C kernel function `gap_julia` to convert "large"
Int64 values to GAP big integers; this ensures that the many adapter functions
we install for Int64 arguments work.

This means that we violate the principle of roundtrip type fidelity (Int64 -> GAP large
int -> MPtr), but that is a very minor concern.

For really efficient conversions of e.g. Array{Int64,1} to GAP plists, we need
dedicated conversion functions which directly copy over bits. These can be
added transparently in the future.

Finally, we add some minor optimization to julia_to_gap for UInt64, Int128, UInt128

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/gap/convert.gd
+++ b/pkg/GAPJulia/JuliaInterface/gap/convert.gd
@@ -142,7 +142,7 @@
 #!  <List>
 #!  <Item>
 #!    <C>Int64</C> to immediate integer when it fits,
-#!    otherwise to &Julia; object wrapper,
+#!    otherwise to a &GAP; large integer,
 #!  </Item>
 #!  <Item>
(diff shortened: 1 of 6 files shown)


fingolfin:
@ThomasBreuer I have now completely rewritten this. Tests pass

PR #358, opened 2020-03-05
let Documenter.jl test Julia examples

ThomasBreuer:


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsch
 version = "0.2.6"
 
 [deps]
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
fingolfin: If we only need this for tests, we should instead add it to the `[extras]` section, shouldn't we?

PR #361, opened 2020-03-05
Change default to not build any GAP packages

fingolfin:
We still download all packages, but don't build any of them, to save a lot of time when installing GAP.jl.

Power users can override this by setting the GAP_INSTALL_PACKAGES environment variable suitably when installing GAP.jl

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -31,8 +31,9 @@ if install_gap
     run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
     run(`make -j$(Sys.CPU_THREADS)`)
 
-    gap_install_packages =  get(ENV, "GAP_INSTALL_PACKAGES", "yes")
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -31,18 +31,24 @@ if install_gap
...
         run(`make bootstrap-pkg-full`)
         cd("pkg")
         # eliminate a few big packages that take long to compile
         pkgs = Base.Filesystem.readdir()
         pkgs = Base.filter(x -> occursin(r"^(Normaliz|semigroups|simpcomp)", x), pkgs)
         run(`rm -rf $pkgs`)
         run(`../bin/BuildPackages.sh`)
+    elseif gap_install_packages == "full-but-no-build"
+        # download all packages, but don't attempt to build any
+        run(`make bootstrap-pkg-minimal`)
```
ThomasBreuer: replace `minimal` by `full`?

PR #365, opened 2020-03-06
Switch to minimal ('core') GAP with a minimal set of packages

fingolfin:
We only install the minimally required packages, plus <https://github.com/gap-packages/PackageManager> (with version 1.0 hardcoded for now). Let's see if that works. Then the next step would be to add some helper functions, e.g. `GAP.add_package("PKGNAME")` which makes it convenient for users to install packages by delegating to `PackageManager`


docs/src/index.md
```diff
@@ -43,6 +42,16 @@ TODO: Describe access to arbitrary GAP variables and functions via `GAP.Globals.
...
 
 
+## Managing GAP packages
+
+```@docs
+GAP.Packages.load
+GAP.Packages.install
+GAP.Packages.update
+GAP.Packages.remove
+```
```
fingolfin: @mohamed-barakat With this you can do e.g. the following:
```
$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.3.1 (2019-12-30)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using GAP
 ┌───────┐   GAP 4.11.0-1-ge33ffb1 of today
 └───────┘   Architecture: x86_64-apple-darwin18.7.0-julia64-kv7
 Configuration:  gmp 6.1.2, Julia GC, Julia 1.3.1, readline, KernelDebug
 Packages:   GAPDoc 1.6.3, PrimGrp 3.3.2, SmallGrp 1.4.1, TransGrp 2.0.4

julia> GAP.Packages.install("Browse")
#I  Getting PackageInfo URLs...
#I  Retrieving PackageInfo.g from http://www.math.rwth-aachen.de/~Browse/PackageInfo.g ...
#I  Downloading archive from URL http://www.math.rwth-aachen.de/~Browse/Browse-1.8.9.tar.bz2 ...
#I  Saved archive to /var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T//tmN90Qhv/Browse-1.8.9.tar.bz2
#I  Extracting to /Users/mhorn/.gap/pkg/Browse ...
#I  Checking dependencies for Browse...
#I    GAPDoc >= 1.6: true
#I  Running compilation script on /Users/mhorn/.gap/pkg/Browse ...
true
```

This is still a bit rough -- for example, it prints GAP info messages -- which we could of course disable by lowering the info level, but they contain helpful info which we don't have otherwise. Maybe we can work with @mtorpey on a slightly more powerful API that gives us more status information?

More importantly, there is no way to declare in your Julia package that you depend on GAP's `IO` package; so you'll have to call `GAP.Packages.install("IO")` or so; but that would always try to do something, even if "io" is already installed and compiled. So you should for now maybe also call `GAP.Globals.TestPackageAvailability("IO")`  to decide whether to call `GAP.Packages.install("IO")`.

On the long run, we could provide Julia wrapper packages for all those GAP packages; something  like `GAP-IO.jl`, `GAP-Homalg.jl`, etc. and these could model dependencies etc. as in GAP (the `Project.toml` file could be generated from the `PackageInfo.g` to that end, and the Julia package would use `GAP.LoadPackageAndExposeGlobals` or something similar. More elaborate wrappers could be created later. Each such wrapper could contain its own copy of the relevant GAP package, and add itself as a GAP root path (many variations of this scheme are of course possible).

The main advantage of this approach would be that one could declare dependencies on GAP packages like any other Julia package dependency, and then installing your package using `Pkg.add("foobar")` in Julia would automatically ensure all required GAP packages are provided. But the main (and major) drawback would be that we'd have to somehow maintain all those wrappers. The only realistic way to do IMHO would be to fully automate their generation. All in all, I am not sure this is worth the effort, but I'd like to think a bit more about this.

In the meantime, I hope at least your most immediate needs are addressed?

mohamed-barakat:
I will split my thoughts into several posts. Here is the first one.

1. First of all, I like all the three functions and will start using them once they are there. However:
The latest `GAP.jl` v0.3.3 comes with an up to date GAP and a complete package repository. Using one of the above functions for a deposited package currently:
  * ignores the package directories shipped with `GAP.jl`;
  * downloads packages into `~/.gap/pkg`, which I am not quite happy with (at least for the moment), mainly since the current GAP
    - does not by default look for packages in an additional path like `~/.gap/dev`, so I place my development repositories in `~/.gap/pkg`, and am a bit concerned about the above functions possibly messing up with them;
    - does not save a downloaded package under ~/.gap/pkg/name-x.y.z in case it is cloned from a git-repository.

This is why I added to `HomalgProject.jl` v0.1.5 the function [`CompileGapPackage`](https://github.com/homalg-project/HomalgProject.jl/blob/master/src/HomalgProject.jl#L67), 
which compiles a deposited package "in place". I use this function [here](https://github.com/homalg-project/HomalgProject.jl/blob/master/src/HomalgProject.jl#L42) and [here](https://github.com/homalg-project/HomalgProject.jl/blob/master/src/HomalgProject.jl#L118). If you regard this function useful I can move it to `GAP.jl`.

src/packages.jl
```diff
@@ -64,3 +64,59 @@ function LoadPackageAndExposeGlobals(package::String, mod::Module; all_globals::
...
+
+TODO: write text, link to GAP's LoadPackage
+
+TODO: support version qualifier?
+"""
+function load(desc::String)
+    return Globals.LoadPackage(julia_to_gap(desc), false)
+    # TODO: can we provide more information in case of a failure?
+    # GAP unfortunately only gives us info messages...
+end
```
ThomasBreuer: Concerning the question about more information, what is wrong with info messages?
The typical advice in GAP is to call `SetInfoLevel(InfoPackageLoading,4);` and then to call `LoadPackage` again.

Concerning optional arguments of `load`, also the flag for showing/suppressing package banners could be supported.

mohamed-barakat:
2. The idea of automatically generating Julia packages for each GAP package definitely has the pros you addressed above, but I see the following drawbacks apart from the considerable maintenance overhead you mentioned:
  * You will `spam` the Julia package system with a lot of GAP packages. Who will be the maintainer of these packages? You cannot expect from the GAP package maintainers to maintain the Julia wrapper package.
  * I realized that `LoadPackageAndExposeGlobals( "packagename", Main, all_globals = true )` is still somewhat far from providing the a similar GAP-REPL user experience. For me, this was one of the reasons to start `HomalgProject.jl`.

mohamed-barakat:
3. The obvious downside of a minimal GAP distribution is that then `GAP.jl` is simply not GAP. And the very standard user is now forced to learn about installing packages, something she was (hopefully) never bothered with. I understand the merits of a slick `GAP.jl`. But then I would suggest either to

  * split `GAP.jl` into two packages: `GAP.jl` and `GAP-packages.jl`, where the former only contains exactly the packages it promises to compile and the latter compiles all of GAP deposited packages of the corresponding distribution;
  * or to allow `LoadPackageAndExposeGlobals` or a variant of it to do the package dependency analysis in the background and compile whatever is needed from the *deposited* packages in `GAP.jl`.

You see, my main concern is the user experience. It is, of course, a good user experience nice to have a `GAP.jl` which installs without a huge delay. But without further actions, it is only a `MinimalGAP.jl`.

src/packages.jl
fingolfin: I think the current behavior of GAP is very user unfriendly, already on the GAP side. There is no reason (other than technical ones, i.e., "nobody implemented this") why `LoadPackage("foo")` when entered interactively couldn't immediately tell the user things like: "can't load package `foo` because its dependency `bar` is missing", or "can't load `foo` because it require `bar`, and that requires `qux`, which is missing". And instead of "missing" of course we could also have other error reasons.

Anyway, that's not even what I had in mind here: point is that LoadPackage could fail because the package is not present at all; or because it failed to load (e.g. wasn't compiled, or a required external dependency is missing). Depending on which it is, the user might want to do different things to resolve it.

fingolfin:
@mohamed-barakat thank you for sharing your thoughts! Some replies:

Regarding 1.:
   
* "the above functions [...] ignores the package directories shipped with
  GAP.jl" -- the above functions ignore those directories only in so far as
  that they don't touch them. One can of course argue whether this is good or
  bad or neutral. I am not sure if you are doing so, and if so, in which way,
  and what the arguments are, so I can't  really debate this :-)

* I agree that using ``~/.gap` is problematic. For starters, if one uses
  several GAP major versions in parallel (e.g. 4.9, 4.10, 4.11), one may want
  to install different version of a package for each of them (see
  https://github.com/gap-system/gap/issues/3917).

  Secondly, there is of course a difference between "packages the user
  installed via PackageManager because they want to use them" -- those are
  fine in `~/.gap`, as that dir is owned by the user.

  But in your case, it's your project `HomalgProject.jl` that wants the
  package, and it shouldn't necessarily force the `io` package on the user.

  One solution for this would be to provide a function in `GAP.jl` which
  takes the name of a GAP package; if that package is available, do nothing
  (or perhaps load it). If it is missing (or misses deps?!? that bit is a bit
  harder to check for / handle), call
  `ExtendRootDirectories(GAP_PACKAGES_FOR_HOMALGPROJECT_JL)`. If the package
  is still missing, install it in `GAP_PACKAGES_FOR_HOMALGPROJECT_JL/pkg/`.
  Then we are done; or perhaps we load it.

  Note that the variable `PKGMAN_CustomPackageDir` exists precisely to allow
  overriding the install destination.
  
* As to your function `CompileGapPackage`, there is also `PKGMAN_CompileDir`
  doing a similar thing but maybe in a more general way. See also
  <https://github.com/gap-packages/PackageManager/issues/50>.
   
* "does not save a downloaded package under ~/.gap/pkg/name-x.y.z " -- I don't
  understand what your point here is?

Regarding 2.:

* I am pretty sure the Julia people are not concerned about "package spam",
  so that's IMHO a non-issue. And maintainer would something like "the OSCAR
  team". And ideally, I'd want these packages all to reside in a single
  repository (this requires an extension to `Pkg.jl`, see
  https://github.com/JuliaLang/Pkg.jl/issues/1251), possibly even within the
  GAP.jl repository, to make it maximally easy to maintain them.

  *That said*, it's not as if I am fully behind this idea myself; it solves
  some issues, but indeed also has its own set of issues. It's not something
  for the immediate future, but something to ponder. Of course in general it'd
  be better if there were "proper" Julia wrappers for some of these packages;
  which we do more than just load them.

  And perhaps this idea is just completely bad, and we should instead focus on
  a better GAP package manager and GAP package installation experience (we
  should work on that anyway, the main question is who can and will spend how
  much time on it, and how soon...)

* `LoadPackageAndExposeGlobals` is a crutch, and I was never really happy about it;
  I only approved it because Sebastian / you had some real need and I didn't want to
  block progress over a purely hypothetical "better" solution. But on the long run,
  I think there is

Regarding 3.: I firmly believe that the current GAP model, were we try to
include everything and the kitchen sink, is not going to scale and must be
changed anyway. Lots of basically unmaintained packages hinder progress
elsewhere, and testing all possible combinations of packages for compatibility
has long ago become infeasible; so we only test a few combinations, and so
often fail to notice conflicts via our automated tests. Downstream packages
for Linux distributions also seem to prefer splitting up GAP. I never was particularly
happy about this, but I can understand some of the reasons, now better than ever.

You say that (GAP) users in the past were "(hopefully) never bothered with". I
believe that this is true for 99% of all users by the simple virtue that they
were never even aware of the existence of packages, and never loaded one :-).
For the others, this was true if the package was bundled with GAP, and in the
right version; oh, and if it didn't require compilation. So overall, I believe
that of those people actually loading packages, countless actually *were*
bothered by this (and often enough failed because of it!). E.g. because the
latest version of a package was not bundled with their GAP release, or the GAP
release itself was outdated by several years and the latest version of the
package they needed only worked with a GAP development version. Or a package
they needed was simply not deposited.

And then there are packages like transgrp and atlasrep which already now don't
include all their data, and only download it on the fly when needed resp. ask
the user to download it.

Of course this is much better today, because we release GAP much more regularly,
have improved the ecosystem, have `PackageManager` (which still needs a lot of
work, but at least it *exists*), and we strive to further reduce these pain points.
I think bundling everything mainly was necessary because there was no viable
alternative. So personally, I hope we can improve `PackageManager` (possibly also with the
help of OSCAR funds) that GAP itself can one day switch to only shipping
a minimal set of packages, and make it trivial for users to install more.
Some ideas towards that end:
- trying to do `LoadPackage("foo")` for package which is not installed could
  offer the user to install the package for them.
- we could provide precompiled binaries for all "deposited" packages and major
  OS variants, by cribbing code from the Julia and/or Python people
- PackageManager could be made more resilient in general

By the way, downloading the full GAP 4.11.0 .tar.bz2 from St Andrews can be
done in under 6 seconds (on plesken). At least when one is lucky, i.e., if the
server is not offline (as it frequently is, for a couple seconds, round the
clock), or for some reason suddenly only downloads at the fraction of the
speed, and thus takes 10 minutes. Which is why I moved the download to GitHub;
it takes ~40 seconds to download GAP from there, but at least it is rather
consistent about this :-).


In the end, we are both concerned with the user experience, but different
aspects of it :-). I don't have a great conclusion to draw from this, and I
had a reason why I made this a PR instead of just merging it directly.
Gotta think and discuss more about this...

mohamed-barakat:
@fingolfin: Thank you for your thorough reply.

Regarding 1.:
  * For me it does not make much sense to let `GAP.jl` download all the deposited packages with gap-4.11.0, especially those who need compilation and make the default behavior of the provided installation functions (which rely on `PackageManager`) ignore these downloaded directories. I would, therefore, suggest either to
    - compile in the downloaded package directories, or
    - do not download these packages in the first place. 
   The latter option is in line with my suggestion of splitting `GAP.jl` into a really slick `CoreGap.jl` and `GapPackages.jl`.
  * I am happy with your suggestion [gap-packages/PackageManager#50](https://github.com/gap-packages/PackageManager/issues/50) which will make `CompileGapPackage` obsolete (thanks for the `PKGMAN_CompileDir` hint).
  * Still, I am not quite convinced with your suggestion regarding `GAP_PACKAGES_FOR_HOMALGPROJECT_JL/pkg/`. This makes sense when I want to install further CAP related packages for example, and less when it comes to a deposited GAP package that needs compilation. My concern: What if there is another Julia package `XYZ.jl` that also needs `io` and a user who wants to use `XYZ.jl` and `HomalgProject.jl` simultaneously. Which `io` binary will be used? Why should `io` be compiled twice against the same gap in `GAP.jl`?

mohamed-barakat:
Regarding 2.:

* I would also suggest postponing the discussion about automatically wrapping individual GAP packages to a future point. I assume that some smaller steps which are achievable in the near future could help advance this discussion at a later point. Regarding the extension of `Pkg`, I can already see a WIP-PR (https://github.com/JuliaLang/Pkg.jl/pull/1422) :) In any case, I like your suggestion of bundling all such packages in one repository with the OSCAR team as a maintainer.
* Currently `LoadPackageAndExposeGlobals` is essential to emulate a similar user experience as in GAP. You can see [here](https://github.com/homalg-project/HomalgProject.jl/blob/master/docs/src/examples.md) how -- apart from the obvious differences in syntax -- a Julia session strongly resembles a GAP session. In fact, I am not aware of your concerns involving `LoadPackageAndExposeGlobals`.

mohamed-barakat:
Regarding 3.: A quick response. Yes, Julia seems to be a good opportunity to improve the user experience. I got some positive feedback on how easy it became to access `homalg` and `CAP` through Julia (and what will be the future components of) `OSCAR.jl`. Risking to sound off-topic allow me to raise what I believe would contribute most to positive user experience:
* A binary distribution for `GAP.jl` :) We cannot assume that a macOS user has enough tools to compile `GAP.jl`, putting the entire user experience at risk. I am almost sure you have some advanced ideas and you probably also identified steps and problems. And although I am technically far behind I am willing to regard this as a top priority for `GAP.jl` and free some time and resources to work on it.
* Even with a binary distribution of `GAP.jl` I am very interested in a convenient way to install non-deposited packages, and we have plenty of them on https://github.com/homalg-project.

deps/build.jl
```diff
@@ -26,28 +26,32 @@ if install_gap
     println("Installing GAP ...")
     cd(extra_gap_root)
     run(`rm -rf gap-$(gap_version)`)
-    # TODO: gap-system.org availability is flaky, which often breaks CI builds; so use
-    # github mirror instead, which is more reliable, albeit somewhat slower
-    #filename = download("https://www.gap-system.org/pub/gap/gap-4.$(gap_version.minor)/tar.bz2/gap-$(gap_version).tar.bz2")
-    filename = download("https://github.com/gap-system/gap/releases/download/v$(gap_version)/gap-$(gap_version).tar.bz2")
+    println("Downloading GAP $(gap_version)")
```
mohamed-barakat:
```suggestion
    println("Downloading core GAP $(gap_version)")
```

docs/src/index.md
mohamed-barakat: I indeed am using `GAP.Globals.TestPackageAvailability("IO")` in [`CompileGapPackage`](https://github.com/homalg-project/HomalgProject.jl/blob/master/src/HomalgProject.jl#L67) used [here](https://github.com/homalg-project/HomalgProject.jl/blob/master/src/HomalgProject.jl#L42). Thanks for the hint.

src/packages.jl
mohamed-barakat: Yes, even with `SetInfoLevel(InfoPackageLoading,4)` it is always a non-trivial task to see what is missing, the first challenge being the deciphering of the long, long output. And even with an experienced eye, one has to do this exercise several times with almost to further help from GAP until one gets all dependencies satisfied. Julia is really a dream in this regard: `] add XYZ`, that's it.

ThomasBreuer: We must distinguish between installing and loading GAP packages.
The task to get the dependencies satisfied belongs to the installation; this belongs on the Julia side to Julia's `add`, and on the GAP side it was assumed to happen outside a GAP session until the idea of a package manager came up.
GAP's `LoadPackage` corresponds to Julia's `using`; in particular, `LoadPackage` should never attempt to change anything in the user's GAP installation.

(Of course I do not object to getting better information from `LoadPackage` or `TestPackageAvailability`.)

fingolfin:
- "do not download these packages in the first place": that's exactly what this PR does :-)

- "`LoadPackageAndExposeGlobals` is essential to emulate a similar user experience as in GAP." -- maybe, but to me that's not a goal at all, to the contrary: So you want a command that kind of pretends GAP packages are Julia package, and that behaves sort of as if one could do `using SomeGAPPackage`. But to me, that's just a hacky bandaid: GAP packages are not Julia packages, and I don't think it's a good idea for us to go around and pretend as if they were. Certainly not within OSCAR: I don't want to expose regular Julia users to GAP and GAP packages, as they are completely alien to the Julia universe, starting from the naming conventions. In my eyes `LoadPackageAndExposeGlobals` is completely out of scope for GAP.jl. I'd prefer to remove it here -- you could of course put it into a package of your own, but I'd urge you to reconsider this, too: if all you want is your interactive Julia sessions to look like GAP sessions, why are you even bothering to use Julia? In the example you linked to, I see nothing that you couldn't do with almost identical GAP code, so what is the point?

- "We cannot assume that a macOS user has enough tools to compile GAP.jl" -- yes, of course. A binary distribution for GAP.jl (or for GAP, for that matter) would be the ultimate goal. It's tricky to achieve, though, mainly because of the dependency from GAP to Julia (which also seems to depend on the Julia version, so we'd need separate binaries for each supported major Julia version -- at least for GAP itself, hopefully not for the packages). But at least it gets a lot easier if we don't try to cover *all* distributed packages at once. Each individual GAP packages that requires compilation requires extra effort to support. And for many of them I am not sure that effort is even worth it. Actually, this extra effort even affects many packages that don't require compilation. E.g. `polymaking` or `alnuth` which want to launch external binaries that may or may not be present. If we ship just a minimal set of packages, that simplifies a lot of things: shorter download time, easier to eventually get a binary distribution for this, and also we avoid the issue where some GAP packages build for some users but not for others due to random circumstances. Instead, users who explicitly try to load a package at least get a clear error if that fails.

- "Even with a binary distribution of GAP.jl I am very interested in a convenient way to install non-deposited package" -- but that's no easier or harder depending on what packages we preinstall -- you use the same command `GAP.Packages.install("package_name")` either way. So I am not sure what the point here is?

fingolfin:
@mohamed-barakat Just as a small update on this: GAP.jl has for quite some time now shipped a compiled GAP (via `GAP_jll`), and I also had compiled binaries for many GAP packages (including e.g. `io`) but no good way to make them available to users.

The repository with generated wrappers for all distributed GAP packages also has existed now for quite some time, but still is not quite ready for prime time. But I recently made progress. See https://github.com/oscar-system/GAP_pkg/

fingolfin:
Still want this but this PR is a dead end. We'll continue on this in other PRs.

PR #369, opened 2020-03-12
Improve printing of GAP objects, fix it for recursive lists

### LINKED ISSUE #324: printing of self-referring list crashes
```julia
julia> s = GAP.julia_to_gap( "[ 1, ~, 3 ]" )
GAP: "[ 1, ~, 3 ]"

julia> GAP.Globals.EvalString(s)
Error, Panic: tried to print a path of unsupported type 'component object'
Error showing value of type Main.ForeignGAP.MPtr:
ERROR: Error thrown by GAP
(...)
```

fingolfin:
Fixes #324

### DIFF
--- a/pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
@@ -330,6 +330,40 @@ static Obj FuncJuliaGetFieldOfObject(Obj self, Obj super_obj, Obj field_name)
     return gap_julia(field_value);
 }
 
+static Obj IsOutputStream;
+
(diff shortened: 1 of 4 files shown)


pkg/GAPJulia/JuliaInterface/src/JuliaInterface.c
```diff
@@ -330,6 +330,40 @@ static Obj FuncJuliaGetFieldOfObject(Obj self, Obj super_obj, Obj field_name)
     return gap_julia(field_value);
 }
 
+static Obj IsOutputStream;
+
+static Obj FuncStreamViewObj(Obj self, Obj stream, Obj obj)
```
fingolfin: Perhaps it would be better to not give the function a CamelCase name which invites people to use it. I'll change it to something like STREAM_VIEWOBJ

PR #376, opened 2020-03-12
Add basic wrappers for GAP PackageManager

### LINKED ISSUE #367: Add "nice" wrappers for GAP "PackageManager"
E.g. `GAP.install_package`, `GAP.update_package`, `GAP.remove_package` but also for GAP functionality: `GAP.install_package`, `GAP.package_available`, ...

fingolfin:
This was part of PR #365 but is of independent interest, and the discussion on PR #365 might go one for some more time.

One thing we still might want to change (or could change after merging this) is to add an (optional?) argument to `GAP.Packages.install` to specify an alternate installation location (see discussion on #365 for some reasons for this). And perhaps similar such arguments for `load`, `remove`, `update` would be useful?

Moreover, we could add `GAP.Packages.build` which for now might invoke `PKGMAN_CompileDir` (see also https://github.com/gap-packages/PackageManager/issues/50)

Resolves #367

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -46,6 +46,16 @@ TODO: Describe access to arbitrary GAP variables and functions via `GAP.Globals.
 TODO: describe Help system integration
 
 
+## Managing GAP packages
+
+```@docs
+GAP.Packages.load
+GAP.Packages.install
+GAP.Packages.update
+GAP.Packages.remove
+```
+
+
 
 ## Index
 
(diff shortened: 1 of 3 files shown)


fingolfin:
BTW, of course instead of putting this into a submodule `GAP.Packages`, we could have functions like `GAP.install_package` instead of `GAP.Packages.isntall`. Alas, the pattern I am using right now was inspired by `Pkg.jl`; it also naturally groups things together that belong together, so I kind like it.

mohamed-barakat:
I also like the similarity to Julia's `Pkg`. What about renaming `install` -> `add`? But I am also happy with `install`.

PR #378, opened 2020-03-13
Add hasproperty and setproperty methods for GlobalsType

fingolfin:
This allows setting GAP variables `foo` by assigning to `GAP.Globals.foo`,
and to test whether a variable is bound by using `hasproperty`.

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -1,6 +1,6 @@
 ## Internal ccall's
 
-import Base: getproperty, propertynames
+import Base: getproperty, hasproperty, setproperty!, propertynames
 
 function RAW_GAP_TO_JULIA(ptr::Ptr{Cvoid})::Any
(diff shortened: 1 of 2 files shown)


src/ccalls.jl
```diff
@@ -150,34 +151,37 @@ function call_gap_func(func::GapObj, args...; kwargs...)
...
 
 """
     Globals
 
-TODO: describe how this allows accesing any GAP variable or function as `GAP.Globals.VARIABLE_NAME`.
+TODO: describe how this allows accesing any GAP variable or function as
+`GAP.Globals.VARIABLE_NAME`.
 """
-Globals = GlobalsType(IdDict{Symbol,Cuint}())
+Globals = GlobalsType()
```
fingolfin: I should mention that this PR also gets rid of our "gvar cache" -- I did some benchmarks and discovered that it actually hurts performance

PR #384, opened 2020-03-21
Remove JuliaExperimental's kernel extension

fingolfin:
This simplifies things. I don't expect we'll need that kernel extension again, everything should be doable by calling into JuliaInterface or at worst via `ccall`.


fingolfin:
@ThomasBreuer is this OK from your point of view? If so, please merge :-)

PR #386, opened 2020-03-21
Move pkg/GAPJulia/ content into pkg/

fingolfin:
... at least I can't think of a reason why we'd want or need that `GAPJulia` subdirectory?


ThomasBreuer:
At the time when the directory structure was chosen, the idea was to regard the two GAP packages as belonging together, and if they would be distributed with GAP then the intermediate level inside the main `pkg` directory would express this.
Now the viewpoint is to distribute the GAP packages with GAP.jl and to provide the extra `pkg` directory. In this sense, the `GAPJulia` directory is unnecessary.

PR #399, opened 2020-04-01
track & check hash for source_tarball in build.jl

kalmarek:
don't download a new one if the old is present

with this it takes (here, locally)
* 100s to unpack
* 10s to configure
* 130s to make (-j4)


kalmarek:
very well, closing this then!

PR #400, opened 2020-04-03
Use artifacts system to manage GAP download; make things Julia version dependent; more

### LINKED ISSUE #254: Get GAP to use the same GMP as Julia, Nemo, ...
... by passing suitable build flags to it. That said, I don't even know if Nemo and Julia really use the same GMP right now? In any case, we should at least try to link against a foreign GMP, or even against our own GMP (built by GAP if it cannot find any "global" GMP installation).

### LINKED ISSUE #335: Add ~/.julia/gap.sh back or revise OSCAR install instructions
I just discovered that <https://oscar.computeralgebra.de/install/> promises that one can start the GAP built by GAP.jl via this command:

    ~/.julia/gap.sh

But that doesn't work anymore due to 600ce6d7ec3276d1399f0cd13061eb6d28aceb7a. We have two options:

1. add back that symlink
2. update the OSCAR install instructions.

My concern with 1 (and the reason why I removed the symlink in the first place) is that one may indeed have multiple versions of GAP.jl installed, and in different environments; and each of them might come with a different GAP version. With the old setup, `~/.julia/gap.sh` simply pointed at the last one installed; which means that it works well when you try it immediately; but at a later point, it might get broken (e.g. if you install GAP.jl in an environment, then remove that environment).

For 2, the question becomes: what should the instructions say then? One option would be to provide an alternative command which queries Julia for some path, e.g.:

    $(julia -e 'print(dirname(Base.find_package("GAP")))')/../gap/bin/gap.sh

Of course that's not pretty at all. Anybody got some other idea?

### LINKED ISSUE #377: Installing GAP.jl simultaneously within multiple Julia versions may crash
Based on empiric evidence, it seems that using a GAP compiled against Julia 1.2 from within Julia 1.3, or vice versa, does not work as it results in a crash. I suspect that this is due to subtle differences in the Julia kernel that affect our Julia GC integration (things like: "the content of some internal struct changed around, so our code doing this gets broken).

The ideal solution would be to find that code in the GAP kernel as well as Julia interface and remove it. Alas, that may not be feasible (we may not be able to identify the code, and/or there might be no way to avoid it). Perhaps @rbehrends has some thoughts on it.

The alternative then is to make sure we don't used code compiled for one Julia version in another. But right now, if one installs a specific version of GAP.jl , say 0.3.4, then that gets a install in a subdirectory like `~/.julia/packages/GAP/6rBgt`, and in there we install a copy of GAP. That copy then is used whenever GAP.jl 0.3.4 is loaded, regardless of whether it is Julia 1.2, 1.3, 1.4, or whatever. As a result if GAP.jl was installed with Julia 1.2, and we then try to load it in 1.3, we get a crash, and vice versa (have not yet tried what happens with 1.4).

One way to resolve this would be to use the fact that GAP supports out-of-tree build: instead of compiling GAP directly in `~/.julia/packages/GAP/6rBgt/gap-4.11.0`, we could instead create a subdirectory `~/.julia/packages/GAP/6rBgt/gap-4.11.0-julia-1.2` and run `../gap-4.11.0/configure && make` in there. Then use that dir as GAP root. This way, we share the source code but get separate binaries.

However, this does not resolve the issue for the C kernel of JuliaInterface. Perhaps there actually is no issue there, then we are done. Otherwise, we'll need a way to also compile and load different versions of that kernel extension...

There is a dual aspect to all this: right now, if one installs GAP 0.3.3 and 0.3.4 in parallel, they each get their own full copy of GAP. In particular, upgrading GAP.jl does not reuse the existing GAP installation. That's a bit annoying...

We could overcome this by installing GAP into a different place, e.g. into `~/.julia/packages/GAP/gap-4.11.0`. I don't see a downside to this right now (other than that Julia resp. Pkg.jl might not like us do that?). In that case, we could still have the dir in which we compile GAP in the same place relative to GAP.jl, i.e., in `~/.julia/packages/GAP/6rBgt/gap-4.11.0-julia-1.2`. Or, assuming we are reasonably sure we'll not use funky custom compilation options in future GAP.jl versions, we could use build dirs like `~/.julia/packages/GAP/gap-4.11.0/v1.2` or so.

BTW, compiled packages other than JuliaInterface should be fine to use across different Julia versions.

### LINKED ISSUE #392: Loading GAP.jl then Singular.jl causes ERROR: InitError: too many gmp
When I first load `GAP.jl` and then `Singular.jl` I get the following error

```julia
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.3.1 (2019-12-30)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(v1.3) pkg> status
    Status `~/.julia/environments/v1.3/Project.toml`
  [e30172f5] Documenter v0.24.7
  [c863536a] GAP v0.3.5
  [3e1990a7] Hecke v0.8.0
  [50bd374c] HomalgProject v0.1.8 [`~/.julia/dev/HomalgProject`]
  [7073ff75] IJulia v1.21.1
  [472f376f] LoadFlint v0.1.2
  [f1435218] Oscar v0.2.0
  [d720cf60] Polymake v0.3.2
  [bcd08a7b] Singular v0.2.1
  [42bdb5c4] Theta v0.1.0

julia> using GAP

julia> using Singular
ERROR: InitError: too many gmp
Stacktrace:
 [2] __init__() at /Users/mo/.julia/packages/LoadFlint/3t3NS/src/LoadFlint.jl:39
 [3] _include_from_serialized(::String, ::Array{Any,1}) at ./loading.jl:692
 [4] _require_search_from_serialized(::Base.PkgId, ::String) at ./loading.jl:776
 [5] _tryrequire_from_serialized(::Base.PkgId, ::UInt64, ::String) at ./loading.jl:707
 [6] _require_search_from_serialized(::Base.PkgId, ::String) at ./loading.jl:765
 [7] _tryrequire_from_serialized(::Base.PkgId, ::UInt64, ::String) at ./loading.jl:707
 [8] _require_search_from_serialized(::Base.PkgId, ::String) at ./loading.jl:765
 [9] _require(::Base.PkgId) at ./loading.jl:1001
 [10] require(::Base.PkgId) at ./loading.jl:922
 [11] require(::Module, ::Symbol) at ./loading.jl:917
during initialization of module LoadFlint
```

The other way round works does not raise an error:

```julia
julia> using Singular

Welcome to Nemo version 0.17.1-dev

Nemo comes with absolutely no warranty whatsoever


julia> using GAP
```

### LINKED ISSUE #395: Update build.jl to guard for exceptions
@kalmarek suggested:
> btw. looking at `build.jl` script, it's always a good idea to save `orig_dir = pwd()` and guard all blocks of bash commands with `try ... catch... finally cd(orig_dir) end`

### LINKED ISSUE #396: Track hash of GAP tarball, avoid extra downloads
We should compute a SHA256 hash (or whatever) of the GAP tarball, so we can reuse it instead of deleting and downloading again. Moreover, this can be used to verify the download was successful.

Actually, perhaps we can even switch to using the [Julia Artifacts system](https://julialang.github.io/Pkg.jl/dev/artifacts/) to keep track of the tarball?

fingolfin:
Several changes and improvements to how we build and use GAP inside of GAP.jl
are made by this patch:

1. We use the Julia artifacts systems to get the GAP source code: this
   way, Julia takes care of downloading the code and even performs a checksum
   validation (and in Julia 1.4, we get a nice download progress bar); by
   using artifacts, we ensure that each GAP release (such as 4.11.0, 4.11.1,
   ...) is downloaded only once; while previously any minor update of GAP.jl
   had to download GAP again. This saves storage and of course time.
   Resolves #396. Closes #399.

2. We switch to building GAP out-of-tree (this is of course key for allowing
   us to share a single GAP installation among multiple versions of GAP.jl).
   This way, we don't have to touch anything inside the GAP artifact directory!
   Unfortunately, for packages there is no out-of-tree support right now,
   so for those we still have to build in-tree.

3. We now build separate versions of GAP for each Julia version, to avoid
   subtle compatibility issues or even crash caused by GAP accessing low-level
   Julia APIs. This is done by (a) inserting the Julia version into the name
   of the directory in which we perform the GAP out-of-tree builds; and by
   setting ARCHEXT to insert the Julia version into the GAP arch string; this
   way, packages produce multiple binaries, one for each distinct GAP-Julia
   version pair.
   Resolves #377.

4. Add `GAP.gap_exe()` to make it easy to launch the GAP inside GAP.jl:
   Just enter `$(julia -e 'import GAP; print(GAP.gap_exe())')`.
   Resolves #335 (together with an update to the OSCAR website)

5. Restore the working directory if `build.jl` runs into an error.
   Resolves #395.

6. TODO: use GMP, readline and zlib binary builders
   Resolves #392. Resolves #254.


fingolfin:
This PR is mostly done now; I still want and need to do the BinaryProvider stuff, but not tonight. But what is there can already be reviewed. If I don't get to implement the `BinaryProvider` stuff tomorrow, I'll remove all traces from it from this PR again, then we can merge it (as it already contains a bunch of useful improvements) and continue with a new PR adding the BinaryProvider stuff.

mohamed-barakat:
This is impressive Max. I'll have a detailed look at it tomorrow. Many thanks.

fingolfin:
Ah, right, I forgot about `etc/generate_artifacts.jl`, I need to fix that to actually work! I cheated a bit to create the initial `Artifacts.toml`

deps/build.jl
```diff
@@ -1,68 +1,116 @@
...
+
+# TODO: maybe make the artifact loading lazy, for user's who want to use their own GAP???
+# OTOH, these are very few people and I am not sure we should care about them?
+@show artifact_dir = abspath(artifact"gap-4-11-0")
+
+# that artifact directory contains a single subdirectory, with a name like `gap-4.X.Y`;
+# get that name
+content = readdir(artifact_dir)
+@assert length(content) == 1
+gapdirname = content[1]
```
kalmarek, line 20: maybe matching with regex `r"gap-(4\.\d+\.\d+)"` is better suited here?

deps/build.jl
```diff
@@ -1,68 +1,116 @@
...
-filename = download("https://github.com/gap-system/gap/releases/download/v$(gap_version)/gap-$(gap_version).tar.bz2")
-run(`tar xjf $(filename)`)
-cd("gap-$(gap_version)")
-run(`./configure --with-gc=julia --with-julia=$(julia_binary)`)
-run(`make -j$(Sys.CPU_THREADS)`)
+
+# the location in which we initiate GAP's "out of tree build"
+gap_bin_root = abspath(joinpath(@__DIR__, "build", "$(gapdirname)-julia-$(VERSION)"))
+
+@show gap_src_root
```
kalmarek, line 32: are these kept to be shown in `--verbose` mode?

deps/build.jl
```diff
@@ -1,68 +1,116 @@
...
+
+# Download and install binaries
+for dependency in dependencies
+    download(dependency, basename(dependency))
+    evalfile(basename(dependency))
+end
+deps_prefix = joinpath(@__DIR__,"usr")
+
+# remember the working directory so that we can restore it later on
+orig_dir = pwd()
```
kalmarek, line 53: `orig_dir` can go;

btw. I like the solution of `cd(dir) do ... end` very much! (I didn't even know this signature existed:)

deps/build.jl
```diff
@@ -1,68 +1,116 @@
...
+
+println("Compiling GAP ...")
```
kalmarek, line 55: If we want to print stuff in `--verbose` I'd move the above @show to here as
```julia
@info "Compiling GAP with" gap_src_root extra_gap_root gap_bin_root
```

deps/build.jl
```diff
@@ -1,68 +1,116 @@
...
-GAPROOT = "$gap_root"
+GAPROOT = "$gap_bin_root"
 """
 
-path = abspath(joinpath(@__DIR__, "deps.jl"))
-println(path)
-open(path, "w") do outputfile
-    print(outputfile,deps_string)
+println(deps_jl)
+open(deps_jl, "w") do outputfile
```
kalmarek, line 112: I just learned about `write(filename, x)` which does exactly the same for `x::String`; You may, or may not want to it ;-)

etc/generate_artifacts.jl
```diff
@@ -0,0 +1,53 @@
...
+# then download and check? Or perhaps we allow using an already present local .tar.gz??
+gap_version = v"4.11.0"
+sha256 = "bf070a10dc2ac60593fe65df0f873d72d4d67501e86eb262a6f6bd48fc868602"
+
+name = "gap-$(version)"
+gap_tarball = "$(name).tar.bz2"
+url1 = "https://github.com/gap-system/gap/releases/download/v$(gap_version)/$(gap_tarball)"
+url2 = "https://files.gap-system.org/gap-4.$(gap_version.minor)/tar.bz2/$(gap_tarball)"
+
+# ensure tarball is available
```
kalmarek, line 15: this looks unfinished

etc/generate_artifacts.jl
```diff
@@ -0,0 +1,53 @@
...
+# Query the `Artifacts.toml` file for the hash bound to the name "iris"
+# (returns `nothing` if no such binding exists)
+gap_hash = artifact_hash(name, artifacts_toml)
+
+# If the name was not bound, or the hash it was bound to does not exist, create it!
+if gap_hash == nothing || !artifact_exists(gap_hash)
+    # create_artifact() returns the content-hash of the artifact directory once we're finished creating it
+    gap_hash = create_artifact() do artifact_dir
+        # We create the artifact by simply downloading a few files into the new artifact directory
+        iris_url_base = "https://archive.ics.uci.edu/ml/machine-learning-databases/iris"
```
kalmarek, line 33: can you do ML on iris dataset with GAP? ;P

kalmarek:
@fingolfin dismiss the comments at your will, `generate_artifacts.jl` looks generally unfinished

deps/build.jl
fingolfin: Either way works; but the actual dirname doesn't really matter, so I decided to instead enforce that there is just one. I could in addition also enforce a format of the dirname, but since this is not actually used later on, I just decided to take whatever is there :-)

deps/build.jl
fingolfin: I followed your good advice to use a single `@info` statement from below.

deps/build.jl
fingolfin: Removed, thanks

deps/build.jl
fingolfin: Ah, very nice -- changed

deps/build.jl
fingolfin: Done!

etc/generate_artifacts.jl
fingolfin: Yeah. I've now completed it.

fingolfin:
@ThomasBreuer @kalmarek thanks for your feedback, I think I addressed all your remarks.

Artifacts.toml
```diff
@@ -0,0 +1,9 @@
+[gap-4-11-0]
+git-tree-sha1 = "48e8399d220c782f4c6cf862f8ac0f436a2a6053"
```
mohamed-barakat, line 2: What is this SHA?

deps/build.jl
mohamed-barakat: If there is any chance that the directory `artifact_dir` could collect garbage then it is safer to insist on a specific dirname and even the existence of a specific file in this directory.

mohamed-barakat:
Minor issue: The download progress bar seems broken.

image

mohamed-barakat:
I get the following error

image

kalmarek:
@fingolfin are the `deps/build_*` necessary? these should be downloaded on demand, I think...

mohamed-barakat:
It does not work for me

```julia
julia> using GAP
Downloading artifact: gap-4-11-0
######################################################################## 100.0%-#O#- #   #
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   48e8399d220c782f4c6cf862f8ac0f436a2a6053
│   Calculated git-tree-sha1: 5f84ea130b97f5f163682f0f9bac5ebff45e7d18
Downloading artifact: gap-4-11-0
######################################################################## 100.0%-#O#- #   #
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   48e8399d220c782f4c6cf862f8ac0f436a2a6053
│   Calculated git-tree-sha1: 5f84ea130b97f5f163682f0f9bac5ebff45e7d18
┌ Error: ErrorException("Unable to automatically install 'gap-4-11-0' from '/Users/mo/.julia/dev/GAP/Artifacts.toml'")
└ @ Pkg.Operations /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1365
Downloading artifact: gap-4-11-0
######################################################################## 100.0%-#O#- #   #
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   48e8399d220c782f4c6cf862f8ac0f436a2a6053
│   Calculated git-tree-sha1: 5f84ea130b97f5f163682f0f9bac5ebff45e7d18
Downloading artifact: gap-4-11-0
######################################################################## 100.0%##O=#  #
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   48e8399d220c782f4c6cf862f8ac0f436a2a6053
│   Calculated git-tree-sha1: 5f84ea130b97f5f163682f0f9bac5ebff45e7d18
ERROR: LoadError: Unable to automatically install 'gap-4-11-0' from '/Users/mo/.julia/dev/GAP/Artifacts.toml'
Stacktrace:
 [2] ensure_artifact_installed(::String, ::Dict{String,Any}, ::String; platform::Pkg.BinaryPlatforms.Platform, verbose::Bool, quiet_download::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Artifacts.jl:894
 [3] ensure_all_artifacts_installed(::String; platform::Pkg.BinaryPlatforms.Platform, pkg_uuid::Nothing, include_lazy::Bool, verbose::Bool, quiet_download::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Artifacts.jl:958
 [4] download_artifacts(::Pkg.Types.Context, ::Array{String,1}; platform::Pkg.BinaryPlatforms.MacOS, verbose::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:609
 [5] download_artifacts(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; platform::Pkg.BinaryPlatforms.MacOS, verbose::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:588
 [6] download_artifacts(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:587
 [7] up(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Pkg.Types.UpgradeLevel) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1169
 [8] up(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; level::Pkg.Types.UpgradeLevel, mode::Pkg.Types.PackageMode, update_registry::Bool, kwargs::Base.Iterators.Pairs{Symbol,Base.DevNull,Tuple{Symbol},NamedTuple{(:io,),Tuple{Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:224
 [9] up(::Pkg.Types.Context; kwargs::Base.Iterators.Pairs{Symbol,Any,NTuple{4,Symbol},NamedTuple{(:level, :mode, :update_registry, :io),Tuple{Pkg.Types.UpgradeLevel,Pkg.Types.PackageMode,Bool,Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:192
 [10] resolve(::Pkg.Types.Context; kwargs::Base.Iterators.Pairs{Symbol,Base.DevNull,Tuple{Symbol},NamedTuple{(:io,),Tuple{Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:230
 [11] (::Pkg.Operations.var"#85#89"{String,Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.PackageSpec})() at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1367
 [12] with_temp_env(::Pkg.Operations.var"#85#89"{String,Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.PackageSpec}, ::String) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1286
 [13] (::Pkg.Operations.var"#84#88"{Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.Context,Pkg.Types.PackageSpec,String,Pkg.Types.Project,String})(::String) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1357
 [14] mktempdir(::Pkg.Operations.var"#84#88"{Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.Context,Pkg.Types.PackageSpec,String,Pkg.Types.Project,String}, ::String; prefix::String) at ./file.jl:682
 [15] mktempdir(::Function, ::String) at ./file.jl:680 (repeats 2 times)
 [16] sandbox(::Function, ::Pkg.Types.Context, ::Pkg.Types.PackageSpec, ::String, ::String, ::Pkg.Types.Project) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1324
 [17] build_versions(::Pkg.Types.Context, ::Array{Base.UUID,1}; might_need_to_resolve::Bool, verbose::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:873
 [18] build(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:788
 [19] build(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; verbose::Bool, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:698
 [20] build at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:682 [inlined]
 [21] #build#108 at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:680 [inlined]
 [22] build at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:680 [inlined]
 [23] #build#105 at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:677 [inlined]
 [24] build(::String) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:677
 [25] top-level scope at /Users/mo/.julia/dev/GAP/src/GAP.jl:19
 [26] include(::Module, ::String) at ./Base.jl:377
 [27] top-level scope at none:2
 [28] eval at ./boot.jl:331 [inlined]
 [29] eval(::Expr) at ./client.jl:449
 [30] top-level scope at ./none:3
in expression starting at /Users/mo/.julia/dev/GAP/src/GAP.jl:12
caused by [exception 1]
Unable to automatically install 'gap-4-11-0' from '/Users/mo/.julia/dev/GAP/Artifacts.toml'
Stacktrace:
 [2] ensure_artifact_installed(::String, ::Dict{String,Any}, ::String; platform::Pkg.BinaryPlatforms.Platform, verbose::Bool, quiet_download::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Artifacts.jl:894
 [3] ensure_all_artifacts_installed(::String; platform::Pkg.BinaryPlatforms.Platform, pkg_uuid::Nothing, include_lazy::Bool, verbose::Bool, quiet_download::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Artifacts.jl:958
 [4] download_artifacts(::Pkg.Types.Context, ::Array{String,1}; platform::Pkg.BinaryPlatforms.MacOS, verbose::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:609
 [5] download_artifacts(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; platform::Pkg.BinaryPlatforms.MacOS, verbose::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:588
 [6] download_artifacts(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:587
 [7] up(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Pkg.Types.UpgradeLevel) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1169
 [8] up(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; level::Pkg.Types.UpgradeLevel, mode::Pkg.Types.PackageMode, update_registry::Bool, kwargs::Base.Iterators.Pairs{Symbol,Base.DevNull,Tuple{Symbol},NamedTuple{(:io,),Tuple{Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:224
 [9] up(::Pkg.Types.Context; kwargs::Base.Iterators.Pairs{Symbol,Any,NTuple{4,Symbol},NamedTuple{(:level, :mode, :update_registry, :io),Tuple{Pkg.Types.UpgradeLevel,Pkg.Types.PackageMode,Bool,Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:192
 [10] resolve(::Pkg.Types.Context; kwargs::Base.Iterators.Pairs{Symbol,Base.DevNull,Tuple{Symbol},NamedTuple{(:io,),Tuple{Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:230
 [11] resolve(; kwargs::Base.Iterators.Pairs{Symbol,Base.DevNull,Tuple{Symbol},NamedTuple{(:io,),Tuple{Base.DevNull}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:228
 [12] (::Pkg.Operations.var"#85#89"{String,Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.PackageSpec})() at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1362
 [13] with_temp_env(::Pkg.Operations.var"#85#89"{String,Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.PackageSpec}, ::String) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1286
 [14] (::Pkg.Operations.var"#84#88"{Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.Context,Pkg.Types.PackageSpec,String,Pkg.Types.Project,String})(::String) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1357
 [15] mktempdir(::Pkg.Operations.var"#84#88"{Pkg.Operations.var"#59#63"{Bool,Pkg.Types.Context,String,Pkg.Types.PackageSpec,String},Pkg.Types.Context,Pkg.Types.PackageSpec,String,Pkg.Types.Project,String}, ::String; prefix::String) at ./file.jl:682
 [16] mktempdir(::Function, ::String) at ./file.jl:680 (repeats 2 times)
 [17] sandbox(::Function, ::Pkg.Types.Context, ::Pkg.Types.PackageSpec, ::String, ::String, ::Pkg.Types.Project) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1324
 [18] build_versions(::Pkg.Types.Context, ::Array{Base.UUID,1}; might_need_to_resolve::Bool, verbose::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:873
 [19] build(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Bool) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:788
 [20] build(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; verbose::Bool, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:698
 [21] build at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:682 [inlined]
 [22] #build#108 at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:680 [inlined]
 [23] build at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:680 [inlined]
 [24] #build#105 at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:677 [inlined]
 [25] build(::String) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:677
 [26] top-level scope at /Users/mo/.julia/dev/GAP/src/GAP.jl:19
 [27] include(::Module, ::String) at ./Base.jl:377
 [28] top-level scope at none:2
 [29] eval at ./boot.jl:331 [inlined]
 [30] eval(::Expr) at ./client.jl:449
 [31] top-level scope at ./none:3
ERROR: Failed to precompile GAP [c863536a-3901-11e9-33e7-d5cd0df7b904] to /Users/mo/.julia/compiled/v1.4/GAP/YqHod_wHM1S.ji.
Stacktrace:
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1272
 [3] _require(::Base.PkgId) at ./loading.jl:1029
 [4] require(::Base.PkgId) at ./loading.jl:927
 [5] require(::Module, ::Symbol) at ./loading.jl:922
```

Artifacts.toml
fingolfin: That's a SHA1 hash for the entire artifact directory. It is created using the same algorithm `git` uses to create "tree objects", hence the name.

fingolfin:
@mohamed-barakat the progress bar is generated by Julia 1.4.0, which also performs the download, so I have no control over that. It does work fine over here, though.

The error is odd. But that makes me wonder.. @ThomasBreuer @kalmarek: did you actually try to use this PR, or just reviewed it visually. While it works for me and Travis, it'd be interesting if other people have similar issues.

@mohamed-barakat Can you tell me the exact Julia version you are using? Can you show me the content of `~/.julia/artifacts`?

fingolfin:
@kalmarek the `deps/build_*` are not "necessary" in that yes, they'd be downloaded again; but why force users to re-download them again and again? I saw other Julia projects add them for this reason, too (including at least one instance where Tim Holy did that, and if it's good enough for him, it's good enough for me ;-).

In any case, I look forward to replacing these with using `FOO_jll` packages instead, but I think we should do that in all our packages at once, coordinated; and there is <https://github.com/oscar-system/LoadFlint.jl/pull/4> to take into account here, too...

kalmarek:
@fingolfin 
as a matter of fact I didn't try this one at home, I'll do later today, if time allows.

> including at least one instance where Tim Holy did that, and if it's good enough for him, it's good enough for me ;-).
Fair enough :D we just need to remember to update these periodically

> In any case, I look forward to replacing these with using FOO_jll packages instead, but I think we should do that in all our packages at once, coordinated;
Very much so!

fingolfin:
@mohamed-barakat I've tested this now on several computers and can't replicate. Since the SHA256 for the tarball seems to match (there is no error about that), I assume the download worked. The next thing Pkg then does is to extract the tarball into a fresh temporary directory. So some wild ideas what could go wrong:
1. The extraction somehow "fails" resp. produces a different output than elsewhere, e.g. because ...
  1a. ... the `tar` on your computer is somehow different and does something odd, or
  1b. ... macOS somehow inserts extra files (like it does when opening a directory in the finder and it adds `.DS_store` files
2. computing the tree hash somehow fails

To text variant 1, it would be good to get some extra information from you:

- output of `sw_vers`
- output of `which tar`
- content of `~/.julia/artifacts`
- perhaps the temporary dir Julia created is still around -- it'll be a subdir of `~/.julia/artifacts` which DOES NOT have a name that looks like a hexadecimal hash. Assuming you can identify such a directory, please run `/bin/ls -aR > /tmp/files.txt` then attach `/tmp/files.txt` here, so that I can compare it to what I get.

ThomasBreuer:
@fingolfin I have downloaded the pull request and notifying this version of GAP.jl to Julia 1.4.0 using `Pkg.develop` works:
The artifact gets downloaded, the progress bar works, and `using GAP` makes GAP available.

fingolfin:
@mohamed-barakat ah, one more idea by @fieker: could you please verify that your disk has enough free space (I assume it does, but who knows).

Another culprit I suspect is APFS resp. the macOS facilities that drives its own file versioning system and time machine; these have caused me headaches when using `git` in the past. Weird.

fingolfin:
@ThomasBreuer thank you. I also tried it on an Arch Linux server (plesken) in Siegen, and it worked fine there. I think this problem may somehow be specific to macOS 10.15, and/or there are other issues on Mohamed's computer

mohamed-barakat:
```
~/.julia/artifacts $ ls
43cc221a07e02bca83c81f004c0fa3c915802baa/
5f84ea130b97f5f163682f0f9bac5ebff45e7d18/
~/.julia/artifacts $ cd 5f84ea130b97f5f163682f0f9bac5ebff45e7d18
~/.julia/artifacts/5f84ea130b97f5f163682f0f9bac5ebff45e7d18 $ /bin/ls -aR > /tmp/files.txt
```
[files.txt](https://github.com/oscar-system/GAP.jl/files/4450135/files.txt)

mohamed-barakat:
I have Julia Version 1.4.0 (2020-03-21) running on Mac OS X 10.5.3. I am planning to upgrade today or tomorrow to 10.5.4.

mohamed-barakat:
> @mohamed-barakat ah, one more idea by @fieker: could you please verify that your disk has enough free space (I assume it does, but who knows).

Yes, I have 6TB free :)
 
> Another culprit I suspect is APFS resp. the macOS facilities that drives its own file versioning system and time machine; these have caused me headaches when using `git` in the past. Weird.

I attached `files.txt` above and it was unable to find any Mac OS X specific files there, e.g. `.DS_Store`.

fingolfin:
@mohamed-barakat thanks. Please tell me the output of `which tar`, as well.

Also, could you please do this in a temp dir (please use Julia 1.4 in the final command)
```bash
curl -L -O https://github.com/gap-system/gap/releases/download/v4.11.0/gap-4.11.0.tar.bz2
mkdir foo
cd foo
tar xf ../gap-4.11.0.tar.bz2
julia -e 'using Pkg; println(Base.SHA1(Pkg.GitTools.tree_hash(pwd())))'
```

mohamed-barakat:
```
$ tar --version
bsdtar 3.3.2 - libarchive 3.3.2 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.6
```

fingolfin:
"Good" news, I can reproduce the issue on Travis! I've some leads now -- I strongly suspect BSD tar vs GNU tar now; thanks to the instructions I just wrote down for Mohamed, I was now able to verify this, and now I can reproduce it, too! So I can now debug this locally, which is a great improvement.

mohamed-barakat:
> "Good" news, I can reproduce the issue on Travis! I've some leads now -- I strongly suspect BSD tar vs GNU tar now; thanks to the instructions I just wrote down for Mohamed, I was now able to verify this, and now I can reproduce it, too! So I can now debug this locally, which is a great improvement.

Thank you very much @fingolfin :)

```
/tmp $ curl -L -O https://github.com/gap-system/gap/releases/download/v4.11.0/gap-4.11.0.tar.bz2
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   608  100   608    0     0   1006      0 --:--:-- --:--:-- --:--:--  1004
100  386M  100  386M    0     0  2326k      0  0:02:50  0:02:50 --:--:-- 2494k
/tmp $ mkdir foo
/tmp $ cd foo
/tmp/foo $ tar xf ../gap-4.11.0.tar.bz2
/tmp/foo $ julia -e 'using Pkg; println(Base.SHA1(Pkg.GitTools.tree_hash(pwd())))'
5f84ea130b97f5f163682f0f9bac5ebff45e7d18
```

fingolfin:
OK, got it, it's as I suspected:
```
$ diff -ru with-gnutar with-bsdtar | head -n 10
Only in with-gnutar/gap-4.11.0/pkg/matgrp/doc: ._manual.tex
Only in with-gnutar/gap-4.11.0/pkg/matgrp/htm: ._chapters.htm
Only in with-gnutar/gap-4.11.0/pkg/qpa-version-1.30: ._CHANGES
Only in with-gnutar/gap-4.11.0/pkg/qpa-version-1.30: ._PackageInfo.g
Only in with-gnutar/gap-4.11.0/pkg/qpa-version-1.30: ._README
Only in with-gnutar/gap-4.11.0/pkg/qpa-version-1.30: ._version
Only in with-gnutar/gap-4.11.0/pkg/qpa-version-1.30/doc: ._qpadocumentation.xml
Only in with-gnutar/gap-4.11.0/pkg/transgrp: ._LICENSE
Only in with-gnutar/gap-4.11.0/pkg/transgrp: ._data
Only in with-gnutar/gap-4.11.0/pkg/transgrp: ._htm
```
So there are metadata files in there, which Apple's fork of BSD tar inserts. Upon extracting the tarball, Apple's tar does not extra these `._*` files but instead uses them to set metadata. GNU tar knows nothing about this and happily extracts the file.

I am afraid that means I'll have to create custom GAP tarballs which I cleaned up to remove these files. Will also notify the package authors about this, and for the next GAP release, we'll filter out such files, and reject package updates that contain them.

mohamed-barakat:
Thanks for investigating.

mohamed-barakat:
It worked for me with commit  #76f8851:

```julia
$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.4.0 (2020-03-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(@v1.4) pkg> build GAP
Downloading artifact: gap-4-11-0
```

mohamed-barakat:
Works for me with commit 0ad9082

PR #409, opened 2020-04-13
keyword arguments: something that had been missing

ThomasBreuer:
- ~~added a `gap_to_julia` method for the non-recursive conversion
  of GAP records to Julia dictionaries~~
- added an example of `CallJuliaFunctionWithKeywordArguments` where
  the keyword arguments are proper Julia objects
- ~~added comments (in the code) about the current meaning of the
  first arguments `::Type{Array{Obj,1}}` and `::Type{Dict{Symbol,Obj}}`
  for `gap_to_julia`~~
- replaced the individual Julia code for creating polynomial rings
  by the use of `CallJuliaFunctionWithKeywordArguments`

(The removals are due to a better solution which is meanwhile available by #430.)

### DIFF
--- a/pkg/JuliaExperimental/gap/singular.g
+++ b/pkg/JuliaExperimental/gap/singular.g
@@ -212,16 +212,13 @@ InstallMethod( ContextGAPSingular,
       degree_bound:= 0;
     fi;
 
-    # Create the Singular ring.
-    # If we would be able to deal with keyword arguments
-    # then wrapping would not be needed here.
-    dict:= GAPToJulia( rec( ring:= FContext!.JuliaDomainPointer,
-                          indeterminates:= names,
-                          cached:= true,
-                          ordering:= JuliaSymbol( ordering ),
-                          ordering2:= JuliaSymbol( ordering2 ),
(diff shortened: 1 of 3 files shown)


src/gap_to_julia.jl
```diff
@@ -268,6 +275,31 @@ function gap_to_julia( ::Type{T}, obj::GapObj ) where T <: StepRange
 end
 
 ## Dictionaries
+##
+## Note that the argument '::Type{Dict{Symbol,Obj}}' just means that
+## the values of the returned dictionary are the components of the given
+## GAP record,
+## that is, they are not converted recursively to Julia.
+## (If these components are already Julia objects then the returned dictionary
+## does *not* have the type 'Dict{Symbol,Obj}'.)
```
fingolfin, line 284: Same remarks as for arrays apply here.

fingolfin:
@ThomasBreuer ping?

fingolfin:
@ThomasBreuer do you plan to update this PR?

ThomasBreuer:
@fingolfin yes, since #430 has solved the questions how to address recursive and non-recursive conversion, this pull request can be updated

PR #410, opened 2020-04-13
replaced `syJmp_buf` by `jmp_buf`,

ThomasBreuer:
in order to adjust GAP.jl to GAP's pull request https://github.com/gap-system/gap/pull/3938


ThomasBreuer:
should be obsolete because of #436

PR #421, opened 2020-04-17
Switch to using JLLs

### LINKED ISSUE #415: REPL editing broken in $(julia -e 'import GAP; print(GAP.gap_exe())')
The backspace or the arrows corrupt the input line.

fingolfin:
Note that this means we'll load an extra copy of the GMP for now. If this turns out to be a problem, I could drop the GMP_jll bit again.

This is not done; in particular, I am experimenting more with how to launch the GAP inside GAP.jl. Right now I am experimenting with just launching it via julia: In particular, instead of `$(julia -e 'import GAP; print(GAP.gap_exe())')` one can do `import GAP; GAP.Globals.SESSION()` to get a GAP prompt. This is missing a banner (easy to fix) and also doesn't handle command line arguments (slightly more work to fix, but doable)

Fixes #415


mohamed-barakat:
This is really nice:
```
import GAP; GAP.Globals.SESSION()
```
Is it possible to assign GAP.Globals.SESSION to an exported function called `gap` or `startgap`? I would suggest following the sage convention here:
```
using GAP
gap()
gap>
```

fingolfin:
That hypothetical `gap()` function should do more things, too; e.g. it could turn the break loop back on temporarily

Issue #390, opened 2020-03-24
access to Julia variables from GAP


ThomasBreuer:
Currently Julia module objects in GAP admit access to variables of the module via `\.`.
```
gap> Julia;
<Julia module Main>
gap> Julia.GAP;
<Julia module GAP>
gap> Julia.sqrt;
function( arg... ) ... end
gap> JuliaEvalString( "a = 1" );
1
gap> Julia.a;
1
```
One can access Julia variables also with `JuliaGetGlobalVariable`, but there are differences.
First, there is a problem with caching.
```
gap> JuliaEvalString( "a = 2" );
2
gap> Julia.a;
1
gap> JuliaGetGlobalVariable( "a" );
2
```
Second, functions and modules get wrapped with `_JuliaFunction` and `_WrapJuliaModule` by the access with `\.` (see above) but not by `JuliaGetGlobalVariable`.
```
gap> JuliaGetGlobalVariable( "sqrt" );
<Julia: sqrt>
gap> JuliaGetGlobalVariable( "GAP" );
<Julia: Main.GAP>
```
(Third, see #175 .)

The caching problem must be solved, the easiest solution is to remove the caching.
Concerning the other differences, what do we want?
Which of `<Julia: sqrt>` and `Julia.sqrt` is better?

fingolfin:
Yeah, let's remove the caching.

As to the handling of functions: `Julia.sqrt` is nicer in the sense that it can be used as a function more efficiently. The result of `JuliaGetGlobalVariable( "sqrt" )` is nicer in that it prints nicer. So here my solution would be to add nicer `(Print|View|Display)(String)?` methods; to do that, easiest would be to export `IS_JULIA_FUNC` from the kernel to GAP, and then install high-ranked methods for `IsFunction` which check `IS_JULIA_FUNC` and then either print `<Julia: FOOBAR>`, or else call `TryNextMethod` .

Finally, once these improvements are in place, I am not sure whether we even want to keep `JuliaGetGlobalVariable`? The only reason I can think of (beyond backwards compatibility, of course!) is performance: is perhaps `JuliaGetGlobalVariable("a")` considerably faster than `Julia.a`, if called "often enough"?

ThomasBreuer:
Pull request #423 deals with the (hopefully) non-controversial parts of this issue:

- Caching the values of arbitrary Julia variables in GAP's `IsJuliaModule` objects cannot work because the values may be changed in Julia.
[...]

PR #423, opened 2020-04-27
access to Julia variables from GAP

ThomasBreuer:
(This deals with parts of #390.)
- made `IS_JULIA_FUNC` accessible from GAP
- added a `ViewString` method for GAP functions created by `_JuliaFunction`
- fixed the documentation of `JuliaGetGlobalVariable`,
  the returned value can be a GAP object as well
- set also the `String` attribute in `IsJuliaModule` objects
  (`String( Julia )` yields `"<object>"` otherwise)
- removed the caching of arbitrary Julia variables in `IsJuliaModule` objects,
  because the value may change on the Julia side;
  caching for Julia modules is kept
- forbid `Unbind\.` in `IsJuliaModule` objects

### DIFF
--- a/pkg/JuliaInterface/gap/JuliaInterface.gi
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gi
@@ -30,16 +30,15 @@ BindGlobal( "_JULIA_ISA", JuliaFunction( "isa" ) );
 
 BindGlobal( "_WrapJuliaModule",
   function( name, julia_pointer )
-    local module;
+    local str;
 
-    module := rec( storage := rec( ) );
+    str:= Concatenation( "<Julia module ", name, ">" );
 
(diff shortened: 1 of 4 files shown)


pkg/JuliaInterface/gap/JuliaInterface.gi
```diff
@@ -96,12 +104,14 @@ end );
...
   function( module, rnum )
-    Unbind\.( module!.storage, rnum );
+    Error( "cannot unbind Julia variables" );
 end );
 
 
 InstallGlobalFunction( JuliaSymbolsInModule,
   module -> RecNames( module!.storage ) );
+#T This makes no sense anymore,
+#T after removing the caching for arbitrary Julia variables.
```
fingolfin, line 114: So let's just remove it? Nothing uses it

PR #427, opened 2020-04-27
Start GAP via Julia

fingolfin:
This was borne out of need but now actually seems quite nice...

The need was that in order to use JLLs, we need to link GAP against libraries
in various prefixes; prefixes which depend on artifacts, and thus on the hash
codes of those artifacts, which can and do change as the JLLs get updated.

So in order to load the "regular" GAP binary, we'd have to set OS specific
environment variables to point the dynamic linker at the right lib
directories. But the values for those variables can change, so we can't just
put them into the generated gap.sh file! We need to query Julia for them...
but to do that, we need to load those JLLs. But we can't just do that via
`using`, as they might not have been added to the current environment! The
only thing we can reliably "use" is GAP.jl itself. But if we do that, we are
effectively loading a full copy of GAP already! So if we go that way, we
immediately *double* the startup time of GAP.

So the logical next step is: why not "use" the GAP we are loading anyway?
This way, we also don't need to mess with passing flags to the dynamic linker
via environment variables; Julia takes care of all of that for us!

Note that this part of PR #421 but should work by its own, so I am submitting it as a separate PR to test that theory.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -92,9 +92,33 @@ println("Generating gap.sh ...")
 
 gap_sh_path = joinpath(gap_bin_root, "bin", "gap.sh")
 write(gap_sh_path,"""
-#!/bin/sh
+#!/bin/bash
+# This is a a Julia script which also is a valid bash script; if executed by
+# bash, it will execute itself by invoking `julia`. Of course this only works
+# right if `julia` exists in the PATH and is the "correct" julia executable.
+# But you can always instead load this file as if it was a .jl file via any
+# other Julia executable.
+#=
+exec julia --startup-file=no -- "\${BASH_SOURCE[0]}" "\$@"
+=#
 
-exec "$(gap_bin_root)/gap" -l "$(extra_gap_root);$(gap_src_root)" "\$@"
+# pass command line arguments to GAP.jl via a small hack
+ENV["GAP_SHOW_BANNER"] = "true"
+__GAP_ARGS__ = ARGS
+using GAP
+
+# GAP.jl passes --norepl to GAP, which means that init.g never
+# starts a GAP session; we now run one "manually". Note that this
+# may throw a "GAP exception", which we need to catch; thus we
+# use Call0ArgsInNewReader to perform the actual call.
+ccall(:Call0ArgsInNewReader, Cvoid, (Any,), GAP.Globals.SESSION)
+
+# call GAP's "atexit" cleanup functions
+ccall(:Call0ArgsInNewReader, Cvoid, (Any,), GAP.Globals.PROGRAM_CLEAN_UP)
+
+# Finally exit by calling GAP's FORCE_QUIT_GAP(). See comments in GAP.jl for
+# an explanation of why we do it this way.
+GAP.Globals.FORCE_QUIT_GAP()
 """)
 chmod(gap_sh_path, 0o755)
 
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -96,7 +96,15 @@ end
 function initialize(argv::Array{String,1})
     lib = joinpath(GAPROOT, ".libs", "libgap")
     gap_library = Libdl.dlopen(lib, Libdl.RTLD_GLOBAL)
-    error_handler_func = @cfunction(error_handler, Cvoid, ())
+    handle_signals = isdefined(Main, :__GAP_ARGS__)  # a bit of a hack...
+    error_handler_func = handle_signals ? C_NULL : @cfunction(error_handler, Cvoid, ())
+
+    # Initialize __JULIAINTERNAL_LOADED_FROM_JULIA; this also allows us to
+    # detect whether GAP_Initialize and GAP's `init.g` completed successfully
+    # (if they didn't, then this GAP code won't be called, which we can easily
+    # detect by checking for the value of __JULIAINTERNAL_LOADED_FROM_JULIA).
+    append!(argv, ["-c", """BindGlobal("__JULIAINTERNAL_LOADED_FROM_JULIA", true );"""])
+
     ccall(
         Libdl.dlsym(gap_library, :GAP_Initialize),
         Cvoid,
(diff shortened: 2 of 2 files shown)


deps/build.jl
```diff
@@ -109,8 +109,20 @@ println("Generating gap.sh ...")
...
 write(gap_sh_path,"""
-#!/bin/sh
-
-exec "$(gap_bin_root)/gap" -l "$(extra_gap_root);$(gap_src_root)" "\$@"
+#!/bin/bash
+# This is a a Julia script which also is a valid bash script; if executed
+# by bash, it'll execute itself via julia.
+#=
+export GAP_SHOW_BANNER=true
+exec julia --color=yes --startup-file=no "\${BASH_SOURCE[0]}" "\$@"
```
fingolfin, line 117: Here's a problem with that: it hardcodes `julia` here :/. I guess I could make this dynamic by using `Sys.BINDIR` here; but then it'd be specific on the Julia version that was used to compile it. Which is reasonable, but it *could* break.

Then again, the old approach by launching the `gap` exe suffered from the exact same problem as it had to hardcode the rpath of libjulia; so I guess using Sys.BINDIR it is for now.

BTW, note that this `gap.sh` file actually is *also* a valid Julia script, so one could invoke it that way, too. 

However, we now are running into a bit of a chicken-and-egg kind of problem: I suggested to e.g. @mohamed-barakat to use `$(julia -e 'import GAP; print(GAP.gap_exe())')` to invoke the GAP-in-GAP.jl; well, we could change that to  `julia --color=yes --startup-file=no PATH/TO/gap.sh  "#$@"` except that we don't know `PATH/TO/gap.sh`... to figure that out ... we need to load GAP.jl. But if we do that, we are loading it "in the wrong way", so we can't just go back the REPL mode, and we can't just pass arguments to it... HRMM

Perhaps I can tweak things so that  we *can* enter the GAP prompt after loading GAP.jl "normally". For that I'd need at least...
- a way to reset the GAP error handler (should be done in the GAP kernel, but I can also hack it into JuliaInterface for now
- reset `ERROR_OUTPUT` to its standard value
- ... ?

As an added benefit, we could then perhaps also add a nice `GAP.prompt()` proper around `GAP.Globals.SESSION()` which would do the same?

In the meantime, though, I noticed more issue with this PR:
- ctrl-C crashes; we ought to add `--handle-signals=no` to the arguments -- but then somehow ctrl-C does nothing anymore?
- something is broken with input / line editing
- ... needs much more testing ...

fingolfin: I added `--handle-signals=no`, and also reactivated ctrl-c handling in GAP, and now handling `SIGINT` in GAP-via-Julia works on Linux. It is still broken on macOS, though. I thought that it was perhaps doing something macOS/Darwin/Mach-O specific, but I see nothing. I did notice that Julia has a dedicated thread for handling signals, including SIGINT, which uses `kqueue`. So perhaps that supersedes the signal handling in GAP? But then why only on macOS, not on Linux? Unclear to me. Perhaps @rbehrends has an idea?

fingolfin: Another issue: starting this "GAP-in-Julia" sometimes segfaults. Unfortunately this happens very infrequently which makes it difficult to debug.

fingolfin:
Positive: Running `testinstall.g` seems to work fine! Except it dies at the end:
```
...
#I  No errors detected while testing

Panic in /home/mhorn/.julia/artifacts/848f0f318405266896d24d11e4baf5cf7bf3f203/gap-4.11.0/src/gap.c:1742: Caught error at top-most level, probably quit from library loading
```
But that may be simply because we have a "dangling setjmp" here: the error location in gap.c is I think the last time we use `TRY_IF_NO_ERROR` (which uses `setjmp`) in the GAP kernel startup code before returning to Julia ; unlike in a normal GAP session, reading `init.g` completes immediately. Now we call GAP from Julia; but we never set this `setjmp` buffer again. Then the test suite terminates by calling `QUIT_GAP`; but that in turn is implemented in the kernel by calling `ReadEvalError();` -- but this does a `longjmp` to the target set by that "last" `setjmp` (it's not actually the last, but it's the last at the toplevel, all others return the jmp buffer back to the value set by that `setjmp` invocation). Solution: we could invoke `setjmp` in GAP.jl or `JuliaInterface`, pointing it at a "safe" bit of code? Note that this will also affect regular `GAP.jl`, if a user calls `GAP.Globals.QUIT_GAP` (which probably is a bad idea, though ;-).

Anyway, I think I can fix this. I just wanted to make a note here so I won't forget.

deps/build.jl
rbehrends: There have been a lot of changes to `SIGINT` handling in Julia, which make it difficult to track down the exact behavior; basically, the [logic was changed](https://github.com/JuliaLang/julia/pull/16174) to deliver `SIGINT` only during GC safepoints (deferring them until then) and that affected overall signal handling logic. Also, with pthreads there are technically no guarantees which thread gets a signal, which may also result in system-specific differences. This is something I'll have to look into.

fingolfin:
Passing the `-h` option crashes, because that invokes `QUIT_GAP` while `GAP_Initialize` is running, and we are just *not* prepared for that. I guess one may consider this to be a flaw in libgap / `GAP_Initialize` which has no way to indicate this happened... Perhaps checking the GAP variable `QUITTING` right after the call to `GAP_Initialize` might work?

fingolfin:
@rbehrends @ThomasBreuer @mohamed-barakat  This PR seems to be mostly ready. It would be good if you could try it out to see if it "works" for you. I.e., both using GAP from Julia, and as "standalone". This should also fully resolve issue #415.

Also, I'd appreciate some input on the following problem: Namely, I wonder (once again) how we should suggest to people that they start the GAP bundled with GAP.jl, and how to make that most efficient... Right now, in the GAP.jl master branch, one can do this:
```
$(julia -e 'import GAP; print(GAP.gap_exe())')
```
This works, and still works with this PR -- but we really are loading GAP twice with that. The question is how bad of a problem that is. I can think of ways to avoid this, but they'll make that script more complicated (e.g. I could add a global var which GAP.jl considers when loading; if it is set, it prints the required path and immediately aborts Julia; the invocation could then look something like this:
```
$(julia -e '__GAP_PRINT_EXE_PATH_AND_ABORT=true ; import GAP')
```

The question is, who are the people that will want to use this "standalone" GAP, and how important is it for "them" to have a "short" and "simple" script to start standalone-GAP? How important is the startup speed? 

Perhaps I am worrying over nothing?

mohamed-barakat:
Excellent. Command-line editing works in all three cases:

```julia
(@v1.4) pkg> build GAP

julia> using GAP

julia> GAP.Globals.SESSION()
gap>
```

```bash
 # $(julia -e 'import GAP; print(GAP.gap_exe())')
 ┌───────┐   GAP 4.11.0 of 29-Feb-2020
 └───────┘   Architecture: x86_64-apple-darwin19.4.0-julia64-kv7-v1.4
 Configuration:  gmp 6.1.2, Julia GC, Julia 1.4.1, readline
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2019.09.04, AutPGrp 1.10.2, CRISP 1.4.5, Cryst 4.1.23, CrystCat 1.1.9, CTblLib 1.2.2, FactInt 1.6.3, FGA 1.4.0,
             Forms 1.2.5, GAPDoc 1.6.3, genss 1.6.6, IO 4.7.0, IRREDSOL 1.4, LAGUNA 3.9.3, orb 4.8.3, Polenta 1.3.9, Polycyclic 2.15.1, PrimGrp 3.4.0, RadiRoot 2.8,
             recog 1.3.2, ResClasses 4.7.2, SmallGrp 1.4.1, Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 2.0.5, utils 0.69
gap>
```

```bash
# ~/.julia/dev/GAP/deps/build/gap-4.11.0-julia-1.4/bin/gap.sh
 ┌───────┐   GAP 4.11.0 of 29-Feb-2020
 └───────┘   Architecture: x86_64-apple-darwin19.4.0-julia64-kv7-v1.4
 Configuration:  gmp 6.1.2, Julia GC, Julia 1.4.1, readline
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2019.09.04, AutPGrp 1.10.2, CRISP 1.4.5, Cryst 4.1.23, CrystCat 1.1.9, CTblLib 1.2.2, FactInt 1.6.3, FGA 1.4.0,
             Forms 1.2.5, GAPDoc 1.6.3, genss 1.6.6, IO 4.7.0, IRREDSOL 1.4, LAGUNA 3.9.3, orb 4.8.3, Polenta 1.3.9, Polycyclic 2.15.1, PrimGrp 3.4.0, RadiRoot 2.8,
             recog 1.3.2, ResClasses 4.7.2, SmallGrp 1.4.1, Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 2.0.5, utils 0.69
gap>
```

src/GAP.jl
```diff
@@ -96,22 +104,62 @@ function initialize(argv::Array{String,1})
...
+        "__JULIAINTERNAL_LOADED_FROM_JULIA",
     )
+    if val == C_NULL
+        # HACK: it would be better to call Julia's `exit` here; but
+        # unfortunately we can't access GAP's exit code, which is stored in
+        # `SystemErrorCode`, a statically linked (and hence invisible to us)
+        # GAP C kernel variable. Hence we instead call FORCE_QUIT_GAP with no
+        # arguments, which just calls the `exit` function of the  C standard
+        # library with the appropriate exit code.
+        # This is complicated by the fact that we can't
```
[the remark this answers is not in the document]
fingolfin: Fixed. I also rewrote this comment and the one in gap.sh -- I checked and GAP actually calls Julia's `jl_atexit_hook` so I think we are fine exiting this way.

Issue #334, opened 2020-01-30
Turn off GAP loading banner by default


fingolfin:
While it has been (and still is) useful for debugging while working on GAP, I think the convention for Julia packages is to not show that. Perhaps we can make it optional somehow.

That said, without this, I wouldn't have noticed issue #333.

ThomasBreuer:
This has been done via #342 .

PR #429, opened 2020-04-29
Show banner again by default

fingolfin:
... unless we are being loaded from Oscar (code for that taken from
Singular.jl); can (still) be overridden with the `GAP_SHOW_BANNER`
environment variable.

This is based on the recent PI decision on banners.

Note that this simply shows the default GAP banner, e.g.:
```
julia> using GAP
 ┌───────┐   GAP 4.11.0 of 29-Feb-2020
 └───────┘   Architecture: x86_64-apple-darwin18.7.0-julia64-kv7-v1.4
 Configuration:  gmp 6.1.2, Julia GC, Julia 1.4.1, readline
 Packages:   GAPDoc 1.6.3, PrimGrp 3.4.0, SmallGrp 1.4.1, TransGrp 2.0.5
```
I could change it to a custom banner, of course (discussion on that is ongoing between the PIs)

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -159,11 +159,37 @@ end
 function run_it()
     gaproots = abspath(joinpath(@__DIR__, "..")) * ";" * sysinfo["GAP_LIB_DIR"]
     cmdline_options = ["", "-l", gaproots, "-T", "-A", "--nointeract", "-m", "1000m"]
-    if get(ENV, "GAP_SHOW_BANNER", "false") != "true"
+    if haskey(ENV, "GAP_SHOW_BANNER")
+        show_banner = ENV["GAP_SHOW_BANNER"] == "true"
+    else
+        show_banner = isinteractive() &&
+                !any(x->x.name in ["Oscar"], keys(Base.package_locks))
+    end
+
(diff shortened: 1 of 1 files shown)


mohamed-barakat:
I also would prefer to see the banner. I am also thinking of a banner for `HomalgProject` which acknowledges the used computer algebra systems but does not show their versions to avoid in example files. But in an interactive session, it is nice if people see the components.

fingolfin:
@ThomasBreuer fixed the gap_module issue

fieker:
On Thu, Apr 30, 2020 at 07:50:43AM -0700, ThomasBreuer wrote:
> @ThomasBreuer requested changes on this pull request.
> 
> `gap_module` is a local variable in `__init__`,
> perhaps use it as an argument of `run_it` in order to use it inside this function.
> 
> My personal preference is not to show banners, but if there is an
> agreement that every ornerstone shall show a banner then I do not
> object.  (Wasn't there a convention for Julia packages to not show
> banners, see #334?)
Yes to both. We, as PIs decided that our end-products have banners -
when used interactively
And yes, julia decided against a banner - except for Julia.

Issue #428, opened 2020-04-27
Add constructors for conversion from GAP to Julia


fingolfin:
We should salvage at least part of PR #323 by @kalmarek ! That is, I very much would like to have constructors for BigInt, BigFloat, Float64, Int64, etc. etc. that accept a `GapObj`, and which would throw a `GAP.ConversionError` in case of a failure. I'd also refactor `gap_to_julia` to use these constructor where appropriate / sensible (just as PR #323 did).

At least for the basic types, this should be easy enough. For containers, all kinds of pitfalls lurk, as also discussed in issue #319, but that shouldn't stop us from at least providing the basic ones. I.e. this is not an "all-or-nothing" situation, we can do it gradually.

@ThomasBreuer can you look into this?

ThomasBreuer:
@fingolfin Yes

fingolfin:
This is what I meant when I wrote:
> At least for the basic types, this should be easy enough. For containers, all kinds of pitfalls lurk [...]

I.e., for starters I'd already be happy if we had at least constructors for Int, BigInt, Float, BigFloat, String, ... even if we had none for containers (at least at first)

BTW, I would not use a second argument for constructor to pass the recursion dictionary; I'd rather use a keyword argument, just as we do with `gap_to_julia` methods.

Anyway, In the end, I don't care so much whether constructors are implemented using `gap_to_julia`, or the way around, or a mixture, as long as it works and is understandable :-). If you think writing the constructors using `gap_to_julia` is preferable, that's good enough for me.

BTW, one could *also* think about providing `convert` methods -- see also <https://discourse.julialang.org/t/4159/24> and <https://discourse.julialang.org/t/4159/35>

ThomasBreuer:
@fingolfin Can this issue be closed?

Issue #319, opened 2020-01-20
refactor gap_to_julia


kalmarek:
mostly valid julia code but some pseudocode as well; comments welcome;
EDIT: draft is in https://github.com/oscar-system/GAP.jl/pull/323

kalmarek:
* overloading julia constructors allow much more natural calls like `Matrix(obj)` than `gap_to_julia(Matrix, obj)`

I updated the code @fingolfin @ThomasBreuer; With hacks/backward compatibility at the end it passes the first ~ 60 tests in `test/conversion.jl`
I won't have much time this or next week to push it further, I'd be glad if anyone from `GAP.jl` take it over;

ThomasBreuer:
@kalmarek @fingolfin I try to incorporate the ideas.

kalmarek:
given what @fingolfin wrote https://github.com/oscar-system/GAP.jl/issues/315#issuecomment-577244963, I think this requires more consideration and overall time spent in design, as for now we have two intermingled layers of conversion;

In Polymake.jl we convert everything to julia understandable type whenever we can. the fast-track (without conversion) is hidden in low-level interface behind keyword, but I'd need to know more about gap↔julia interface...

fingolfin:
@kalmarek could you perhaps turn this into a (non-working) PR? I.e. just copy&past your edits into/over `adapter.jl`, it's fine if it doesn't compile. But then this way, it's easy to inspect the diff, and also to comment on it line by line.

For GAP<->Julia, based on our experience designing several similar tools, Sebastian and me followed the design principle that at the lowest level, automatic conversions are kept to an absolute minimum. Higher level code can of course implement further conversions; but if these turn out to be "wrong" for some usecase, or inefficient, it's important to provide an escape hatch. Many interfaces between disparate systems unfortunately get this wrong and this then causes massive headaches down the road. I described this design philosophy in greater detail here: https://github.com/oscar-system/GAP.jl/wiki/Rules-for-conversion-between-GAP-and-Julia-objects

kalmarek:
yes, I've read this document, but it documents more wishes than actual state (some containers are converted, some part of containers are wrapped, but those parts as standalone are converted?).
```julia
julia> GAP.julia_to_gap(Dict{Symbol, Any}(:aaa=>1, :bbb=>Int32(3)))
GAP: rec( aaa := 1, bbb := <Julia: 3> )

[...]

PR #430, opened 2020-04-29
constructors for conversion to Julia, bugfixes

ThomasBreuer:
(using a lot of ideas from the code proposed by @kalmarek in #319)

- added methods for constructors (`Int64`, `BigInt`, `Tuple`, `Vector`, `Dict`, ...) taking a `GapObj` as its first argument;
  when both recursive and non-recursive conversion make sense, one can choose via the keyword argument `recursive`;
  the idea is to leave the handling of identical subobjects to `gap_to_julia`, and to call `gap_to_julia` by the constructors (see the comments for #428)

- changed `gap_to_julia`:
  - the generic variant `gap_to_julia(::GapObj)` throws an error if no conversion is provided for the argument; up to now, the argument had been returned
  - the decision about recursive or non-recursive conversion (of GAP lists and records) is now controlled by the new keyword argument `recursive`;
    up to now, non-recursive conversion of GAP lists was possible by choosing the target type `Array{GAP.Obj,1}`, which was logically not correct (since a GAP list or record may contain proper Julia objects)
  - fixed some of the conversion methods, added missing ones

- extended the tests for `gap_to_julia`

- added tests for the constructors, analogous to the tests for `gap_to_julia`


src/gap_to_julia.jl
```diff
@@ -78,164 +94,153 @@ function gap_to_julia(::Type{Rational{T}}, x::Int64) where T <: Integer
...
-    if Globals.IsStringRep(obj)
-        return CSTR_STRING(obj)
-    elseif Globals.IsString(obj)
-        return CSTR_STRING( Globals.CopyToStringRep( obj ) )
-    else
-        throw(ArgumentError("<obj> is not a string"))
-    end
+## Strings
+function gap_to_julia(::Type{String}, obj::GapObj)
+    Globals.IsStringRep(obj) && return CSTR_STRING(obj)
```
kalmarek, line 125: what does `UNSAFE_CSTR_STRING` return? if there is a way to expose `Ptr{UInt8}` (i.e. pointer to raw memory), maybe `unsafe_string` would be better. maybe.

src/gap_to_julia.jl
```diff
@@ -78,164 +94,153 @@ function gap_to_julia(::Type{Rational{T}}, x::Int64) where T <: Integer
...
+        recursion_dict[obj] = new_array
+        for i in 1:len_list
+            current_obj = ElmList(obj, i)  # returns 'nothing' for holes in the list
+            if recursive
+                new_array[i] = get!(recursion_dict, current_obj) do
+                    gap_to_julia(T, current_obj, recursion_dict;
+                                 recursive=true)
+                end
+            else
+                new_array[i] = current_obj
```
kalmarek, line 167: does this even work with `current_obj = nothing`?
```julia
julia> x = Vector{Int}(undef, 3)
3-element Array{Int64,1}:
 139983047916960
 139983047177608
 139983047843440

julia> x[1] = nothing
ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type Int64
Closest candidates are:
  convert(::Type{Int64}, ::Polymake.Integer) at /home/kalmar/.julia/dev/Polymake/src/convert.jl:55
  convert(::Type{Int64}, ::Polymake.Rational) at /home/kalmar/.julia/dev/Polymake/src/convert.jl:58
  convert(::Type{JT}, ::CT) where {JT<:Number, CT<:Union{CxxWrap.CxxWrapCore.CxxSigned, CxxWrap.CxxWrapCore.CxxUnsigned}} at /home/kalmar/.julia/packages/CxxWrap/LfTHV/src/CxxWrap.jl:114
  ...
Stacktrace:
 [1] setindex!(::Array{Int64,1}, ::Nothing, ::Int64) at ./array.jl:826
 [2] top-level scope at REPL[14]:1
```

src/gap_to_julia.jl
```diff
@@ -78,164 +94,153 @@ function gap_to_julia(::Type{Rational{T}}, x::Int64) where T <: Integer
...
+    if Globals.IsMatrixObj(obj)
+        if ! haskey(recursion_dict, obj)
+            elm = Globals.MatElm
+            nrows = Globals.NumberRows(obj)
+            ncols = Globals.NumberColumns(obj)
+            new_array = Matrix{T}(undef, nrows, ncols)
+            recursion_dict[obj] = new_array
+            for i in 1:nrows
+                for j in 1:ncols
+                    current_obj = elm(obj, i, j)
```
kalmarek, line 185: does it return `nothing` for missing entries?

kalmarek: or there are no missing entries in GAP Matrices?

src/gap_to_julia.jl
```diff
@@ -78,164 +94,153 @@ function gap_to_julia(::Type{Rational{T}}, x::Int64) where T <: Integer
...
-                gap_to_julia(T,current_obj, recursion_dict)
+    elseif Globals.IsList(obj)
+        if ! haskey(recursion_dict, obj)
+            nrows = length(obj)
+            ncols = nrows == 0 ? 0 : length(obj[1])
+            new_array = Matrix{T}(undef, nrows, ncols)
+            recursion_dict[obj] = new_array
+            for i in 1:nrows
+                for j in 1:ncols
+                    current_obj = ElmList(ElmList(obj, i), j)
```
kalmarek, line 206: well, this does return `nothing`

src/gap_to_julia.jl
fingolfin: A list with holes cannot be converted into an Array{Int64,1} unambiguously, so triggering an error is IMHO the only correct action. Of course we  still might want to test foe it explicitly and throw a dedicated error, but that's a minor detail in my eyes.

Of course one can define conversions for this case, eg "skip all holes" (can be achieved by calling GAP's `Compacted` function before the conversion), but this are not "natural" (e.g. array indices change), so I don't think it's a good idea to use them automatically

src/gap_to_julia.jl
fingolfin: A GAP matrix has no missing entries

src/gap_to_julia.jl
ThomasBreuer: The term GAP matrix is perhaps misleading.
The conversion in question shall deal with GAP objects such that the result is a Julia matrix (2-dim. array).
The candidates on the GAP side are `IsMatrixObj` objects, which are in general not in `IsList` and which do in general not support nested `ElmList` access, and `IsList` objects whose entries are again `IsList` objects, which are in general not `IsMatrixObj` objects.
The former ones cannot have holes, the latter ones can.

Depending on the prescribed Julia type, it may work to convert a nested GAP list with holes to a Julia matrix that contains `nothing` in some places; one of the tests looks as follows.
```
n = GAP.EvalString("[[1,2],[,4]]")
@test GAP.gap_to_julia(Array{Union{Int64,Nothing},2},n) == [1 2; nothing 4]
```

src/gap_to_julia.jl
fingolfin: I am not sure I understand what you are getting at here. `CSTR_STRING` is a helper function defined in GAP.jl which takes a GAP `T_STRING`, gets a `Ptr{UInt8}`, and converts it to a Julia string via `deepcopy(unsafe_string(ptr))`. And `UNSAFE_CSTR_STRING` is a badly-named sibling to that function which instead uses `unsafe_wrap(Array{UInt8,1}, char_ptr, string_len)`

That said,  the only place calling `UNSAFE_CSTR_STRING` immediately calls `deepcopy` on the result, too; I'd suggest that we move that `deepcoopy` into `UNSAFE_CSTR_STRING` and rename it to e.g. `CSTR_STRING_AS_ARRAY` or so

src/gap_to_julia.jl
```diff
@@ -78,164 +94,153 @@ function gap_to_julia(::Type{Rational{T}}, x::Int64) where T <: Integer
...
-        array = UNSAFE_CSTR_STRING( obj )
-        return deepcopy(array)
-    elseif Globals.IsList( obj )
-        return Array{UInt8,1}(map(i->gap_to_julia(UInt8,obj[i]),1:length(obj)))
-    else
-        throw(ArgumentError("<obj> is not a list"))
-    end
+## Convert GAP string to Vector{UInt8} (==Array{UInt8,1})
+function gap_to_julia(::Type{Vector{UInt8}}, obj::GapObj)
+    Globals.IsStringRep(obj) && return deepcopy(UNSAFE_CSTR_STRING(obj))
```
fingolfin, line 136: as I said above: I'd move `deepcopy` into `UNSAFE_CSTR_STRING` and then rename `UNSAFE_CSTR_STRING`

test/constructors.jl
```diff
@@ -0,0 +1,195 @@
...
+    x = GAP.EvalString( "2^100" )
+    @test_throws InexactError Int64(x)
+    @test Int128(x) == BigInt(2)^100
+    @test BigInt(x) == BigInt(2)^100
+    x = GAP.EvalString( "2^62" )  # not an immediate integer
+    @test Int64(x) == 2^62
+
+    ## BigInts
+    x = GAP.EvalString("2^100")
+    @test BigInt(x) == BigInt(2)^100
```
fingolfin, line 24: Does `big(x)` also work? If not, we should probably try to get it to work

fieker:
On Thu, Apr 30, 2020 at 08:41:46AM -0700, Max Horn wrote:
> @fingolfin commented on this pull request.
> 
> 
> 
> > +    @test GAP.Obj(true) == true
> +    x = GAP.EvalString("Z(3)")
> +    @test GAP.Obj(x) == x
> +    @test GAP.Obj(0) == 0
> +
> +    ## Border cases
> +    x = GAP.EvalString( "2^100" )
> +    @test_throws InexactError Int64(x)
> +    @test Int128(x) == BigInt(2)^100
> +    @test BigInt(x) == BigInt(2)^100
> +    x = GAP.EvalString( "2^62" )  # not an immediate integer
> +    @test Int64(x) == 2^62
> +
> +    ## BigInts
> +    x = GAP.EvalString("2^100")
> +    @test BigInt(x) == BigInt(2)^100
> 
> Does `big(x)` also work? If not, we should probably try to get it to work
On the Oscar-side of things, big is mostly irrelevant as we're not using
the BigInt, Rational{BigInt} and BigFloat in general...

src/gap_to_julia.jl
kalmarek: first of all, the comment was probably misplaced ;) `UNSAFE_CSTR_STRING` is called a few lines lower;

but since we're at this here, according to docs there is no need for `deepcopy` after `unsafe_string` as it copies the data; it's `unsafe_wrap` that does not copy; 

Since you're discussing renaming: is there a reason of calling it `CSTR_STRING` instead of `unsafe_string`? (the same for `UNSAFE_CSTR_STRING`/`unsafe_wrap`)? (I assume `T_STRING` is a type? I couldn't find it through Github search).

fingolfin: yes, `T_STRING` is the internal name inside the GAP kernel for the string type.

Thanks for the pointer on deepcopy. The names of the functions most likely have no deeper meaning; they simply are based on the name of the GAP kernel function `CSTR_STRING`. I don't mind renaming them at all.

Issue #332, opened 2020-01-28
Bug in integer promotion/conversion/multiplication


fingolfin:
Problem is caused by `Int64` values `>= 2^60`, which cannot be stored in a small int.

```
julia>  2^60 * GAP.Globals.Factorial(50)
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `*' on 2 arguments
ERROR: Error thrown by GAP
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] error_handler() at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/GAP.jl:50
 [3] #call_gap_func#5(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(GAP.call_gap_func), ::Main.ForeignGAP.MPtr, ::Int64, ::Vararg{Any,N} where N) at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/ccalls.jl:149
 [4] call_gap_func(::Main.ForeignGAP.MPtr, ::Int64, ::Vararg{Any,N} where N) at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/ccalls.jl:139
 [5] #call#7(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Main.ForeignGAP.MPtr, ::Int64, ::Vararg{Any,N} where N) at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/GAP.jl:136
 [6] (::Main.ForeignGAP.MPtr)(::Int64, ::Vararg{Any,N} where N) at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/GAP.jl:136
 [7] *(::Int64, ::Main.ForeignGAP.MPtr) at /Users/mhorn/Projekte/OSCAR/GAP.jl/src/adapter.jl:69
 [8] top-level
```

fingolfin:
Really annoying is that right now we can't even avoid this by using `julia_to_gap`:
```
julia> GAP.julia_to_gap(2^60) * GAP.Globals.Factorial(50)
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `*' on 2 arguments
ERROR: Error thrown by GAP
```

Only thing that works is to first convert to a BigInteger or Int128:
```
[...]

PR #434, opened 2020-05-05
proposed fix for a conversion inconsistency

ThomasBreuer:
Currently the following happens.
```
gap> x:= JuliaEvalString( "2^60" );
1152921504606846976
gap> GAPToJulia( 2^60 );
<Julia: 1152921504606846976>
```
The number `x` is not an immediate integer, and `Julia.Base.sqrt( x )` yields an error.
The result of `GAPToJulia` (computed using `gap_to_julia`) looks correct.
`JuliaEvalString` calls `gap_julia`, which calls `ObjInt_Int8` in this case;
I propose to use `NewJuliaObj` instead.

After this change, the tests exhibit a problem with the method
```
julia_to_gap(x::Int64)  = x
```
which is simply wrong.

Perhaps this proposal is not a good idea.
Converting `Int64` to GAP becomes more expensive because the argument has to be checked.
For the moment, I have no good idea to improve this.

### DIFF
--- a/pkg/JuliaInterface/src/convert.c
+++ b/pkg/JuliaInterface/src/convert.c
@@ -54,7 +54,7 @@ Obj gap_julia(jl_value_t * julia_obj)
         if (INT_INTOBJ_MIN <= v && v <= INT_INTOBJ_MAX) {
             return INTOBJ_INT(v);
         }
-        return ObjInt_Int8(v);
+        return NewJuliaObj(julia_obj);
     }
     if (is_gapobj(julia_obj)) {
         return (Obj)julia_obj;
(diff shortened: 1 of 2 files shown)


fingolfin:
This reverts parts of PR #357. As such, I *think* it will just replace one set of problems with another; there is simply no good solution here.

ThomasBreuer:
@fingolfin Thanks.
I think `JuliaEvalString` must return objects that are valid arguments for Julia functions,
in the same way as `GAP.EvalString` must return valid GAP objects;
`GAP.EvalString( "2^60" )` yields a `GapObj`.

In the case of `"2^60"`, Julia itself creates an object of type `Int64`.
(`GAPToJulia( 2^60 )` yields a `BigInt`, but this is another question.)

I see that the current proposal causes efficiency problems, but which logical problems had been addressed by #357? 

(By the way, the test failure is due to an explicit test for GAP's opinion about the type of 2^60.)

fingolfin:
@ThomasBreuer see issue #332

PR #441, opened 2020-05-06
Work related to BuildPackage.sh

fingolfin:


fingolfin:
So BuildPackage.sh actually *fails* but the Travis build is not aborted. That's really annoying... Hrmpf

PR #442, opened 2020-05-06
Instead of running `Pkg.build("GAP")` tell the user to do so

### LINKED ISSUE #435: Don't call Pkg.build("GAP")
We call `Pkg.build("GAP")` inside `GAP.jl` if we can't find GAP. But if the user did `Pkg.add("Oscar")` and got GAP this way, but never did `Pkg.add("GAP")`, then `Pkg.build("GAP")` actually runs into an error.

It seems there is just no good way for a package to "build itself"... Well, I guess we could run `build.jl` ourselves (or a close approximation), which should work, but opens its own can of worms.

fingolfin:
Unfortunately `Pkg.build("GAP")` will not always work (see comment
inside the patch)

Resolves #435

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -11,13 +11,14 @@ module GAP
 const julia_version = "$(VERSION.major).$(VERSION.minor)"
 deps_jl = abspath(joinpath(@__DIR__, "..", "deps", "deps-$(julia_version).jl"))
 if !isfile(deps_jl)
-    # HACK: we need to compile GAP once for each Julia version, but Julia only
+    # We need to compile GAP once for each Julia version, but Julia only
     # builds it for us once; so we need to check if the package was actually
-    # built, and if not, trigger a build now. This *seems* to work well in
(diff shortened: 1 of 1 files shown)


src/GAP.jl
```diff
@@ -11,13 +11,14 @@ module GAP
...
-    # practice, but I am not sure if we are strictly speaking "allowed" to do
-    # this
-    import Pkg
-    Pkg.build("GAP")
+    # built, and if not, tell the user to trigger a build. We used to just
+    # trigger on by ourselves using `Pkg.build("GAP")` but that doesn't work
+    # in general; e.g. if the user install Oscar.jl, and never did `add GAP`,
+    # it will just run into an error which then is even more confusing to the
+    # enduser than the message we are printing here.
+    error("""Please run `using Pkg; Pkg.add("GAP"); Pkg.build("Polymake");`""")
```
mohamed-barakat, line 21: why `Pkg.build("Polymake")`?

fingolfin: Copy&paste :-( fixed now, thanks!

PR #445, opened 2020-05-07
updated .gitignore

mohamed-barakat:


.gitignore
```diff
@@ -3,7 +3,7 @@
 /Manifest.toml
 /deps/build
 /deps/build.log
-/deps/deps-*.jl
+/deps/deps*.jl
```
fingolfin, line 6: Hmm, any `deps/deps.jl` file you have should be a left-over from an older version, and would best be removed. This is why I didn't make the ignore rule so broad. Do you have a specific reason you need it like this?

ThomasBreuer: @fingolfin I have a `deps/deps.jl` from May 5th, which was last changed a minute before `deps-1.4.jl` and `build.log` were changed for the last time.

mohamed-barakat:
Even if I delete `deps/deps.jl` it keeps popping up, also with the current master 2c84ce7.

PR #448, opened 2020-05-10
Ensure GNU readline resp. ncurses finds terminfo

### LINKED ISSUE #415: REPL editing broken in $(julia -e 'import GAP; print(GAP.gap_exe())')

fingolfin:
Partially addresses #415.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -193,6 +193,14 @@ function run_it()
 end
 
 function __init__()
+
+    # The following is needed to get readline input to work right; see also
+    # https://github.com/JuliaPackaging/Yggdrasil/issues/455 and
+    # https://github.com/oscar-system/GAP.jl/issues/415
+    if !haskey(ENV, "TERMINFO_DIRS")
+        ENV["TERMINFO_DIRS"] = joinpath(@__DIR__, "..", "deps", "usr", "share", "terminfo")
+    end
+
     ## Older versions of GAP need a pointer to the GAP.jl module during
     ## initialization, but at this point Main.GAP is not yet bound. So instead
     ## we assign this module to the name __JULIAGAPMODULE.


mohamed-barakat:
Still broken for me:

On `pr-448` I did
```
(@v1.4) pkg> build GAP
```
then invoked
`$(julia -e 'import GAP; print(GAP.gap_exe())')`
from the shell. Command-line editing is still broken, especially the backspaces do not work.

ThomasBreuer:
I can confirm what @mohamed-barakat describes:
When the GAP session is started with `$(julia -e 'import GAP; print(GAP.gap_exe())')` then command line editing does not work.

But in this situation, `GAPInfo.KernelInfo.ENVIRONMENT.TERMINFO_DIRS` is not bound,
whereas this variable is available in the GAP session started from Julia using `GAP.Globals.SESSION()`.

fingolfin:
This patch was indeed only for the GAP-via-Julia code path.

Adding it for GAP is a bit tricky; that is, setting `TERMINFO_DIRS` is easy (e.g. we can modify `gap.sh` to set it before launching `gap`); the problem is to find the right value: of course we know the right value when we build GAP.jl. But it can change later one, namely if `Ncurses_jll` gets updated... Then GAP.jl will happily use the new version, but in the new version, the paths all change (as the artifact has a new SHA256 checksum)... 

So gap.sh would have to query Julia for the right path, and be careful to do this in a way that avoids much overhead, which means that `using GAP` is not the best idea (as we'd end up starting GAP twice). This is why I am working on PR #427, where we start GAP via Julia, and thus bypass the need for all this trickery.

fingolfin:
I'll merge this for now as at least it fixes the GAP prompt when invoked from Julia. Complete solution will come via other PRs.

PR #449, opened 2020-05-11
Use get! for IdDict (available since Julia 1.1)

fingolfin:


mohamed-barakat:
I see that you deleted lines, but from the title of the PR I guessed you must have added ones.

fingolfin:
The new lines Were there already, just commented out

PR #453, opened 2020-05-12
Add GAP.prompt() method

fingolfin:
Note that I deliberately do not export the function; `prompt` could clash with other packages, but I think `GAP.prompt()` is very expressive and still short.

Please test this. Some things I tried (at one point, at least; but so far only on my mac, so testing on Linux would be really important):
- line editing should work
- break loops in GAP should work
- ctrl-c in GAP should work, e.g. to interrupt `while true do od;`
- after leaving the GAP prompt (via ctrl-D or `quit;`), everything (???) in Julia should be back to normal; e.g. entering `while true end` and then pressing `ctrl-c` repeatedly should abort the Julia loop and *not* crash GAP.

Of course I may have forgotten things that ought to be saved and restored. Please look out for anything odd.

@rbehrends please have a look at the `signal` hackery I perform there.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -81,7 +81,13 @@ function reset_GAP_ERROR_OUTPUT()
     Globals.MakeReadOnlyGlobal(julia_to_gap("ERROR_OUTPUT"))
 end
 
+disable_error_handler = false
+
 function error_handler()
+    global disable_error_handler
+    if disable_error_handler
+        return
+    end
     str = gap_to_julia(Globals._JULIAINTERFACE_ERROR_OUTPUT)
     reset_GAP_ERROR_OUTPUT()
     error("Error thrown by GAP: ", str)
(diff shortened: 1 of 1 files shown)


mohamed-barakat:
> * Line editing in GAP works,
> * break loops in GAP work (`return` in break loop, `quit` from break loop, nested break loops),
> * in GAP works,
> * and Julia seems to be o.k. after leaving the embedded GAP session.
> 
> I like the feature to have several GAP sessions embedded subsequently into one Julia session:
> Compute something in a GAP session, return to Julia, use the new GAP objects in Julia, and later repeat this; throughout you have access to all Julia variables from GAP and to all GAP variables from Julia.
> (Strictly speaking, typing `quit;` at the GAP prompt in the embedded session does not really quit the GAP session, because the next `GAP.prompt()` in Julia takes you back to the same GAP session.)

@ThomasBreuer This sounds interesting. Is it possible to prepare a Jupyter-Notebook which demonstrates these abilities?

ThomasBreuer:
@mohamed-barakat I do not think that there is an easy way to demonstrate this feature in a Jupyter notebook. The input `GAP.prompt()` in a notebook runnning a Julia kernel just yields a GAP prompt in an output cell, one does not get access to the GAP session --the next input cell expects Julia input.

fingolfin:
It is exactly as @ThomasBreuer says -- of course it would be interesting to have "GAP Jupyter cells" by piggybacking on `IJulia` (thus sidestepping all issues with GAP's `JupyterKernel` package which seems to be essentially unmaintained). But this will require dedicated efforts on our part.

mohamed-barakat:
> @mohamed-barakat I do not think that there is an easy way to demonstrate this feature in a Jupyter notebook. The input `GAP.prompt()` in a notebook runnning a Julia kernel just yields a GAP prompt in an output cell, one does not get access to the GAP session --the next input cell expects Julia input.

Is it then possible to paste your experiments here, so I can see what you did and test it on my MAC?

mohamed-barakat:
Thank you very much @fingolfin. What I tested worked fine:

```julia
(@v1.4) pkg> build GAP

julia> using GAP

julia> GAP.prompt()
gap> a
> := 1;
1
gap> a
> :=1;
1
gap> Error()p
Error,  called from
not in any function at *stdin*:5
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk>
Syntax error: ; expected
Error()p
 ^^^^^^^
gap> Error();
Error,  called from
not in any function at *stdin*:5
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> a := 1 + 1;
2
brk>
gap> f := function() Error(); end;
function(  ) ... end
gap> f();
Error,  at *stdin*:5 called from
<function "f">( <arguments> )
 called from read-eval loop at *stdin*:6
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> a := l;
Syntax warning: Unbound global variable in *errin*:1
a := l;
     ^
Error, Variable: 'l' must have a value in
  Error(  ); at *stdin*:5 called from
<function "f">( <arguments> )
 called from read-eval loop at *errin*:1
brk> a := 1;
1
brk>
gap> while true do od;
^CError, user interrupt in
  ; at *stdin*:6 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:6
you can 'return;'
brk>
gap> while true do od;
^CError, user interrupt in
  ; at *stdin*:6 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:6
you can 'return;'
brk>
gap>

julia> GAP.prompt()
gap> while true do od;
^CError, user interrupt in
  ; at *stdin*:1 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:1
you can 'return;'
brk>
gap>

julia>
```

PR #463, opened 2020-05-18
Switch from Travis to GitHub Actions (WIP)

fingolfin:
GitHub Actions are quite a bit faster these days, have an IMHO better web interface, and have fewer random hiccups (as far as I can tell).

TODO:
- restore coverage computation, to cover same number of files as before
- add Documenter.jl upload job


fingolfin:
TODO: add Documenter.jl upload job

PR #468, opened 2020-05-20
test whether a test works as expected

ThomasBreuer:


fingolfin:
```
########> Diff in /home/travis/build/oscar-system/GAP.jl/pkg/JuliaInterface/ts\
t/juliainterface01.tst:92
# Input is:
gapmodule:= JuliaModule( "GAP" );
# Expected output:
<Julia: GAP>
# But found:
<Julia: Main.GAP>
########
```

ThomasBreuer:
@fingolfin Yes, this is exactly what I do not understand: Why does Travis get `<Julia: Main.GAP>`?
With the new setup for starting GAP from Julia, `<Julia: GAP>` should be returned.

PR #472, opened 2020-06-10
extend the documentation of GAP.jl

ThomasBreuer:
- Added docstrings to several functions of GAP.jl,
  in particular, filled the "TODO" placeholders.

- related code changes:

  - Added `hasproperty` methods for `GapObj` that delegate to `ISB_REC`
    (which had been left out because earlier Julia versions did not support
    `hasproperty`).

  - Changed the `in` method that delegates to GAP's `\in` such that
    `Any` is allowed for the first argument (which may be also a Julia object).

  - Changed the `one` method that delegates to a GAP operation such that
    `ONE_MUT` is called instead of `ONE`; this way, the mutability of the
    result is the same as that of the argument (like in the cases of `ZERO`
    and `AINV`).

  - Added an `inv` method for `GapObj` that delegates to GAP's `INV_MUT`
    (which yields a result of the same mutability as the argument).
    [Yes, the names of the GAP operations are not consistent.]

  - Support an optional version number in `GAP.Packages.load`.

  - Avoid an error in the `getdoc` method if the argument is a `GapObj`
    without `NameFunction` value.

  - The GAP function `UrlOfManualEntry` computes an URL of a link
    in the "official" GAP manuals in the web
    from the bookname and label information.
    This is preliminary in the sense that the relevant addresses
    might be changed in the future,
    in order to improve the setup on the GAP side;
    for example, currently there is essentially only one supported version
    of a GAP (package) manual.
    The fact that this works only for GAPDoc based manuals should not be a
    restriction for Julia related stuff.

  - `GAP.compute_links_to_GAP_manuals` is called in `docs/make.jl`,
    in order to replace "external references" to GAP (package) manuals
    in the HTML file produced by Documenter.jl.
    (For examples, see the entries about `GAP.Packages.load` and
    `GAP.Packages.update`.)
    This approach is not optimal, Documenter.jl prints warnings because it
    expects explicit addresses, but this way we get specific links in the
    HTML version of the manual,
    and the version in the interactive Julia session is at least readable.
    *One open question* is:
    Where is the most natural place for the JuliaInterface manual?
    Since it is a GAP package, the first place to look for it is
    [the overview of GAP packages on the GAP website](https://www.gap-system.org/Packages/packages.html),
    with the HTML version of the manual  under https://www.gap-system.org/Manuals/pkg/.
    On the other hand, JuliaInterface relies on GAP.jl, perhaps the documentation should be stored
    close to that of GAP.jl.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -63,7 +63,7 @@ using Julia syntax features.
 ```@docs
 Globals
 call_gap_func
-EvalString
+evalstr
 getindex
 setindex!
 getproperty
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -9,10 +9,30 @@ CurrentModule = GAP
 
 ## Introduction
 
-TODO: describe the goals and non-goals of this package
+GAP.jl is a low level interface from Julia to
+[the computer algebra system GAP](https://www.gap-system.org).
+The term "low level" means that the aim is
+to give Julia access to all GAP objects,
+to let Julia call GAP functions,
+and to provide conversions of low level data
+(integers, Booleans, strings, arrays/lists, dictionaries/records)
+between the two systems.
 
-TODO: also link to JuliaInterface docs (also JuliaExperimental?)
+In particular, it is *not* the aim of GAP.jl to provide Julia types
+for higher level GAP objects that represent algebraic structures,
+such as groups, rings, fields, etc.,
+and mappings between such structures.
 
+The connection between GAP and Julia is in fact bidirectional, that is,
+GAP can access all Julia objects,
+call Julia functions,
+and perform conversions of low level data.
+This direction will become interesting on the Julia side
+as soon as GAP packages provide functionality that is based on
+using Julia code from the GAP side.
+
+The viewpoint of an interface from GAP to Julia is described in
+[the manual of the GAP package JuliaInterface](GAP_ref(JuliaInterface:Title page)).
 
 ## Types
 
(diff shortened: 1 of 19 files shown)


docs/src/index.md
```diff
@@ -37,17 +57,126 @@ julia_to_gap
...
 
-TODO: describe the various convenience / adapter methods we install, e.g. for
-basic arithmetic, accessing GAP list and record entries, calling GAP function, etc.
+This section describes how one can manipulate GAP objects from the Julia side,
+using Julia syntax features.
+
+```@docs
+Globals
+call_gap_func
+EvalString
```
fingolfin, line 66: See also issue #473

src/GAP.jl
```diff
@@ -24,8 +24,18 @@ include(deps_jl)
 """
     FFE
 
-Wraps a pointer to a GAP FFE immediate object.
+Wrap a pointer to a GAP FFE immediate object.
```
fingolfin, line 27: Perhaps we should also explain what "FFE" stands for (even if this is only internal documentation)
```suggestion
Wrap a pointer to a GAP FFE (= "finite field element") immediate object.
```

src/GAP.jl
```diff
@@ -39,7 +49,48 @@ import Markdown
...
+true
+
+julia> isa( GAP.EvalString( "2^59" ), GAP.GapObj )     # a small GAP integer
+false
+
+julia> isa( GAP.EvalString( "Z(2)" ), GAP.GapObj )     # a GAP FFE
+false
+
+julia> isa( GAP.EvalString( "true" ), GAP.GapObj )     # a Boolean
+false
```
fingolfin, line 76: Instead of showing what this is *not* we could also just show the respective types? I.e.,
```suggestion
julia> typeof( GAP.EvalString( "true" ) )     # a Boolean
Bool
```
and similar for the other cases

src/ccalls.jl
```diff
@@ -15,6 +15,35 @@ function EvalStringEx(cmd::String)
...
+# Examples
+```jldoctest
+julia> GAP.EvalString( "1+2" )
+3
+
+julia> GAP.EvalString( "x:= []; y:= 0" )
+0
+
+julia> GAP.EvalString( "Add( x, 1 )" )
+ERROR: Error thrown by GAP: Error, List Element: <list>[2] must have an assigned value
```
fingolfin, line 33: See also #474 (this is not a request to change this PR, I just wanted to point out the issues it motivated me to submit ;-)

src/ccalls.jl
```diff
@@ -15,6 +15,35 @@ function EvalStringEx(cmd::String)
...
+[...]
+
+julia> GAP.EvalString( "x:= []" )
+GAP: [  ]
+
+julia> GAP.EvalString( "Add( x, 1 ); 0" )
+0
```
fingolfin, line 40: This one is a bit weird and against the original intention of `EvalString`. But I am fine with accepting it, but I count this as yet another motivation to resolve issue #474 (one shouldn't have to add a trailing `0` like this to execute an `Add`)

src/help.jl
```diff
@@ -4,10 +4,41 @@ function GAP_help_string(topic::String, onlyexact::Bool = false)
     return GAP.gap_to_julia(GAP.Globals.HelpString(GAP.julia_to_gap(topic), onlyexact))
 end
 
+"""
+    show_GAP_help(topic::String, onlyexact::Bool = false)
```
fingolfin, line 8: Hmm, elsewhere we use `gap` in lowercase in function names, and that also seems to be more the "Julia way", so perhaps this function should also do that?
```suggestion
    show_gap_help(topic::String, onlyexact::Bool = false)
```

src/packages.jl
```diff
@@ -81,23 +81,35 @@ module Packages
...
-TODO: copy text from <https://gap-packages.github.io/PackageManager/doc/chap2.html> resp.
-link to that
+Download and install the newest released version of the GAP package
+given by `desc` in the `pkg` subdirectory of GAP's build directory
+(variable `GAP.GAPROOT`).
+Return `true` if the installation is successful or if the package
+was already installed, and `false` otherwise.
+
+`desc` can be either the package name or an URL of an archive or repository
+of the package, or an URL of its `PackageInfo.g` file.
```
fingolfin, line 109:
```suggestion
`desc` can be either the name of a package or the URL of an archive or repository
containing a package, or the URL of a `PackageInfo.g` file.
```
We might also want to rename `desc` to `spec`, to align our terminology with that of `Pkg.jl`

ThomasBreuer:
The test failure is due to a failed build for Singular.jl in Julia 1.3 under Linux; in the other situations, Singular.jl builds and the tests pass.

Issue #476, opened 2020-06-18
Suppress output of `GAP.Packages.install` if package is already installed?


fingolfin:
If one needs a package like "ferret" inside Julia code, one may want to call `GAP.Packages.install("ferret")` to ensure it is available; the operation is essentially idempotent, so this is fine. However, it outputs text which is visible to the user, which is not nice:
```
julia> GAP.Packages.install("ferret")
#I  Getting PackageInfo URLs...
#I  Retrieving PackageInfo.g from https://gap-packages.github.io/ferret/PackageInfo.g ...
#I  The newest version of package "ferret" is already installed
true
````

ThomasBreuer:
The info messages in question are created by PackageManager if the info level of `InfoPackageManager` is at least 3, which is the default value.
If the point is to avoid the messages then we could set the level to 2 in the Julia code, or even to a smaller value.

However, if the point is to make sure that *some* version of the package is available, calling PackageManager's `InstallPackage` is not what one wants, since it will download the newest `PackageInfo.g` in order to make sure that the *newest* version of the package is installed.
Perhaps a function `GAP.Packages.load_and_install_if_necessary`  would be a better solution, which tries to install the package only if it cannot simply be loaded, and signals an error if the package cannot be installed.

ThomasBreuer:
@fingolfin Can this issue be closed?

PR #480, opened 2020-06-24
admit installing a package in `Packages.load`

ThomasBreuer:
A new keyword argument `install` can be used to force
`Packages.load` to install the package in question
if it is not yet installed.

This is intended to resolve issue #476.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -81,17 +81,40 @@ module Packages
 import ...GAP: Globals, julia_to_gap, GAPROOT
 
 """
-    load(spec::String, version::String = "")
+    load(spec::String, version::String = ""; install = false)
 
 Try to load the newest installed version of the GAP package with name `spec`.
 Return `true` if this is successful, and `false` otherwise.
 
 The function calls [GAP's `LoadPackage` function](GAP_ref(ref:LoadPackage));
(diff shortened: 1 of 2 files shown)


src/packages.jl
```diff
@@ -81,17 +81,40 @@ module Packages
...
+    elseif install == true
+        # Check whether an admissible version of the package is installed;
+        # if yes then we do not try to install the package anew.
+        info = Globals.PackageInfo(gspec)
+        available = [i -> info[i].Version for i in 1:length(info)]
+        if all(v -> ! Globals.CompareVersionNumbers(v, gversion), available)
+            # Try to install the package.
+            if Packages.install(spec)
+                # Make sure that the version is admissible.
+                return Globals.LoadPackage(gspec, gversion, false)
```
fingolfin, line 112: While this is useful, I don't think it fits quite for the usecase of @mohamed-barakat (or at least not all of them): e.g. `io` *is* shipped with GAP.jl, as are all other packages -- it just isn't compiled and hence not usable...

That said, if we also reactivate PR #365 (I'll look into it), then I think this PR will fully solve this.

So I think this PR is in the right direction, and already in some settings useful, so I will merge this now, and then look into PR #355.

PR #481, opened 2020-06-30
added documentation and doctests for constructors ...

ThomasBreuer:
... and had to adjust the code, as follows:

- Fixed the `Char(obj::GapObj)` method,
- removed the `Vector{UInt8}(obj::GapObj)` method, it was covered already by the `Vector{T}(obj::GapObj)` method,
- admit `gap_to_julia(Vector{T}, obj)` also if `obj` is in `Globals.IsVectorObj`, added tests,
- fixed `gap_to_julia(StepRange, obj)`; note that `StepRange(obj)` is not supported (but `StepRange{Int,Int}(obj)` is), now we call `convert(T, obj)` instead of `T(obj)` (thanks to @rfourquet for a discussion about this),
- support `gap_to_julia(obj)` for `obj` in `Globals.IsMatrixObj`,
- added tests for `Matrix{T}(obj)` where `obj` is in Globals.IsMatrixObj`.

Perhaps we want to change the behaviour of `gap_to_julia(obj)` for `obj` in `Globals.IsMatrixObj`:
Up to now, there was no such method (without prescribed Julia type).
Currently `Globals.IsList` is checked before `Globals.IsMatrixObj`,
which means that a conversion of a GAP matrix to a nested Julia array
is preferred.
If we would check for `Globals.IsMatrixObj` first then the default
conversion of GAP matrices would produce 2-dimensional Julia arrays.


src/gap_to_julia.jl
```diff
@@ -283,14 +292,22 @@ function gap_to_julia(::Type{T}, obj::GapObj) where {T<:StepRange}
...
-        result = obj[1]:(obj[2]-obj[1]):obj[len]
+        start = obj[1]
+        step = obj[2]-obj[1]
+        stop = obj[len]
     end
 
-    return T(result)
+    # Julia does not support `StepRange(obj)` (but `StepRange{S,T}(obj)`),
+    # but we can always use `convert`.
+    return convert(T, start:step:stop)
```
fingolfin, line 310: I am not sure I understand. Can't we just do this?
```suggestion
    return T(start,step,stop)
```

ThomasBreuer: Perhaps the following is a pathological case.
```julia
julia> convert( StepRange{Int}, 1:2:5 )
1:2:5

julia> StepRange{Int}(1,2,5)
ERROR: MethodError: no method matching StepRange{Int64,S} where S(::Int64, ::Int64, ::Int64)
[...]
```

fingolfin: In a clean Julia session (no packages loaded), I get this:
```
julia> StepRange{Int}(16)
ERROR: MethodError: no method matching StepRange{Int64,S} where S(::Int64)
Stacktrace:
 [1] top-level scope at REPL[1]:1

julia> StepRange{Int}(16,16)
ERROR: MethodError: no method matching StepRange{Int64,S} where S(::Int64, ::Int64)
Stacktrace:
 [1] top-level scope at REPL[2]:1

julia> StepRange{Int}(16,16,16)
ERROR: MethodError: no method matching StepRange{Int64,S} where S(::Int64, ::Int64, ::Int64)
Stacktrace:
 [1] top-level scope at REPL[3]:1

julia> convert(StepRange{Int}, 1:2:3)
1:2:3
```
Based on that, I think it's fair if we also throw an error when `StepRange` is only partially parametrized.

But I guess addition, we need ponder whether we also ought to provide `convert` methods... This is not so nice:
```
julia> str = g"test"
GAP: "test"

julia> String(str)
"test"

julia> convert(String, str)
ERROR: MethodError: Cannot `convert` an object of type Main.ForeignGAP.MPtr to an object of type String
Closest candidates are:
  convert(::Type{T}, ::T) where T<:AbstractString at strings/basic.jl:209
  convert(::Type{T}, ::AbstractString) where T<:AbstractString at strings/basic.jl:210
  convert(::Type{T}, ::T) where T at essentials.jl:171
  ...
Stacktrace:
 [1] top-level scope at REPL[16]:1
```
Of course that's getting beyond the scope of this PR; perhaps we ought to log an issue for it as a reminder to implement it at some point

ThomasBreuer: Concerning the `StepRange` example, I think that `StepRange{Int}` is a valid Julia type, and it is not bad if the conversion of a GAP range via a constructor method for this type works, even if the corresponding conversion of a Julia range currently runs into an error (which may change in the future, who wants to rely on that).

Concerning the addition of `convert` methods, I have created issue #482 .

PR #483, opened 2020-07-10
Optimize use of locks in calling out to GAP.

rbehrends:
This checks if Julia is actually running in multi-threaded mode and, if not, elides the use of locks.

Note that this also gets rid of the `THREADSAFE_GAP_JL` macro and turns them on all the time. Thus, there is actually a small overhead for the sync macros now that are equivalent to a function call. However, that overhead should be a fairly small part of the overall overhead for calling GAP from Julia. It can probably be tweaked further by making them inline if needed.

### DIFF
--- a/pkg/JuliaInterface/src/sync.c
+++ b/pkg/JuliaInterface/src/sync.c
@@ -4,9 +4,12 @@
 
 #ifndef HPCGAP
 static pthread_mutex_t GapLock;
+static int             is_threaded;
 
 void InitGapSync(void)
 {
+    extern int jl_n_threads;
+    is_threaded = jl_n_threads > 1;
     pthread_mutexattr_t attr;
     pthread_mutexattr_init(&attr);
     pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
@@ -16,11 +19,13 @@ void InitGapSync(void)
 
 void BeginGapSync(void)
 {
-    pthread_mutex_lock(&GapLock);
+    if (is_threaded)
+        pthread_mutex_lock(&GapLock);
 }
 
 void EndGapSync(void)
 {
-    pthread_mutex_unlock(&GapLock);
+    if (is_threaded)
+        pthread_mutex_unlock(&GapLock);
 }
 #endif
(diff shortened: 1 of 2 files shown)


fingolfin:
I see you reverted the bit about THREADSAFE_GAP_JL... Still, I'd like to suggest that instead of just assuming that this "should" not affect performance, let's actually verify that using some microbenchmarks? For other reasons's I've done that in issues #485 and #486. See also PR #487 which also will have to be adapted to call `Begin/EndGapSync`

fingolfin:
@rbehrends what's the status of this PR?

rbehrends:
It can be merged as is, but the functionality is still off by default.

There are two things that I've been looking at that have stopped me from enabling it by default.

1\. The primary one is that we don't handle GAP errors yet. This is arguably an issue beyond the scope of this PR, as discussed [here](https://github.com/gap-system/gap/pull/3256#issuecomment-632763171). We will probably need some functionality in line with the existing `RegisterSyLongjmpObserver()`, which currently doesn't quite do the job (we basically need to hook into both the `setjmp()`, the `longjmp()`, and the normal return parts. This could also be useful for other situations in GAP code where we need to unwind the stack without incurring unnecessary overhead, as using `GAP_TRY` would if we placed it everywhere.

Note that this is not a problem if running Julia in single-threaded mode and only arises if `JULIA_NUM_THREADS > 1`, as no action is taken in that case.

2\. The second concern is performance in the multi-threaded case. There is currently no measurable performance overhead for single-threaded mode; this is not surprising, as we basically do nothing in this case. However, there is a measurable overhead for multi-threaded mode (`JULIA_NUM_THREADS > 1`). This can be improved a bit, but part of the options might require dropping down to assembly to improve the locking mechanism. This would potentially introduce considerable complexity. Obviously, any overhead still beats crashes that we would get otherwise if we actually tried to run GAP code in multiple threads.

An alternative solution would be to have a `gap_sync` method or something similar. Most of the overhead is only incurred upon the initial lock. So, by declaring a section as synchronized via `gap_sync() do ... end` or something similar, we would incur that overhead only once per such section and the `BeginGapSync()` and `EndGapSync()` functions would just have to test whether the global lock is actually held and only block if another thread is holding it.

PR #487, opened 2020-07-12
Make calls from Julia to GAP faster

### LINKED ISSUE #486: Reduce overhead for calls from Julia to GAP
The overhead for simple calls to GAP functions is rather (too) high at the moment. Compare the following timings; first on the Julia side (see also issue #485):
```julia
using GAP
G = GAP.Globals.CyclicGroup(10)

function test_isabelian_1()
  x = true
  for i = 1:1000000
    x = x && GAP.Globals.IsAbelian(G)
  end
end

function test_isabelian_2()
  f = GAP.Globals.IsAbelian
  x = true
  for i = 1:1000000
    x = x && f(G)
  end
end
```
Here are the timings:
```
julia> @benchmark test_isabelian_1()
BenchmarkTools.Trial:
  memory estimate:  15.26 MiB
  allocs estimate:  1000000
  --------------
  minimum time:     213.316 ms (0.00% GC)
  median time:      218.107 ms (0.00% GC)
  mean time:        220.472 ms (0.77% GC)
  maximum time:     257.289 ms (8.72% GC)
  --------------
  samples:          23
  evals/sample:     1

julia> @benchmark test_isabelian_2()
BenchmarkTools.Trial:
  memory estimate:  15.26 MiB
  allocs estimate:  1000000
  --------------
  minimum time:     158.783 ms (0.00% GC)
  median time:      164.770 ms (0.00% GC)
  mean time:        165.974 ms (0.99% GC)
  maximum time:     183.713 ms (9.34% GC)
  --------------
  samples:          31
  evals/sample:     1
```

UPDATE: the old timings were taken when Oscar was loaded and are much higher. I leave them here for reference:
```
julia> @benchmark test_isabelian_1()
BenchmarkTools.Trial:
  memory estimate:  15.26 MiB
  allocs estimate:  1000000
  --------------
  minimum time:     843.564 ms (0.00% GC)
  median time:      867.109 ms (0.00% GC)
  mean time:        867.623 ms (0.00% GC)
  maximum time:     891.991 ms (0.00% GC)
  --------------
  samples:          6
  evals/sample:     1

julia> @benchmark test_isabelian_2()
BenchmarkTools.Trial:
  memory estimate:  15.26 MiB
  allocs estimate:  1000000
  --------------
  minimum time:     565.581 ms (0.00% GC)
  median time:      568.943 ms (0.00% GC)
  mean time:        569.282 ms (0.00% GC)
  maximum time:     576.711 ms (0.00% GC)
  --------------
  samples:          9
  evals/sample:     1
```

Now compare this to the following GAP equivalent GAP code and its timings:
```gap
G := CyclicGroup(10);

test_isabelian_1 := function()
  local x, i;
  x := true;
  for i in [1.. 1000000] do
    x := x and IsAbelian(G);
  od;
end;

test_isabelian_2 := function()
  local f, x, i;
  f := IsAbelian;
  x := true;
  for i in [1.. 1000000] do
    x := x and f(G);
  od;
end;
```
And here are the benchmark results (I entered the above code into a Julia session with help of `GAP.prompt()`):
```
julia> @benchmark GAP.Globals.test_isabelian_1()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     20.128 ms (0.00% GC)
  median time:      20.618 ms (0.00% GC)
  mean time:        20.787 ms (0.00% GC)
  maximum time:     27.967 ms (0.00% GC)
  --------------
  samples:          241
  evals/sample:     1

julia> @benchmark GAP.Globals.test_isabelian_2()
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     18.202 ms (0.00% GC)
  median time:      18.921 ms (0.00% GC)
  mean time:        19.084 ms (0.00% GC)
  maximum time:     24.778 ms (0.00% GC)
  --------------
  samples:          262
  evals/sample:     1
```

That's an overhead factor of <s>30-40</s> 9-11, quite bad!  Of course in actual real world code, it might not be quite as bad as long as we don't call very short GAP functions many many times from a tight loop; but it still concerns me.

I do have an older incomplete branch which was meant to address this, but up to now, I never carefully benchmarked this, so I wasn't aware how bad it is. I'll now dig out that branch and see if I can complete it.

fingolfin:
**UPDATE: once again rebased and timings updated; also added more extensive GAP microbenchmarking results to show how close we are to matching the GAP interpreter.**

... by avoiding conversion calls for immediate integers, immediate FFE and
GapObj; and also by directly ccalling into the GAP function handlers.

Before this patch:

    julia> @btime test_simple()
      25.844 ms (0 allocations: 0 bytes)

    julia> @btime test_isabelian_1()
      195.323 ms (1000000 allocations: 15.26 MiB)

    julia> @btime test_isabelian_2()
      159.374 ms (1000000 allocations: 15.26 MiB)

After this patch (note the avoided allocations):

    julia> @btime test_simple()
      25.546 ms (0 allocations: 0 bytes)

    julia> @btime test_isabelian_1()
      53.635 ms (0 allocations: 0 bytes)

    julia> @btime test_isabelian_2()
      25.289 ms (0 allocations: 0 bytes)

For comparison, here are timings for roughly equivalent GAP functions;

    julia> @btime GAP.Globals.test_simple()
      5.134 ms (0 allocations: 0 bytes)

    julia> @btime GAP.Globals.test_isabelian_1()
      20.187 ms (0 allocations: 0 bytes)

    julia> @btime GAP.Globals.test_isabelian_2()
      18.176 ms (0 allocations: 0 bytes)

Here is the Julia microbenchmark:
```julia
    using GAP, BenchmarkTools
    function test_simple()
      for i = 1:1000000
        x = GAP.Globals.IsAbelian
      end
    end

    G = GAP.Globals.CyclicGroup(10)

    function test_isabelian_1()
      x = true
      for i = 1:1000000
        x = x && GAP.Globals.IsAbelian(G)
      end
    end

    function test_isabelian_2()
      f = GAP.Globals.IsAbelian
      x = true
      for i = 1:1000000
        x = x && f(G)
      end
    end
```
And here is the GAP version (can be entered using `GAP.prompt()`)
```gap
    test_simple := function()
      local x, i;
      for i in [1.. 1000000] do
        x := IsAbelian;
      od;
    end;

    G := CyclicGroup(10);

    test_isabelian_1 := function()
      local x, i;
      x := true;
      for i in [1.. 1000000] do
        x := x and IsAbelian(G);
      od;
    end;

    test_isabelian_2 := function()
      local f, x, i;
      f := IsAbelian;
      x := true;
      for i in [1.. 1000000] do
        x := x and f(G);
      od;
    end;
```
Resolves #486

TODO:
- [x] add tests that pass in/out various types combos
- [x] add tests that verify that calls with 0-6 args work fine
- [ ] in threaded mode, we ought to call `BeginGapSync` and `EndGapSync` (and then we also need to worry about catching errors, so that we can call `EndGapSync` regardless)

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -2,13 +2,38 @@
 
 import Base: getproperty, hasproperty, setproperty!, propertynames
 
+#
+# low-level GAP -> Julia conversion
+#
 function RAW_GAP_TO_JULIA(ptr::Ptr{Cvoid})
+    # convert immediate ints and FFEs directly, to void (un)boxing
+    as_int = reinterpret(Int, ptr)
+    if as_int & 1 == 1
+        return as_int >> 2
+    elseif as_int & 2 == 2
+        return reinterpret(FFE, ptr)
+    end
     return ccall(:julia_gap, Any, (Ptr{Cvoid},), ptr)
 end
 
+#
+# low-level Julia -> GAP conversion
+#
 function RAW_JULIA_TO_GAP(val::Any)::Ptr{Cvoid}
     return ccall(:gap_julia, Ptr{Cvoid}, (Any,), val)
 end
+#RAW_JULIA_TO_GAP(x::Bool) = x ? gap_true : gap_false
+RAW_JULIA_TO_GAP(x::FFE) = reinterpret(Ptr{Cvoid}, x)
+RAW_JULIA_TO_GAP(x::GapObj) = pointer_from_objref(x)
+function RAW_JULIA_TO_GAP(x::Int)
+    # convert x into a GAP immediate integer if it fits
+    if x in -1<<60:(1<<60-1)
+        return Ptr{Cvoid}(x << 2 | 1)
+    end
+    return ccall(:gap_julia, Ptr{Cvoid}, (Any,), x)
+    #return ccall(:ObjInt_Int, Ptr{Cvoid}, (Int,), x)
+end
+
 
 function evalstr_ex(cmd::String)
     res = ccall(:GAP_EvalString, Any, (Ptr{UInt8},), cmd)
@@ -161,6 +186,8 @@ GAP: "2xD8"
 
 ```
 """
+# this is the generic method which supports keyword arguments (mapped to GAP options)
+# and goes through JuliaInterface, which is convenient but a bit slow.
 function call_gap_func(func::GapObj, args...; kwargs...)
     global Globals
     options = false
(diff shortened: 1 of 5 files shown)


src/ccalls.jl
```diff
@@ -3,12 +3,32 @@
...
 end
 
 function RAW_JULIA_TO_GAP(val::Any)::Ptr{Cvoid}
     return ccall(:gap_julia, Ptr{Cvoid}, (Any,), val)
 end
+RAW_JULIA_TO_GAP(x::FFE) = reinterpret(Ptr{Cvoid}, x)
+RAW_JULIA_TO_GAP(x::GapObj) = pointer_from_objref(x)
+function RAW_JULIA_TO_GAP(x::Int64)
+    # if it fits into a GAP immediate integer, convert x to Int64
+    if x in -1<<60:(1<<60-1)
```
rfourquet, line 23:
```suggestion
    if x in -one(Int64)<<60:(one(Int64)<<60-1)
```

fingolfin: What's the difference?

rfourquet: `1` is of type `Int`, so `1<<60` overflows on 32-bit machines (which I assumed are supported as the `x` argument is explicitely typed as `Int64` and not `Int`).

fingolfin: Actually 32bit is not supported at all

fingolfin:
The test failures saying `Error, UndefVarError: T_FUNCTION not defined` are caused by the fact that already `GAP_Initialize` ends up processing the `tst/testall.g` file... We somehow need to insert code that is run after the kernel has fully initialized but before GAP's `init.g` processes command line arguments. I guess I could do this by inserting another `-c BLA` into the command line. But on the long run, I think the libgap API needs to be improved for this kinda of stuff.

Issue #484, opened 2020-07-10
doctest fail with Julia 1.6?


fingolfin:
See https://travis-ci.com/github/oscar-system/GAP.jl/jobs/359229945

This seems annoying, not sure we can do something clever about it

CC @ThomasBreuer

fingolfin:
Perhaps we can use https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Filtering-Doctests to help with this

ThomasBreuer:
In principle, the `doctestfilters` feature seems to solve the problem.
I would prefer the variant where all differences are collected globally, and entered as one list to the function in question (as opposed to the variant where each affected doctest gets such an "annotation").
This works for executing doctests as a part of what `makedocs` does, but it seems not to work for executing doctests as ordinary tests, via `doctest`, see https://github.com/JuliaDocs/Documenter.jl/issues/1364.

(As far as I understand, the failures do not happen with Julia 1.5. I do not see the changed `typeof`/`showarg` outputs in Julia 1.5.0-rc1, but in Julia 1.6.0-DEV.)

fingolfin:
This was fixed in #491

PR #491, opened 2020-07-15
define some doctestfilters

ThomasBreuer:
... in order to avoid failures in doctests with Julia 1.6.0-DEV
(due to changed `typeof` outputs in Julia 1.6)

Note that `doctestfilters` is supported in `makedocs` but currently not in `doctest`,
therefore running doctests as a part of GAP.jl's test suite gets disabled with the proposed changes.

Perhaps there will be support also in `doctest` in the future, see JuliaDocs/Documenter.jl#1364.

The changes are intended to resolve issue #484 .

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -8,6 +8,13 @@ makedocs(
     modules = [GAP],
     clean = true,
     doctest = true,
+    doctestfilters = Regex[
+      r"BitVector|BitArray\{1\}",
+      r"Matrix\{Int64\}|Array\{Int64,2\}",
+      r"Vector\{Any\}|Array\{Any,1\}",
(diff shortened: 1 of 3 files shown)


fingolfin:
Great! Hmm, perhaps we could use such a feature in GAP's `Test`, too... As we also have on GAP test failing:
```
########> Diff in /Users/travis/build/oscar-system/GAP.jl/pkg/JuliaInterface/t\
st/utils.tst:9
# Input is:
JuliaTypeInfo( GAPToJulia( JuliaEvalString( "Array{Any,1}" ), [ 1, 2, 3 ] ) );
# Expected output:
"Array{Any,1}"
# But found:
"Vector{Any}"
########
    1130 ms (334 ms GC) and 233KB allocated for utils.tst
```

PR #492, opened 2020-07-15
more conversion variants and documentation

### LINKED ISSUE #482: provide `convert` methods for GAP to Julia conversion, parallel to constructor methods
This has been suggested by @fingolfin in comments to issue #428 (with links to Julia developers' discussions) and pull request #481 .

ThomasBreuer:
- added `convert` methods for GAP-to-Julia and Julia-to-GAP conversions,
  added tests for these methods
- extended the documentation of conversions,
  added documentation examples for `gap_to_julia` and `julia_to_gap`
- admit `GapObj` as a constructor for Julia-to-GAP conversions
- support `julia_to_gap` for GAP objects with `recursive = true`
  (in case the GAP object has Julia subobjects),
  added a test for this situation

(intended to resolve #482 )


src/julia_to_gap.jl
```diff
@@ -175,4 +189,33 @@ function julia_to_gap(
     return record
 end
 
+## GAP objects: deal with recursion (subobjects may be Julia objects)
+function julia_to_gap(
+    obj::GapObj,
+    recursion_dict::IdDict{Any,Any} = IdDict();
+    recursive::Bool = false,
+)
```
fingolfin, line 197: By chance I was working on changes to the `julia_to_gap` and `gap_to_julia` code in the past couple days as well, and I started to really wonder about the way we set it up. Like, we keep having to repeat these default arguments. 

I think it would make more sense if we refactored both of these functions, by changing *all* methods that actually do something to take the full set of arguments (without default arguments specified); and then have one set of "generic" functions at the top level which add in the default values for arguments if missing. This way we safe we are guaranteed to get consistent defaults.

Actually, I might go a bit further:
- I'd rename the "worker methods" for `julia_to_gap` to `_ julia_to_gap`, and change them all to have *two* arguments: `obj` and `recursion_dict`, where the latter is allow to be `nothing` (so type would be `Union{Nothing, IdDict{Any,Any}}`. Then we don't need to allocate an IdDict that we don't end up using (but perhaps Julia already is clever enough to do that for us?)
- all worker methods would then only invoke `_ julia_to_gap` but never `julia_to_gap`, which is reserved for clients
- the new function `julia_to_gap` would then look like this:
```julia
julia_to_gap(obj; recursive::Bool = true) = _julia_to_gap(obj, recursive ? IdDict() : nothing)
```
- similar for `gap_to_julia`; except we already have a `_gap_to_julia`; but I'd just remove that, and instead teach the high level `gap_to_julia` (and perhaps also `julia_to_gap`) to *also* accept recursive as a positional argument:
```julia
julia_to_gap(obj, recursive) = _julia_to_gap(obj, recursive ? IdDict() : nothing)
```

Thoughts?

src/julia_to_gap.jl
```diff
@@ -175,4 +189,33 @@ function julia_to_gap(
     return record
 end
 
+## GAP objects: deal with recursion (subobjects may be Julia objects)
+function julia_to_gap(
```
fingolfin, line 193: so this is for things like converting a Julia vector of GAP lists of Julia objects? Perhaps expand the comment a little bit explain that?

src/julia_to_gap.jl
```diff
@@ -175,4 +189,33 @@ function julia_to_gap(
...
+    elseif Globals.IsList(obj)
+        len = length(obj)
+        ret_val = NewPlist(len)
+        recursion_dict[obj] = ret_val
+        for i = 1:len
+             ret_val[i] = julia_to_gap(obj[i], recursion_dict; recursive = recursive)
+        end
+    elseif Globals.IsRecord(obj)
+        # FIXME: add a dedicated method for creating an empty GAP record
+        ret_val = evalstr("rec()")
```
fingolfin, line 209:
```julia
function NewPrecord(capacity::Int)
    o = ccall(:GAP_NewPrecord, Any, (Int,), capacity)
    return o
end
```

src/convert.jl
```diff
@@ -0,0 +1,71 @@
...
+Base.convert(::Type{T}, obj::GapObj; recursive::Bool = true) where {T} = gap_to_julia(T, obj; recursive = recursive)
+
+## The following is needed to disambiguate.
+## It is important not to call `gap_to_julia` here,
+## otherwise we get a circular call in `RAW_GAP_TO_JULIA`;
+## note that `ccall` calls `convert`.
+Base.convert(::Type{Any}, obj::GapObj) = obj
+
+## The following is needed to disambiguate.
+Base.convert(::Type{T}, obj::GapObj) where {T>:Nothing} = convert(Base.nonnothingtype_checked(T), obj)
```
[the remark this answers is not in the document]
ThomasBreuer: When the method was missing, Julia threw an error and gave me the hint to define `convert(::Type{T}, ::GapObj) where T>:Nothing`, and I followed this advice (and the code in `some.jl` from `Base`). I am not sure which solution is preferable.

src/julia_to_gap.jl
ThomasBreuer: This sounds reasonable.
(For programmatic access such as via `convert`, we have to support requests for recursive conversions also for objects which do not have subobjects, and some cases were missing.
Your proposal is more conceptual from this viewpoint.)
Do you have already code for these changes?

src/convert.jl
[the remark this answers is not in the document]
rfourquet: The following is defined in "base/some.jl": `convert(::Type{T}, x) where {T>:Nothing}` which I assume is the method leading to ambiguity? Which in turns comes from a call like `convert(::Union{T,Nothing}, ::GapObj)` (the first argument would choose the `Base` method, but 2nd argument would choose the first method defined above) ?
Also, these two disambiguation methods probably need to accept the `recursive` keyword.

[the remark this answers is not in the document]
fingolfin: I still don't understand why we can't specialize this?
```suggestion
Base.convert(::Nothing, obj::GapObj) = convert(Union{}, obj)
```
Perhaps it be easier to understand for me if I knew what happens of this method is commented out. I'll try it by checking out this PR on my computer


And yeah, I guess we need to add the `recursive` keyword

PR #494, opened 2020-07-16
support computed external references

ThomasBreuer:
Preliminary support for computed external references in the documentation
of GAP.jl had been proposed in pull request #472,
but this was of course not satisfactory because it misused Documenter.jl.

In the discussion of JuliaDocs/Documenter.jl/issues/1343,
@mortenpi suggested to get the required functionality by adding a plugin
to Documented.jl, analogous to an implementation of citations by @ali-ramadhan.
This works well, and from my point of view the issue JuliaDocs/Documenter.jl/issues/1343 can be closed.

Thanks a lot to @mortenpi for the hint, and to @ali-ramadhan for the code.

The details of the changes are:

- added a plugin for Documenter.jl
- replaced the GAP function `UrlOfManualEntry` by `MatchURLs`
  (thanks to @frankluebeck for this code)
- removed the now unnecessary GAP function `ComputeLinksToGAPManuals` and the Julia function `compute_links_to_gap_manuals`

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,4 +1,49 @@
-using Documenter, GAP
+using Documenter, GAP, Markdown
+
+abstract type ExternalReference <: Documenter.Builder.DocumentPipeline end
+
+Documenter.Selectors.order(::Type{ExternalReference}) = 3.1  # After cross-references
+
+function Documenter.Selectors.runner(::Type{ExternalReference}, doc::Documenter.Documents.Document)
+    @info "ExternalReference: building external references."
+    compute_external_references(doc)
+end
+
+function compute_external_references(doc::Documenter.Documents.Document)
+    for (src, page) in doc.blueprint.pages
+        empty!(page.globals.meta)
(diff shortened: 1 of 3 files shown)


docs/make.jl
```diff
@@ -1,4 +1,49 @@
-using Documenter, GAP
+using Documenter, GAP, Markdown
+
+abstract type ExternalReference <: Documenter.Builder.DocumentPipeline end
```
fingolfin, line 3: Some comments in this file that give an idea what's going on here would be really nice. Perhaps also a link to https://github.com/JuliaDocs/Documenter.jl/issues/1343 ?

docs/make.jl
```diff
@@ -1,4 +1,49 @@
...
+
+Documenter.Selectors.order(::Type{ExternalReference}) = 3.1  # After cross-references
```
fingolfin, line 5: What does that do? the 3.1 in particular?

PR #504, opened 2020-07-24
Refactor gap_to_julia for Matrix{T}

fingolfin:


ThomasBreuer:
Nice.
(The point is that GAP's `MatElm` handles unbound entries in GAP 4.11, which was not the case in GAP 4.10.)

fingolfin:
Indeed it does (whether that's a bug or a feature is another matter ;-) Also when converting to, say, `Matrix{Int}` in the end it doesn't matter. But of course it does if one convert to `Matrix{Any}`...

PR #505, opened 2020-07-25
Serialize GAP calls from within Julia.

rbehrends:
This is a draft version for ensuring that GAP is not called from more than one Julia thread at a time.

It currently works with `Threads.nthreads() == 1`, but not if you, say, `export JULIA_NUM_THREADS=4` and then run the following test:

```julia
using GAP

n = 1000
Threads.@threads for i = 1:n
  GAP.Globals.SymmetricGroup(7)
end
```

Most likely, there is still a call to GAP that is not properly synced and causing the resultant crashes.

Closes #483


src/ccalls.jl
```diff
@@ -227,7 +234,7 @@ end
 function _call_gap_func(func::GapObj)
     fptr = GET_FUNC_PTR(func, 0)
     ret = ccall(fptr, Ptr{Cvoid}, (Ptr{Cvoid},), pointer_from_objref(func))
-    return _GAP_TO_JULIA(ret)
+    _GAP_TO_JULIA(ret)
```
fingolfin, line 237: Perhaps this is more Julia idomatic (though I still find it confusing); in any case, if we make this change, then I suggest we make it in _all_ `_call_gap_func` methods at once. And perhaps not in this PR? Unless it is relevant somehow for it?

src/sync.jl
```diff
@@ -0,0 +1,65 @@
...
+        assert(is_locked())
+    end
+end
+
+macro sync(expr)
+    if Threads.nthreads() > 1
+        quote
+            try
+                Sync.lock()
+                $(esc(expr))
```
fingolfin, line 33: Gah, I really can not (yet?) deal with this "value of last expression is automatically returned"... I was about to comment that something like what is done in `sync_noexcept` is needed here, too, but I gather that in a regular execution, this line is considered to be the "last" expression, and its value is returned. Right?

A quick test in Julia seems to confirm that.

src/sync.jl
```diff
@@ -0,0 +1,65 @@
+module Sync
+    global mutex = ReentrantLock()
+    global sync_level = repeat([ 0 ], Threads.nthreads())
```
fingolfin, line 3: Using `global` on the global level is redundant, isn't it? On the other hand, as we won't reassign these values, making them `const` will help type inference.
```suggestion
    const mutex = ReentrantLock()
    const sync_level = repeat([ 0 ], Threads.nthreads())
```

src/sync.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    @inline is_locked() = sync_level[Threads.threadid()] > 0
+
+    @inline function lock()
+        tid = Threads.threadid()
+        if sync_level[tid] == 0
+            Base.lock(mutex)
+        end
+        sync_level[tid] += 1
+    end
```
fingolfin, line 13: I don't understand: We are using a `ReentrantLock`. Why do we need to keep track on every thread whether we already locked the lock? I.e. why do we need `sync_level`?

Is it because locks are per-task but you want per-thread locks? But why? Shouldn't we prevent any attempts to invoke GAP simultaneously from two different tasks, even if they happen to run on the same thread? I mean, GAP still won't expect two concurrent executions and things can and will break in various ways, no?

(BTW, note that at least in some build configurations, tasks can migrate between threads... however, I have no idea if this can happen while the task holds a lock is being held... hopefully not).

src/ccalls.jl
rbehrends: Okay, this is actually an accident. I had started wrapping the `_call_gap_func()` bodies in `@sync` until I realized that this was unnecessary, but apparently had forgotten to revert the change to the first of them. We can revert this one, too.

src/sync.jl
rbehrends: It's something that Julia has borrowed from Ruby (I think) in that every statement also forms an expression. For a block statement, the value of the last statement/expression is the value of the block statement as a whole; ditto for try/catch.

It's actually pretty nice for macros in that you can treat most statements and expressions the same.

src/sync.jl
rbehrends: I had added `global` mostly for readability, I think, but I agree that `const` might be better.

src/sync.jl
rbehrends: This is something that's still a bit of a work in progress.

Julia has `ReentrantLock`, which is reentrant on a per-task basis. It also has `SpinLock`, which works on a per-thread basis, but will eat CPU while blocking. So, in a way, that's a placeholder implementation, and I think I may actually have to wrap a pthread mutex for this to work properly.

Even if `ReentrantLock` works, the `lock()` method for `ReentrantLock()` is comparatively expensive, which is why I tried to optimize it for the nested case.

(`ReentrantLock` working on a per-task basis is also what I suspect may be one of the causes for the crashes I'm seeing.)

src/sync.jl
fingolfin: The `const` is the only way to get type stability. Without it, Julia cannot know that `mutex` always refers to a `ReentrantLock`, and that `sync_level` is always a `Vector{int}`.

rbehrends: Sorry, I had already committed the change, but apparently forgot to push it. It's in there now.

deps/build.jl
```diff
@@ -17,6 +17,7 @@ gapdirname = content[1]
 
 # the location the GAP source code is installed
 gap_src_root = abspath(joinpath(artifact_dir, "$(gapdirname)"))
+gap_src_root = ENV["GAPROOT"]
```
fingolfin, line 20: That shouldn't stay in there, at least not in this form.

src/GAP.jl
```diff
@@ -182,7 +183,7 @@ function initialize(argv::Array{String,1})
     # detect by checking for the value of __JULIAINTERNAL_LOADED_FROM_JULIA).
     append!(argv, ["-c", """BindGlobal("__JULIAINTERNAL_LOADED_FROM_JULIA", true );"""])
 
-    ccall(
+    @sync ccall(
```
fingolfin, line 186: Perhaps it would be better wrap the whole content of this function into a single `@sync` block -- or even simply the call to `initialize`?

deps/build.jl
rbehrends: That was testing code that I accidentally committed. Removed.

src/GAP.jl
rbehrends: Sounds good.

fingolfin:
Did you have a look at replacing `ReentrantLock` by using a `Threads.Condition` wrapped around a `Threads.SpinLock` (as suggested on Slack)?

test/runtests.jl
```diff
@@ -10,4 +10,5 @@ include("convert.jl")
 include("macros.jl")
 include("packages.jl")
 include("help.jl")
+include("sync.jl")
```
fingolfin, line 13: Did you perhaps forget to add `test/sync.jl`?

rbehrends:
> Did you have a look at replacing `ReentrantLock` by using a `Threads.Condition` wrapped around a `Threads.SpinLock` (as suggested on Slack)?

I have played around with various implementations, but haven't really seen any performance differences, especially as there is also the try/catch overhead, which appears to be of about the same magnitude (and also potentially disrupts optimizations). I actually removed the check to see if the lock is already being held, as it performed slightly worse and added unnecessary complexity.

A reentrant thread lock might still be useful in allowing several tasks to run GAP code on the same thread, but the logic then becomes a bit tricky.

I may be looking into biased locking, especially as it might allow us to avoid exception management in some cases. This would also be useful for the Singular.jl code.

src/GAP.jl
```diff
@@ -383,10 +384,10 @@ function prompt()
     # save the current SIGINT handler
     # HACK: the hardcoded value for SIG_DFL is not portable, revise this
     # install GAP's SIGINT handler
-    old_sigint = ccall(:signal, Ptr{Cvoid}, (Cint, Ptr{Cvoid}), Base.SIGINT, C_NULL)
+    old_sigint = @sync ccall(:signal, Ptr{Cvoid}, (Cint, Ptr{Cvoid}), Base.SIGINT, C_NULL)
```
fingolfin, line 387: Does that make sense for a call into the C library? Anyway, perhaps we can just restrict GAP.prompt() to the main thread or so somehow? If we ever end up racing this function from multiple threads, all hell will break loose anyway...

src/sync.jl
```diff
@@ -0,0 +1,131 @@
...
+        end))
+        Sync.eval(:(@inline function sync_noexcept(f::Function)
+            f()
+        end))
+        Sync.eval(:(@inline function check_sync(f::Function)
+            f()
+        end))
+    end
+
+    function switch(mode::SyncMode)
```
fingolfin, line 86: Does this function exist only for debugging / benchmarking? Or whatever else is its purpose? In either case, a brief source comment indicating its purpose would be nice.

src/sync.jl
```diff
@@ -0,0 +1,131 @@
...
+    # the mode at whatever state it was during precompilation. Thus,
+    # initially loading GAP.jl in single-threaded mode would also keep
+    # synchronization off in multi-threaded mode.
+    #
+    # However, Julia tracks function dependencies. If a function changes
+    # upon which another depends, both are being recompiled. Thus,
+    # by installing the proper version during __init__(), we force
+    # selective recompilation of the affected functions as needed.
+
+    function _mode_mutex()
```
fingolfin, line 38: This is unused in practice, it seems (as nothing calls `switch()`, it's single caller). It should either be removed or else a brief source comment be added which explains why it's here.

src/sync.jl
rbehrends: The goal here is to let an (experienced) user switch between locking modes. By default, this pins the GAP interpreter to the main thread. However, one can then also switch to full mutex mode in case GAP needs to be accessed from more than one thread, or to turn off locking in case somebody wants to avoid the overhead involved and takes care of synchronization in a different fashion, e.g. via using `@sync` directly.

I agree that there should be a comment and will be adding one. I'll probably also rename `@sync`, since Julia's `Base.@sync` has different semantics and that would be confusing.

fingolfin:
This needs to be rebased, and we need to figure out what to do with it...

lgoettgens:
This PR is now almost 6 years old and has conflicts in almost every possible line it touches.

@fingolfin do you still see a benefit in keeping this? Or would tackling the underlying issue need a full rewrite due to changes on the julia side anyway?

PR #510, opened 2020-08-14
two-step help request via a menu

ThomasBreuer:
This is a draft proposal of an alternative access to documentation.
Instead of showing all matching entries for a help request (which is Julia's way to look at things), a menu with an overview of the matching entries is shown, from which one can choose what one wants to see.
I think this is more useful than looking through several pages of screen output. (Try `?GAP.Globals.Size`.)

A problem is how to test this kind of help access, since one has to enter some characters (at least the `q` for exiting the menu) in interactive mode.
I have tried to simulate a terminal, but this is too simpleminded, since one gets warnings when the tests are run.

(If this kind of help access is interesting then it would perhaps make sense to provide also a variant that is based on GAP's Browse package instead of Julia's REPL menus.)

### DIFF
--- a/pkg/JuliaInterface/gap/helpstring.g
+++ b/pkg/JuliaInterface/gap/helpstring.g
@@ -380,7 +380,7 @@ end);
 #F  HELP_Info( <string>, <onlyexact> ) . . . . . . . deal with a help request
 #F  HELP_String( <string>, <onlyexact> ) . . . . . . deal with a help request
 ##  
-##  The return value of 'HELP_INFO' is either a list (in cases where just
+##  The return value of 'HELP_Info' is either a list (in cases where just
 ##  one entry shall be shown) or a record with the components 'entries' and
 ##  'menu' (in cases where more than one entry may exist, and where an
 ##  intermediate step via choosing from a menu may be appropriate).
@@ -390,7 +390,7 @@ end);
 ##  and 'start' (the position of the first line to show).
 ##
 ##  The return value of 'HELP_String' is a list that contains the
-##  concatenation of the contents of the result of ''HELP_Info' when this
+##  concatenation of the contents of the result of 'HELP_Info' when this
 ##  is called with the same arguments.
 ##
 DeclareGlobalFunction( "HELP_String" );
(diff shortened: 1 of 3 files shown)


pkg/JuliaInterface/gap/helpstring.g
```diff
@@ -400,16 +377,25 @@ end);
 
 #############################################################################
 ##  
+#F  HELP_Info( <string>, <onlyexact> ) . . . . . . . deal with a help request
 #F  HELP_String( <string>, <onlyexact> ) . . . . . . deal with a help request
 ##  
-##  The return value is a list.
-##  Each of its entries is either a list of strings
+##  The return value of 'HELP_INFO' is either a list (in cases where just
```
fingolfin, line 383:
```suggestion
##  The return value of 'HELP_Info' is either a list (in cases where just
```

pkg/JuliaInterface/gap/helpstring.g
```diff
@@ -400,16 +377,25 @@ end);
...
+##  one entry shall be shown) or a record with the components 'entries' and
+##  'menu' (in cases where more than one entry may exist, and where an
+##  intermediate step via choosing from a menu may be appropriate).
+##
+##  Each of the entries is either a list of strings
 ##  or a record with the components 'lines' (a string or a list of strings)
 ##  and 'start' (the position of the first line to show).
 ##
+##  The return value of 'HELP_String' is a list that contains the
+##  concatenation of the contents of the result of ''HELP_Info' when this
```
fingolfin, line 393:
```suggestion
##  concatenation of the contents of the result of 'HELP_Info' when this
```

fingolfin (commented):
This sounds interesting, I'll give it a try after the weekend.

mohamed-barakat:
Immensely better than the current behavior. Thank you very much.

Why should a solution based on `browse` be more flexible? (this will imply that GAP.jl has to compile `browse`)

ThomasBreuer:
@mohamed-barakat The Browse based help overview in GAP offers features such as searching and filtering, like other Browse tables.
(In GAP with Browse, enter `?Size`, then type `fReference` and hit the `Return` key; this restricts the shown list to those entries that contain the substring `Reference`.)

fingolfin:
What's the status of this? I left some comments in September; once those are resolved, can we just merge this, or is there more work you want to do first? Or what else is this waiting on?

PR #512, opened 2020-08-29
Partial fix for dlsym() not finding symbols.

rbehrends:
This PR is a partial fix for the problem that dlsym() occasionally fails to find library symbols (this has primarily been observed with Julia 1.3.1). These issues seems to arise from conflicting symbol search strategies. By specifying the library to search for symbols, we avoid ambiguities and implementation-defined behavior (for example, that Linux defaults to RTLD_LOCAL for dlopen() and macOS defaults to RTLD_GLOBAL, as well as subtle differences in search strategies for RTLD_DEFAULT).

The net effect of this change is that the problems do not manifest anymore on Julia 1.3 on Linux. However, this will now (reliably) crash on macOS with Julia 1.3 (though not 1.4).

The underlying problem on macOS is that multiple instances of `JuliaInterface.so` are being loaded. This can be illustrated with the following code:

```
using GAP, Libdl

# macOS specific, as Libdl does not define that value.

RTLD_DEFAULT = reinterpret(Ptr{Cvoid}, -2)

p1 = Libdl.dlsym(GAP.JuliaInterface_handle, :gap_julia)
p2 = Libdl.dlsym(RTLD_DEFAULT, :gap_julia)
println("dlsym() on JuliaInterface.so handle: $p1")
println("dlsym() on RTLD_DEFAULT:             $p2")
println("Do these values match?               $(p1 === p2)")
```

These calls should normally return the same result (as `dlopen()` is supposed to not load the same dynamic library twice), but on macOS they don't. Linux does not have this problem, so the problem does not crop up there. For macOS and Julia >= 1.4, we can fix this problem by using `ccall(:sym, ...)` semantics, as in those cases `RTLD_DEFAULT` will be used, which happens to be give us references to the version of the library that GAP uses. However, this is very fragile.

The underlying problem that needs to be fixed is that we have to avoid loading `JuliaInterface.so` twice.

This fragility is also on display with `gap.sh -c 'LoadPackage("JuliaInterface");'`, which will crash reliably on macOS (regardless of the Julia version). This does not happen when calling `LoadPackage("JuliaInterface");` from the prompt. It appears that there are differences in initialization order if `-c` is used, which can be observed by adding debugging output to the GAP.jl initialization. This can probably also be traced back to the issue above.

### DIFF
--- a/pkg/JuliaInterface/init.g
+++ b/pkg/JuliaInterface/init.g
@@ -32,7 +32,7 @@ DeclareUserPreference( rec(
 
 ReadPackage( "JuliaInterface", "gap/JuliaTypeDeclarations.gd");
 
-_PATH_SO:=Filename(DirectoriesPackagePrograms("JuliaInterface"), "JuliaInterface.so");
+_PATH_SO:=Filename(DirectoriesPackagePrograms("JuliaInterface"), ".libs/JuliaInterface.so");
 if _PATH_SO <> fail then
     LoadDynamicModule(_PATH_SO);
 fi;
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -285,7 +285,7 @@ function initialize(argv::Array{String,1})
 
     # open JuliaInterface.so, too
     #global JuliaInterface_path = CSTR_STRING(EvalString("""Filename(DirectoriesPackagePrograms("JuliaInterface"), "JuliaInterface.so");"""))
-    global JuliaInterface_path = joinpath(@__DIR__, "..", "pkg", "JuliaInterface", "bin", sysinfo["GAParch"], ".libs", JuliaInterface)
+    global JuliaInterface_path = normpath(joinpath(@__DIR__, "..", "pkg", "JuliaInterface", "bin", sysinfo["GAParch"], ".libs", JuliaInterface))
     global JuliaInterface_handle = Libdl.dlopen(JuliaInterface_path,
         Libdl.RTLD_GLOBAL)
     CJuliaInterface._load(JuliaInterface_handle)
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -165,6 +165,8 @@ JuliaInterface_path = ""
 libgap_handle = C_NULL
 JuliaInterface_handle = C_NULL
 
+include("cbind.jl")
+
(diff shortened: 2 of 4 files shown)


rbehrends:
Update: it turned out that two different versions of `JuliaInterface.so` (i.e. on different paths) were being loaded. By fixing this, the remaining issues seem to go away, other than the problem with using `JuliaInterface` in `-c`, which appears to be an initialization order issue.

thofma:
So should this be merged here?

rbehrends:
I think it should in order to fix the issues with Julia 1.3, but I'd like somebody else to go over and double-check it.

fingolfin:
Thank you for this PR. Unfortunately I was ill with fever, but will post a full review later.

Please hold off with merging this. I habe some concerns:
1. I don't understand this fix (*why* dies it fix the issue);
2. There is no way to test it via our Travis which makes it hard to test alternatives (I hope this can be resolved; eg by adding another Travis job which tries to load Oscar?)
3. I suspect a much simpler fix is possible but point 2 prevents me from testing my theory.

My concern is that thus fix will calcify because of this... still if I can't find an alternative tonight, I'll merge it so that 1.3 support is back.

rbehrends:
I agree that there should be simpler fix (for now) hiding in there and I really don't want to have this merged without review and discussion.

It is quite likely that an easier fix is possible. I am really not sure why the symbol lookup fails in the first place. The fix basically tries to make sure the lookup goes through the library handle. This is done via a macro so that we can easily change how it is implemented without changing the entire code again. The JuliaInterface.so loading issue is apparently unrelated to that.

fingolfin:
Since you can reproduce the issue somewhere? Could you perhaps bisect to find out which commit introduced the issue.

Written from my kid's favorite playground ;-)

rbehrends:
I can reliably reproduce the issue on the `oscar-test` job on the CI server. I've configured that so that it will retrieve the Jenkins configuration and GAP.jl from test branches and then I can trigger a build. That works reasonably well, but takes about 45 minutes each time and is therefore not really useful for bisecting (though in principle possible). This already strips out unnecessary tests, so it's not easy to make it any shorter without drilling into *why* it happens in the first place, which must have to do with its interaction with other packages.

PR #514, opened 2020-09-04
Fix using GAP.jl from Oscar.jl in Julia 1.3

### LINKED ISSUE #509: problem to load GAP.jl 0.4.3 in Julia 1.3.1
(see the discussion for oscar-system/Oscar.jl/pull/144)

@rbehrends @fingolfin Perhaps you know immediately what is wrong with GAP.jl 0.4.3.
When I install Julia 1.3.1 and enter
```
using Pkg; Pkg.add( PackageSpec(name="GAP", version="0.4.3"))
```
then `using GAP` runs into an error:
```
ERROR: InitError: ccall: could not find function julia_gap
Stacktrace:
 [1] _GAP_TO_JULIA at /home/sam/.julia/packages/GAP/tkivN/src/ccalls.jl:16 [inlined]
 [2] getproperty at /home/sam/.julia/packages/GAP/tkivN/src/ccalls.jl:350 [inlined]
 [3] lastindex(::Main.ForeignGAP.MPtr) at /home/sam/.julia/packages/GAP/tkivN/src/adapter.jl:114
 [4] evalstr(::String) at /home/sam/.julia/packages/GAP/tkivN/src/ccalls.jl:64
 [5] reset_GAP_ERROR_OUTPUT() at /home/sam/.julia/packages/GAP/tkivN/src/GAP.jl:142
```
There is no such problem with GAP.jl 0.4.2 or with the current master branch of GAP.jl.

fingolfin:
Fixes #509
Closes #512 

Better fixes may be possible, but this is the minimal change to make to get things working. We then can quickly release GAP.jl 0.4.4. Since I can now reproduce this locally, it'll be easy enough to experiment with further variations.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -286,7 +286,7 @@ function initialize(argv::Array{String,1})
     global JuliaInterface_handle = Libdl.dlopen(JuliaInterface_path)
 
     # Redirect error messages, in order not to print them to the screen.
-    reset_GAP_ERROR_OUTPUT()
+    Base.invokelatest(reset_GAP_ERROR_OUTPUT)
 end
 
 function finalize()
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -282,7 +282,7 @@ function initialize(argv::Array{String,1})
 
     # open JuliaInterface.so, too
     #global JuliaInterface_path = CSTR_STRING(EvalString("""Filename(DirectoriesPackagePrograms("JuliaInterface"), "JuliaInterface.so");"""))
-    global JuliaInterface_path = joinpath(@__DIR__, "..", "pkg", "JuliaInterface", "bin", sysinfo["GAParch"], ".libs", JuliaInterface)
+    global JuliaInterface_path = normpath(joinpath(@__DIR__, "..", "pkg", "JuliaInterface", "bin", sysinfo["GAParch"], JuliaInterface))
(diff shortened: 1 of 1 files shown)


rbehrends (commented):
I don't see any problems with the patch, but I have to admit that I don't understand how a global variable `libgap_path` broke things?

fingolfin:
I don't understand what's going on either, but I tracked this down via `git bisect` and then experimenting until Oscar.jl worked again. I think it might be a bug in Julia 1.3 possibly related to world age tracking -- indeed, an alternative (and arguably better) fix is this:
```patch
diff --git a/src/GAP.jl b/src/GAP.jl
index 624163a..4017ec1 100644
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -286,7 +286,7 @@ function initialize(argv::Array{String,1})
     global JuliaInterface_handle = Libdl.dlopen(JuliaInterface_path)
 
     # Redirect error messages, in order not to print them to the screen.
-    reset_GAP_ERROR_OUTPUT()
+    Base.invokelatest(reset_GAP_ERROR_OUTPUT)
 end
 
 function finalize()
```

I'll update this PR accordingly.

fingolfin:
@rbehrends PR updated, please have another look

rbehrends (approved):
Ah, that makes sense now.

PR #520, opened 2020-09-17
show GAP traceback on errors thrown by GAP

ThomasBreuer:


ThomasBreuer:
The Travis tests with Julia 1.3 (and only these tests) failed because of the following error.
```
ERROR: LoadError: InitError: ccall: could not find function GAP_CanAssignGlobalVariable
Stacktrace:
 [1] CanAssignGlobalVariable at /home/travis/build/oscar-system/GAP.jl/src/ccalls.jl:86 [inlined]
[...]
```
I cannot imagine that this is due to the change in question. What is going on here?

(And I do not understand the Codecov Report. I got an e-mail message which claims a decrease by 58.28%. The report shown on this page is less dramatic, but it seems to claim that the number of lines has decreased by 4, whereas in fact 3 lines have been added, among them is only one code line.)

fingolfin:
I've closed & reopened this issue to trigger Travis CI again, but I don't think it'll resolve the Julia 1.3 issue. Generally the problem here is that 1.3 is very sensitive (not to say buggy) in the way how and when it evaluates the arguments to `ccall` versus parsing&precompiling. This is why we must use `Base.invokelatest` in various places. I suspect you can fix the problem here with `Base.invokelatest`, too. Of course another fix is to require Julia 1.4.... right now I'd prefer to avoid this, but I hope that once Julia 1.6 is out and made LTS, we can just require that as minimum and get rid of all the workarounds.

ThomasBreuer:
Now two more tests failed, one with Julia 1.4 because of
```
[08:12:03] curl: (22) The requested URL returned error: 429 too many requests
```
and one with Julia nightly because of
```
ERROR: Error building `Singular`:
ERROR: LoadError: libcxxwrap_julia_jll not available on this platform
```
I will update the pull request by using `Base.invokelatest`.

PR #522, opened 2020-09-18
changed JuliaIncludeFile

ThomasBreuer:
- added a two-argument version of `JuliaIncludeFile`, where also the name of the Julia module is specified into which the file shall be included,
- in the one-argument case, use the `Main` module as the default; up to now, `@__MODULE__` had been used, which is usually `Main` in the situation of `JuliaIncludeFile` (and if not then it is better to state explicitly which module is intended),
- documented the above statements, and that the filename must be an absolute path; relative paths will almost always be misunderstood by Julia,
- adjusted `JuliaKnownFiles`, which is used to make sure that files get included only once; now it makes sense to include the same file in serveral modules.

### DIFF
--- a/pkg/JuliaInterface/gap/JuliaInterface.gd
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gd
@@ -152,12 +152,16 @@ DeclareGlobalFunction( "IsArgumentForJuliaFunction" );
 #! @EndExampleSession
 #DeclareGlobalFunction( "JuliaEvalString" );
 
-#! @Arguments filename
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Thanks. I wasn't even aware of `JuliaKnownFiles`, and wonder if it is really a good idea?  Neither GAP's `Read` nor Julia's `include` have this behavior. Perhaps if this was in a function `JuliaIncludeOnce`, I'd understand, but as it is, it seems like an extra complication with limited use?

That said, I am not suggesting it should be remove in this PR.

PR #525, opened 2020-09-30
fixed the @g_str macro

ThomasBreuer:
- changed the code such that escaped characters are handled correctly,
- added comments/examples concerning the limitations,
- added a test

(How can one test using `@test_throws` that a macro throws an error?)

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -74,10 +74,42 @@ Create a GAP string by typing `g"content"`.
 ```jldoctest
 julia> g"foo"
 GAP: "foo"
+
+julia> g"ab\\ncd\\\"ef\\\\gh"   # special characters are handled as in GAP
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
No idea about the macro test, sorry

fingolfin:
I discussed the test_throw issue with people on the Julia slack and based on that I think a good solution would be to move the body of the `g_str` macro into a separate helper function, and then the macro just "calls" the helper, nothing else; and the throw test then can test the helper.

Other than that, feel free to merge this. (Feel free to merge it right away and deal with the exception test later/never, too).

ThomasBreuer:
@fingolfin Thanks for discussing this question with the Julia people.
I have introduced a helper function, as suggested.

PR #532, opened 2020-10-05
added a missing `julia_to_gap` method

ThomasBreuer:
for the case that the keyword argument `recursive` is given
but no dictionary of subobjects.
In this situation,
we assume that the object in question does not have subobjects
because otherwise another method would have dealt with it.


ThomasBreuer:
Two tests failed because of
```
Error: Error building `Singular`
```
Restarting did not help.

fingolfin:
Those failures are due to an outside issue (I am working on that) and can be ignored

fingolfin (approved):

Issue #530, opened 2020-10-05
Feature: add super-fast Julia testers for GAP filters


fingolfin:
... which bypasses the generic code to call into GAP functions, etc., and instead just looks up the relevant bit(s) in the GAP types.

Of course make sure to benchmark this carefully before/after.

Application: having things like `isfinite(G::GAPGroup) = GAP.Globals.IsFinite(G.X)` be as fast as possible (at least as fast as in GAP itself -- benchmark!)

PR #534, opened 2020-10-14
make calling GapObj function objects faster when no keyword arguments are involved

rfourquet:
This is a low-hanging-fruits attemp to make calling functions from `GAP.Globals` faster. This is related to #530 in its goal.

This contains two commits which both improve performance, to such an extent that is a bit mysterious for me. Here is a small demo. First, the baseline, using GAP master and a 3-days old Julia master:
```julia
using GAP

julia> l = GAP.evalstr( "[1,2,3]" )
GAP: [ 1, 2, 3 ]

julia> @btime GAP.Globals.IsFinite($l); 
  750.697 ns (8 allocations: 128 bytes)    

julia> @btime $l[1]
  842.671 ns (8 allocations: 192 bytes)
1
```
The first surprise is the following (I would be curious if someone has any idea what happens there): by simply re-evaling `call_gap_func` at the repl, with the same definition which is in the source code, we get:
```julia
julia> @eval GAP function call_gap_func(func::GapObj, args...; kwargs...)
# ... content not reproduced here
end

julia> @btime GAP.Globals.IsFinite($l);
501.041 ns (5 allocations: 80 bytes)

julia> @btime $l[1];
  843.265 ns (8 allocations: 192 bytes)
```
So the 1-argument call `IsFinite(l)` is significantly sped up, but not so `l[1]`.

Anyway, the first commit consists in writing specialized "call syntax" for cases where no keywords are involved, like
```julia
(func::GapObj)(a1) = call_gap_func(func, a1)
(func::GapObj)(a1, a2) = call_gap_func(func, a1, a2)
```
The example becomes:
```julia
julia> @btime GAP.Globals.IsFinite($l);
  180.778 ns (1 allocation: 16 bytes)

julia> @btime $l[1];
  247.137 ns (2 allocations: 48 bytes)
```
The second commit rewrites a version of `call_gap_func` which doesn't take keyword arguments, i.e.
```julia
call_gap_func_nokw(func::GapObj, args...)
```
And have the above "call syntax" use that, e.g. `(func::GapObj)(a1) = call_gap_func_nokw(func, a1)`. The timings become:
```julia
julia> @btime GAP.Globals.IsFinite($l);
  89.268 ns (0 allocations: 0 bytes)

julia> @btime $l[1];
  71.274 ns (0 allocations: 0 bytes)
```
This is an order of magnitude faster than on master!

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -190,6 +190,14 @@ end
 # make all GapObj callable
 (func::GapObj)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
 
+# specialize non-kwargs versions, which increases performance
+(func::GapObj)(a1) = call_gap_func(func, a1)
+(func::GapObj)(a1, a2) = call_gap_func(func, a1, a2)
+(func::GapObj)(a1, a2, a3) = call_gap_func(func, a1, a2, a3)
+(func::GapObj)(a1, a2, a3, a4) = call_gap_func(func, a1, a2, a3, a4)
+(func::GapObj)(a1, a2, a3, a4, a5) = call_gap_func(func, a1, a2, a3, a4, a5)
+(func::GapObj)(a1, a2, a3, a4, a5, a6) = call_gap_func(func, a1, a2, a3, a4, a5, a6)
+
 #
 # below several "fastpath" methods for call_gap_func follow which directly
 # jump to the C handler functions, bypassing JuliaInterface, for optimal
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -187,16 +187,25 @@ function call_gap_func(func::GapObj, args...; kwargs...)
     end
 end
 
+# specialize call_gap_func for the no-keywords case, for performance
+function call_gap_func_nokw(func::GapObj, args...)
+    if TNUM_OBJ(func) == T_FUNCTION && length(args) <= 6
+        _call_gap_func(func, args...)
+    else
+        ccall(:call_gap_func, Any, (Any, Any), func, args)
+    end
+end
+
 # make all GapObj callable
 (func::GapObj)(args...; kwargs...) = call_gap_func(func, args...; kwargs...)
 
 # specialize non-kwargs versions, which increases performance
-(func::GapObj)(a1) = call_gap_func(func, a1)
-(func::GapObj)(a1, a2) = call_gap_func(func, a1, a2)
-(func::GapObj)(a1, a2, a3) = call_gap_func(func, a1, a2, a3)
-(func::GapObj)(a1, a2, a3, a4) = call_gap_func(func, a1, a2, a3, a4)
-(func::GapObj)(a1, a2, a3, a4, a5) = call_gap_func(func, a1, a2, a3, a4, a5)
-(func::GapObj)(a1, a2, a3, a4, a5, a6) = call_gap_func(func, a1, a2, a3, a4, a5, a6)
+(func::GapObj)(a1) = call_gap_func_nokw(func, a1)
+(func::GapObj)(a1, a2) = call_gap_func_nokw(func, a1, a2)
+(func::GapObj)(a1, a2, a3) = call_gap_func_nokw(func, a1, a2, a3)
+(func::GapObj)(a1, a2, a3, a4) = call_gap_func_nokw(func, a1, a2, a3, a4)
+(func::GapObj)(a1, a2, a3, a4, a5) = call_gap_func_nokw(func, a1, a2, a3, a4, a5)
+(func::GapObj)(a1, a2, a3, a4, a5, a6) = call_gap_func_nokw(func, a1, a2, a3, a4, a5, a6)
 
 #
 # below several "fastpath" methods for call_gap_func follow which directly


fingolfin:
First off: thanks for this.

Second, that's a *major* regression compared to Julia 1.5 then! Ouch! Your PR partially repairs that, but not quite. I'll post my numbers in a separate comment.

Third: One major issue with GAP.jl is that we use "foreign types" in Julia, and those do not play well with precompilation etc. Since the concrete type (MPtr) is only known during runtime, a lot of things in GAP.jl are less efficient than they otherwise could be. One (perhaps far-fetched) idea here is that perhaps one could extend Julia a bit to allow something like `foreign type MPtr <: GapObj end` to "predeclare" a foreign type, so that the precompiler deal with it, even though it is only actually defined at runtime. I don't know what the precompiler really needs, though (I imagine at the very least the supertype and size of an instance -- both we could provide. If that would allow me to write the concrete type `MPtr` instead of the abstract type `GapObj` in a bunch of cases, that would already help.

Not having that, I thought about various ways to "trick" Julia into still allowing me to handle things efficiently, e.g. using parametric types. But in practice, all my experiments became quickly rather unwildly, and/or were confusing to use from outside the package and/or still required outside code using GAP.jl to employ similar tricks.

So I gave up on that for the time being (there are too many other things I need to take of). But I'd be interested in sitting down with you, @rfourquet, to discuss this -- perhaps we can come up with a way to do things better. Also would be happy to discuss this with other Julia devs, but I didn't really know where to start... Oh, see also https://github.com/oscar-system/Oscar.jl/issues/128

Another potential way to gain performance is the following: there are some functions in the GAP kernel which map strings to unique integers. That is, the first time you pass in a string, a unique integer is assigned to it, and stored in a hashtable. Any subsequent invocation with an equal string just performs a hashtable lookup and returns the same integer... This is used behind the scenes to implement the `getproperty` etc. methods for `GAP.Globals`. Now, each time you use `GAP.Globals.FOO`, it has to call into the GAP kernel to map `FOO` to this integer. This is very wasteful. What I'd like to happen is that the first time Julia encounters `GAP.Globals.FOO`, it performs that lookup, causing the GAP kernel to assign that unique integer; and then any subsequent occurrence of `GAP.Globals.FOO` could be transformed into using that now-constant integer value corresponding to `FOO` directly. But the only ways to achieve that that I could think of involve (re-)evaluating any methods using `GAP.Globals.FOO` at runtime... Or perhaps to actually only evaluate such methods at runtime, not (pre)compile time.

For a time, I thought I could (ab)use `@pure` to do this, but that doesn't really work (see also 04e8b2eade7925a8da19928f2efa817f0f7bac93 and 0472308d7e0af3ed8219ce6e067e84f97044fb1c for the commits that revert the short lived code).

Anyway, as a consequence, doing things like `IsFinite = GAP.Globals.IsFinite` and using that instead can help a lot. My benchmark numbers in the next comment will hence include those.

fingolfin:
Here is what I get on my machine with Julia 1.5.2:
```
julia> using GAP, BenchmarkTools

julia> IsFinite = GAP.Globals.IsFinite;

julia> l = GAP.evalstr( "[1,2,3]" );

julia> @btime GAP.Globals.IsFinite($l);
  141.825 ns (1 allocation: 16 bytes)

julia> @btime IsFinite($l);
  104.670 ns (1 allocation: 16 bytes)

julia> @btime $l[1];
  696.676 ns (8 allocations: 192 bytes)

julia> @eval GAP function call_gap_func(func::GapObj, args...; kwargs...) ...

julia> @btime GAP.Globals.IsFinite($l);
  142.923 ns (1 allocation: 16 bytes)

julia> @btime IsFinite($l);
  104.374 ns (1 allocation: 16 bytes)

julia> @btime $l[1];
  732.634 ns (8 allocations: 192 bytes)
```

As you can see, re-evaluating `call_gap_func` makes no difference here.

And here is what I get with Julia master as of this morning. Observe that calls to `IsFinite` got substantially slower, while `l[1]` got faster:
```
julia> using GAP, BenchmarkTools

julia> IsFinite = GAP.Globals.IsFinite;

julia> l = GAP.evalstr( "[1,2,3]" );

julia> @btime GAP.Globals.IsFinite($l);
  592.313 ns (8 allocations: 128 bytes)

julia> @btime IsFinite($l);
  554.021 ns (8 allocations: 128 bytes)

julia> @btime $l[1];
  632.236 ns (8 allocations: 192 bytes)

julia> @eval GAP function call_gap_func(func::GapObj, args...; kwargs...) ...

julia> @btime GAP.Globals.IsFinite($l);
  397.085 ns (5 allocations: 80 bytes)

julia> @btime IsFinite($l);
  354.870 ns (5 allocations: 80 bytes)

julia> @btime $l[1];
  640.646 ns (8 allocations: 192 bytes)
```

fingolfin:
With your changes (which I inserted into my active Julia sessions using `@eval GAP ...`, I get these numbers in Julia 1.5.2:
```
julia> @btime GAP.Globals.IsFinite($l);
  76.358 ns (0 allocations: 0 bytes)

julia> @btime IsFinite($l);
  44.101 ns (0 allocations: 0 bytes)

julia> @btime $l[1];
  62.442 ns (0 allocations: 0 bytes)
```
and these in Julia master:
```
julia> @btime GAP.Globals.IsFinite($l);
  69.618 ns (0 allocations: 0 bytes)

julia> @btime IsFinite($l);
  44.859 ns (0 allocations: 0 bytes)

julia> @btime $l[1];
  54.712 ns (0 allocations: 0 bytes)
```

So, this helps a lot on both Julia versions, and also fixes the sever performance regression with Julia master. And `l[1]` got even faster on Julia master.

For the function calls it might look as if it got faster, too, but I did these measurements on my laptop and fluctuations in the benchmarks are quite high, so I'd be careful about drawing conclusions; I ought to redo these on a benchmark server.  But in the meantime, `@benchmark` of course gives some more info, and that looks promising... with Julia 1.5.2:
```
julia> @benchmark IsFinite($l)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     46.879 ns (0.00% GC)
  median time:      48.424 ns (0.00% GC)
  mean time:        50.552 ns (0.00% GC)
  maximum time:     239.901 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     984
```
And with Julia master:
```
julia> @benchmark IsFinite($l)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     44.773 ns (0.00% GC)
  median time:      46.164 ns (0.00% GC)
  mean time:        47.776 ns (0.00% GC)
  maximum time:     180.966 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     985
```

fingolfin:
BTW here is another optimization I'd like to have: in the above example, I set `IsFinite = GAP.Globals.IsFinite`. From that point on `IsFinite` represent a `T_FUNCTION`, i.e. I *know* that we will end up calling `_call_gap_func`; this implies another tiny performance gain is possible:
```
julia> @btime IsFinite($l);
  41.270 ns (0 allocations: 0 bytes)

julia> @btime GAP._call_gap_func(IsFinite, $l);
  38.013 ns (0 allocations: 0 bytes)
```

Of course one natural way to get *that* speedup would be to use a distinct type for `T_FUNCTION` etc.; right now in GAP, all GAP objects have the same type `MPtr`, but in theory there could be multiple types for special GAP "types", such as `T_FUNCTION` etc. (but one has to be *very* careful there, as not for all GAP "types" is this information truly constant... but it is for `T_FUNCTION`). Alas, changing that is a non-trivial endeavour.

fingolfin (approved):
Nice! I kinda assumed that the no keyword case would automatically be made faster thanks to  the Julia optimizer, but clearly I was wrong.

rfourquet:
Ouch, the performance regression compared to Julia 1.5 is quite worrying!

Yes, I read some issues involving the "foreign" `MPtr` type and the fact that the asbtract type `GapObj` has to be used instead of `MPtr` leading to non-optimal performance, as well as the `@pure` experiment. I don't have yet any idea how to improve that, short of asking the Julia devs to look into it.

Concerning having `IsFinite = Globals.IsFinite`, I actually started preparing a PR doing more or less this, I was not aware of the Oscar issue you linked above (or had forgotten about it ;-) ). I was writing something like
```
const IS_FINITE = Ref{GapObj}()

function IsFinite(x)
    if !isassigned(IS_FINITE)
        IS_FINITE[] = Globals.IsFinite
    end
    IS_FINITE[](x)::Bool
end
```
I used a `Ref` believing that it had to be set at run-time instead of compile time, because the objects changes in each new session. Also note that in your example above, there is a small performance gain to be had by declaring `IsFinite` as `const`. 
I only wrote a couple such functions (`IsFinite` and `ELEM_LIST`) because I wanted to ask for feedback on whether this is a sound approach. On top of this PR, this leads very roughly to a 2x speedup. I assumed these can be changed at run-time by users, but was not sure if they can change all these functions. Also, if these can be changed, do we control through which API on the Julia side? In this case, say it's through `change(IsFinite, ...)`, then `change` could invalidate the cached `Ref`... 

I was not aware of the unique integer associated to a string that you mention above, I will try to understand a bit more about that.

rfourquet:
> Nice! I kinda assumed that the no keyword case would automatically be made faster thanks to the Julia optimizer, but clearly I was wrong.

I would have assumed so too, and the performance difference here (particularly on 1.6) shakes my understanding of Julia's performance model...

fingolfin:
I am somewhat concerned about that 1.6 speed regression. Sure, this PR resolve it -- in this instance. But what other things might be affected, I wonder?

fingolfin:
Anyway, with the microbenchmarks from PR #487 / commit a058bffb6cdffa68da70b5bfa18cd50ada3b0ff2, I get the following results; they confirm that this PR is a great win. Thanks!

Julia 1.5 without your patch:
```
julia> @btime test_simple()
  27.103 ms (0 allocations: 0 bytes)

julia> @btime test_isabelian_1()
  138.854 ms (1000000 allocations: 15.26 MiB)

julia> @btime test_isabelian_2()
  88.585 ms (1000000 allocations: 15.26 MiB)
```

Julia 1.5 with your patch:
```
julia> @btime test_simple()
  27.150 ms (0 allocations: 0 bytes)

julia> @btime test_isabelian_1()
  60.345 ms (0 allocations: 0 bytes)

julia> @btime test_isabelian_2()
  29.990 ms (0 allocations: 0 bytes)
```

Julia master without your patch:
```
julia> @btime test_simple()
  25.936 ms (0 allocations: 0 bytes)

julia> @btime test_isabelian_1()
  598.272 ms (8000000 allocations: 122.07 MiB)

julia> @btime test_isabelian_2()
  548.414 ms (8000000 allocations: 122.07 MiB)
```

Julia master with your patch:
```
julia> @btime test_simple()
  25.845 ms (0 allocations: 0 bytes)

julia> @btime test_isabelian_1()
  60.223 ms (0 allocations: 0 bytes)

julia> @btime test_isabelian_2()
  28.126 ms (0 allocations: 0 bytes)
```

fingolfin:
Curiously, even the best timings here are just about catching up to the numbers I got in July in PR #487 ! So perhaps I should bisect the refactoring I've done since then -- I may have very well introduced a regression at some point :/.

fingolfin:
OK the performance regression comes from a2ef6b3f8f5fdb5f6d7f6639c9b0628ba5125c62. That makes sense, I guess: the purpose was to enable the optimization for arbitrary kinds of arguments, but in doing so, I lost the specialization for the "no kwargs" case.

fingolfin:
```
> const IS_FINITE = Ref{GapObj}()
> 
> function IsFinite(x)
>     if !isassigned(IS_FINITE)
>         IS_FINITE[] = Globals.IsFinite
>     end
>     IS_FINITE[](x)::Bool
> end
> ```
> 
> I used a `Ref` believing that it had to be set at run-time instead of compile time, because the objects changes in each new session.

Well, that is correct in principle. However, I was hoping one could do something more like this (pseudo code):
```
function IsFinite(x)
   const _IsFinite = Globals.IsFinite
   @macro_to_add_new_method IsFinite(x) = _IsFinite(X)::Bool
   return _IsFinite(x)::Bool
end
```
I.e. the method on its first call actually uses `@eval` or so to install a new method which hides this original method and is then used in subsequent calls. I have no idea how well (if at all) this would work in Julia and whether it'd destroy performance for code where this happens a lot -- then again, realistically, it'll be a few dozen such functions that are evaluated once at the start and then never again...

And then if one can do the above, once could write another macro which would make it super convenient to declare a bunch of such wrappers.


> Also note that in your example above, there is a small performance gain to be had by declaring `IsFinite` as `const`.

Sure.

> I only wrote a couple such functions (`IsFinite` and `ELEM_LIST`) because I wanted to ask for feedback on whether this is a sound approach. On top of this PR, this leads very roughly to a 2x speedup. I assumed these can be changed at run-time by users, but was not sure if they can change all these functions.

In GAP, in principle *any* global variable can change its value if you are willing to jump through enough hoops (remove its "read only" flag and then reassign it). But code doing so is rare in practice, and code doing so legitimately is even rarer. I'd be willing to just ignore this for at least core feature like `IsFinite`.

> Also, if these can be changed, do we control through which API on the Julia side? In this case, say it's through `change(IsFinite, ...)`, then `change` could invalidate the cached `Ref`...

The change can be done by GAP code or by Julia code equally. Right now there is no good way to detect this. However, we could patch the GAP kernel (specifically function `AssGVarInternal` in `src/gvars.c`) to add a hook (and possibly submit such a patch for integration into the next GAP release).

Another option would be to extend GAP's notion of a "constant global variable" to be applicable to more things (right now it's restricted to booleans and integers), and then use that on key features like `IsFinite` etc, and *then* use the above caching trick only for constant GAP gvars (where it is safe). That has however the potential of  annoying a few GAP power users who like to hack things. Or not, not sure if anybody ever had need to substitute something else for `IsFinite` ...

rfourquet:
> Well, that is correct in principle. However, I was hoping one could do something more like this (pseudo code):

Ok. I thought using a `Ref` would be negligible, but using `@eval` indeed does seem to shave off like 10ns more.

> That has however the potential of annoying a few GAP power users who like to hack things. Or not, not sure if anybody ever had need to substitute something else for IsFinite ...

A more conservative but invasive possibility is to keep `IsFinite` as is in GAP, but to add a `ConstIsFinite` equivalent which is const (and have `IsFinite` fall back on `ConstIsFinite` on the GAP side), and from within GAP.jl, we use `ConstIsFinite` for our caching, and document explicitly to use `GAP.Globals.IsFinite` if one wants to benefit from the mutability feature, whereas `GAP.IsFinite` would be const... But duplicating dozens of functions like that in GAP seems annoying (and I have strictly no idea such an suggestion would be received by GAP maintainers...)

fingolfin:
I am not so keen on this `ConstIsFinite` idea.

rfourquet:
> I am not so keen on this ConstIsFinite idea.

I understand, this doesn't seem clean and makes things more complex. But you seem to have a nicer solution with your PRs against GAP :)

PR #535, opened 2020-10-15
allow seeding GAP's global RNG

rfourquet:
This is in line with the equivalent in Nemo (https://github.com/Nemocas/Nemo.jl/pull/880). The goal would be to have in Oscar a `Oscar.randseed!(seed)` function which calls the subprojects' `SubProject.randseed!(seed)` functions, to improve reproducibility of the system.

This is assuming single-threaded GAP. I saw there is some specific code around random number generation for HPCGAP, but I didn't look into it (should I?)

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -435,10 +435,6 @@ function randseed!(seed::Union{Integer,Nothing}=nothing)
     if seed === nothing
         seed = rand(UInt128)
     end
-    if 0 <= seed < typemax(Int) ÷ 2
-        # avoid too small seeds by hashing them
-        seed = hash(seed, 0x886b5530f00038ef % UInt)
-    end
     # Although GAP's `Reset` method can accept directly integers (also negative
     # ones), we harmonize here the behavior with what Julia does, i.e. for two
     # integers with the same value, even of different types, the initialization
@@ -446,9 +442,11 @@ function randseed!(seed::Union{Integer,Nothing}=nothing)
     # `Random.make_seed` internal function (which errors on negative argument)
     # to normalize this, and create a string out of it.
     str = String(reinterpret(UInt8, Random.make_seed(seed)))
-    MT = Globals.GlobalMersenneTwister
-    Globals.Reset(MT, str)
-    MT
+    Globals.Reset(Globals.GlobalMersenneTwister, str)
+    # when GlobalRandomSource is reset, the seed is taken modulo 2^28, so we just
+    # pass an already reduced seed here
+    Globals.Reset(Globals.GlobalRandomSource, seed % Int % 2^28)
+    nothing
 end
 
 include("lowlevel.jl")
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -45,6 +45,7 @@ import Base: length, finalize
 
 import Libdl
 import Markdown
+import Random
 
 """
     GapObj
@@ -417,6 +418,37 @@ function prompt()
     # restore GAP error handler
     disable_error_handler = false
     reset_GAP_ERROR_OUTPUT()
+end
+
+"""
+    GAP.randseed!([seed::Integer])
+
+Reseed GAP's global RNG with `seed`.
+
+The given `seed` must be a non-negative integer.
+When `seed` is not specified, a random seed is generated from Julia's global RNG.
+
+For a fixed seed, the stream of generated numbers is allowed to change between
+different versions of GAP.
+"""
+function randseed!(seed::Union{Integer,Nothing}=nothing)
+    if seed === nothing
+        seed = rand(UInt128)
+    end
+    if 0 <= seed < typemax(Int) ÷ 2
+        # avoid too small seeds by hashing them
+        seed = hash(seed, 0x886b5530f00038ef % UInt)
+    end
+    # Although GAP's `Reset` method can accept directly integers (also negative
+    # ones), we harmonize here the behavior with what Julia does, i.e. for two
+    # integers with the same value, even of different types, the initialization
+    # is the same (e.g. with `1` or `0x1`); so we re-use Julia's
+    # `Random.make_seed` internal function (which errors on negative argument)
+    # to normalize this, and create a string out of it.
+    str = String(reinterpret(UInt8, Random.make_seed(seed)))
+    MT = Globals.GlobalMersenneTwister
+    Globals.Reset(MT, str)
+    MT
 end
 
 include("lowlevel.jl")
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -432,17 +432,8 @@ For a fixed seed, the stream of generated numbers is allowed to change between
 different versions of GAP.
 """
 function randseed!(seed::Union{Integer,Nothing}=nothing)
-    if seed === nothing
-        seed = rand(UInt128)
-    end
-    # Although GAP's `Reset` method can accept directly integers (also negative
-    # ones), we harmonize here the behavior with what Julia does, i.e. for two
-    # integers with the same value, even of different types, the initialization
-    # is the same (e.g. with `1` or `0x1`); so we re-use Julia's
-    # `Random.make_seed` internal function (which errors on negative argument)
-    # to normalize this, and create a string out of it.
-    str = String(reinterpret(UInt8, Random.make_seed(seed)))
-    Globals.Reset(Globals.GlobalMersenneTwister, str)
+    seed = something(seed, rand(UInt128))
+    Globals.Reset(Globals.GlobalMersenneTwister, seed)
     # when GlobalRandomSource is reset, the seed is taken modulo 2^28, so we just
     # pass an already reduced seed here
     Globals.Reset(Globals.GlobalRandomSource, seed % Int % 2^28)
--- a/test/basics.jl
+++ b/test/basics.jl
@@ -160,5 +160,4 @@ end
     @test gs2 == [random(GMT, G) for _=1:30]
     @test gs3 == [random(GRS, G) for _=1:30]
 
-    @test_throws DomainError GAP.randseed!(-rand(1:1234))
 end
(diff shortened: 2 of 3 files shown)


src/GAP.jl
```diff
@@ -419,6 +420,37 @@ function prompt()
...
+For a fixed seed, the stream of generated numbers is allowed to change between
+different versions of GAP.
+"""
+function randseed!(seed::Union{Integer,Nothing}=nothing)
+    if seed === nothing
+        seed = rand(UInt128)
+    end
+    if 0 <= seed < typemax(Int) ÷ 2
+        # avoid too small seeds by hashing them
+        seed = hash(seed, 0x886b5530f00038ef % UInt)
```
fingolfin, line 440: What is the issue with "too small seeds" ? Doesn't this effectively introduce bias to the seeds, in that many "small" seeds are impossible (they are mapped to "big" ones by the hashing), but not all (as the hash can of course also output something "small")?

src/GAP.jl
```diff
@@ -419,6 +420,37 @@ function prompt()
...
+    end
+    # Although GAP's `Reset` method can accept directly integers (also negative
+    # ones), we harmonize here the behavior with what Julia does, i.e. for two
+    # integers with the same value, even of different types, the initialization
+    # is the same (e.g. with `1` or `0x1`); so we re-use Julia's
+    # `Random.make_seed` internal function (which errors on negative argument)
+    # to normalize this, and create a string out of it.
+    str = String(reinterpret(UInt8, Random.make_seed(seed)))
```
fingolfin, line 448: I don't mind the use of `Random.make_seed` but I don't understand why it is needed to "harmonize" anything -- why can't one just use e.g. `seed % UInt` ? That "harmonizes" the integer, too, doesn't it?

src/GAP.jl
```diff
@@ -419,6 +420,37 @@ function prompt()
...
+    MT = Globals.GlobalMersenneTwister
+    Globals.Reset(MT, str)
```
fingolfin, line 450: This really should also reset `GlobalRandomSource`; this requires an integer seed (`seed % (2^60)` ought to work).

src/GAP.jl
```diff
@@ -419,6 +420,37 @@ function prompt()
...
+    MT
```
fingolfin, line 451: Why does this return MT?

src/GAP.jl
rfourquet: Yes I will remove that. The reasonning (coming from the equivalent PR in Nemo) was to improve the quality of the random streams given two seeds which are not very "reciprocally random", after reading from [GMP's docs](https://gmplib.org/manual/Random-State-Seeding#Random-State-Seeding)

> The “quality” of the seed is the randomness of a given seed compared to the previous seed used

But IIUC, GAP's Mersenne Twister implementation is from the same authors as what `Random` uses (dSFMT) and might do a better job at handling small seeds. Also, in Nemo, I used a cryptographic hash, but simply using `hash` here doesn't necessarily gives a good distribution.

src/GAP.jl
rfourquet: Doing `seed % UInt` limits the number of possible random sequences to 2^64, when the RNG's state can handle much more. Also, I want to enforce getting two different streams when two different seeds are given. With `seed % UInt`, 1 and `1 + big(2)^64` would lead to the same sequence. So rather than accepting only seeds in the [0, 2^64-1] interval to avoid this problem, we can just accept any seed >= 0 using `Random.make_seed`.

src/GAP.jl
rfourquet: Ah right, I will add this.

src/GAP.jl
rfourquet: The idea was to behave like `Random.seed!`:
```julia
help?> seed!
  seed!([rng=GLOBAL_RNG], seed) -> rng
  seed!([rng=GLOBAL_RNG]) -> rng
```
But as `MT` is not a `AbstractRNG` and not much can be done with it through a Julia API, it's maybe better to return `nothing` instead for now.

src/GAP.jl
rfourquet: > "harmonizes" 

BTW, do you suggest a better term than "harmonize"? :)

rfourquet: > I want to enforce getting two different streams when two different seeds are given 

But then, I can't do that for `GlobalRandomSource`, as it takes the seed module 2^28 :shrug:

fingolfin: I am still confused by how this is phrased. I am actually fine with the word "harmoinzed"; what confuses me why this is an issue that involves `make_seed` in the first place. So yeah, `seed % UInt` truncates values to a finite space; but anyway, all RNGs will use a finite state, including Mersenne twister. But yeah, it cuts too much; so you could use `convert(GapObj, seed)` or perhaps `convert(GapObj, big(seed))` or the same but using `julia_to_gap`, to get a GAP integer, and then pass that on. This would also ensure that 1 and 0x1 are treated the same *and* still allow negative integers.

I am not saying you *should* do it that way; there may be good reasons why  using `Random.make_seed` is superior! Rather, my point is that the argument given in the comment does not fit the conclusion, at least in my mind.

src/GAP.jl
```diff
@@ -419,6 +420,35 @@ function prompt()
...
+    # integers with the same value, even of different types, the initialization
+    # is the same (e.g. with `1` or `0x1`); so we re-use Julia's
+    # `Random.make_seed` internal function (which errors on negative argument)
+    # to normalize this, and create a string out of it.
+    str = String(reinterpret(UInt8, Random.make_seed(seed)))
+    Globals.Reset(Globals.GlobalMersenneTwister, str)
+    # when GlobalRandomSource is reset, the seed is taken modulo 2^28, so we just
+    # pass an already reduced seed here
+    Globals.Reset(Globals.GlobalRandomSource, seed % Int % 2^28)
+    nothing
```
fingolfin, line 449: This looks fine now to me, thanks.

That said... Is there a way in Julia to capture the current RNG state for e.g. serialization, and to possibly restore it later? If so, then returning that here might be reasonable. (But I am *not* requesting that you do this here; I am just musing about this; if at all, this could come in a future PR, or never)

src/GAP.jl
rfourquet: > the argument given in the comment does not fit the conclusion, at least in my mind.

It was just a conclusion in my mind :) It's because I'm fairly familiar with `Random.make_seed`, which transforms an integer into the native seed format of dSFMT (a `Vector{UInt32}`). My reasoning was that constructing in-place a `String` from this array and feeding it to GAP's `Init` method would be optimal. But I had two optimizations bugs:
1) `String(UInt8[...])` constructs the string by re-using the array storage, but here I was using `String(reinterpret(UInt8, UInt32[...]))`, in which case the content of the array is copied;
2) much more important, passing a Julia `String` to `Globals.Reset` results in the argument not being considered as a string on the GAP side (and thought a GAP string is created from that argument, I'm not sure in which way), the solution is of course to call `julia_to_gap(str)`, which improves greatly performance. 

Anyway, in the end using `Random.make_seed` or `convert(GapObj, seed)` actually results in the same performance, so I'm happy with any. I guess you prefer the latter, so I can change the PR to using that.

src/GAP.jl
rfourquet: In Julia, the "idiom" (quotes, as I'm not sure many people know it!) for capturing the RNG state is simply to call `copy` on the RNG object, and the state can be restored later with `copy!`:
```julia

julia> rng = MersenneTwister(); rand(rng); rng2 = copy(rng); rng == rng2
true

julia> rand(rng) == rand(rng2)
true

julia> rand(rng); copy!(rng, rng2); rng == rng2
true
```
Here, there is a complication, because two global RNGs are involved. With one RNG, my take would be to have a Julia RNG wrapper (subtype of `AbstractRNG`) around GAP's `IsMersenneTwister` and implement `copy` / `copy!`. Then, like `Random.seed!`, `GAP.randseed!` could simply return an RNG object which wraps GAP's `GlobalMersenneTwister`. But given that two RNGs are involved here, it seems simpler to return `nothing`, but we could very well wrap both RNG types and return a tuple containing both RNGs.

src/GAP.jl
fingolfin: No, I honestly have no preference between `Random.make_seed` or `convert(GapObj, seed)` -- but I do want to be able to understand the final code, if possible, and reading the comment and code it felt to me as if something was missing -- either in my understanding, or in the code/comment, or both. :-)

rfourquet: So I updated with passing directly the integer seed to GAP's seeding routine. It's more straightforward, and actually allows to get the same streams than in GAP for a given seed, which might be a feature. The comment containing "harmonize" is thus deleted :)

rfourquet: I don't know if there will be interest for #544, but having it this way also wouldn't create discrepancy between `randseed!(seed)` and `Random.seed!(GAP.GlobalMersenneTwister, seed)` or `Random.seed!(GAP.default_rng(), seed)` (whatever we call it).

fingolfin (approved):
Thanks!

test/basics.jl
```diff
@@ -134,3 +134,31 @@ end
...
+    GAP.randseed!(seed)
+    gs1 = [random(G) for _=1:30]
+    gs2 = [random(GMT, G) for _=1:30]
+    gs3 = [random(GRS, G) for _=1:30]
+    GAP.randseed!(seed)
+    @test gs1 == [random(G) for _=1:30]
+    @test gs2 == [random(GMT, G) for _=1:30]
+    @test gs3 == [random(GRS, G) for _=1:30]
+
+    @test_throws DomainError GAP.randseed!(-rand(1:1234))
```
fingolfin, line 163: This test fails:
```
randseed!: Test Failed at /Users/travis/build/oscar-system/GAP.jl/test/basics.jl:163
  Expression: GAP.randseed!(-(rand(1:1234)))
    Expected: DomainError
  No exception thrown
```

I think it is outdated and can simply be removed?
```suggestion
```

PR #536, opened 2020-10-16
added `hash` methods for `GapObj` and `FFE`

ThomasBreuer:
because otherwise some `==` results can be wrong

### DIFF
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -257,3 +257,10 @@ for (left, right) in typecombinations
         end
     end
 end
+
+# Since we define the equality of GAP objects (see above),
+# we must provide a safe `hash` method, otherwise `==` results may be wrong.
+# For example, `==` for `Set`s of GAP objects may erroneously return
+# `false` if the default `hash` is used.
+Base.hash(::GapObj) = 0
+Base.hash(::FFE) = 0
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -262,5 +262,5 @@ end
 # we must provide a safe `hash` method, otherwise `==` results may be wrong.
 # For example, `==` for `Set`s of GAP objects may erroneously return
 # `false` if the default `hash` is used.
-Base.hash(::GapObj) = 0
-Base.hash(::FFE) = 0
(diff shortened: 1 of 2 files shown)


src/adapter.jl
```diff
@@ -257,3 +257,10 @@ for (left, right) in typecombinations
         end
     end
 end
+
+# Since we define the equality of GAP objects (see above),
+# we must provide a safe `hash` method, otherwise `==` results may be wrong.
+# For example, `==` for `Set`s of GAP objects may erroneously return
+# `false` if the default `hash` is used.
+Base.hash(::GapObj) = 0
```
rfourquet, line 265: The 2-argument method of `hash` must be extended instead:
```suggestion
Base.hash(::GapObj, h::UInt) = xor(h, 0x0e5c5aeb80a55fa3 % UInt)
```
I added above a magic constant, as is usually the case when writing `hash` methods, to possible improve a bit performance in some (unlikely) cases where another `hash` method is defined for another type which also would return `h` (but at this point this seems paranoid and `hash` can also be just made to return `h` here).

test/basics.jl
```diff
@@ -22,6 +22,25 @@
...
+    x = GAP.evalstr("Z(2)")
+    y = GAP.evalstr("Z(4)^3")
+    @test !(x === y)
+    @test (x == y)
+    @test hash(x) == 0
+
+    # The following would sometimes fail if no dedicated `hash`
+    # method would be available for GAP objects.
+    x = Set{Any}([GAP.evalstr("()"), GAP.evalstr("(1,2)")])
+    y = Set{Any}([GAP.evalstr("()"), GAP.evalstr("(1,2)")])
```
fingolfin, line 42: If it only happens "sometimes" perhapps we should use a loop here to execute that test 100 times or so to increase the chance of catching the issue?

fingolfin (approved):
Left one minor remark but mergeing as-is is also fine by me!

test/basics.jl
ThomasBreuer: Thanks, I have updated the pull request.

PR #538, opened 2020-10-19
add iteration for GAP objects

rfourquet:
This is an alternative to #537 to create Julia containers from GAP containers, but seems more general. For `Set` creation, it's faster than #537, except if the GAP list has a lot of duplicates. Demo:

With a Set constructor (#537):
```julia
julia> lj = rand(1:999, 999); l = GAP.julia_to_gap(lj);

julia> @btime Set{Int}($l);
  382.950 μs (2369 allocations: 144.41 KiB)

julia> @btime Set{Any}($l);
  535.653 μs (4511 allocations: 177.88 KiB)

julia> lj = rand(1:99, 999); l = GAP.julia_to_gap(lj);

julia> @btime Set{Int}($l);
  118.623 μs (220 allocations: 22.99 KiB)
```
This PR:
```julia
julia> lj = rand(1:999, 999); l = GAP.julia_to_gap(lj);

julia> @btime Set{Int}($l);
  197.645 μs (4944 allocations: 142.65 KiB)

julia> @btime Set{Any}($l);
  227.194 μs (5518 allocations: 151.62 KiB)

julia> lj = rand(1:99, 999); l = GAP.julia_to_gap(lj);

julia> @btime Set{Int}($l);
  177.146 μs (4006 allocations: 127.99 KiB)

julia> @btime Set{Int}(GAP.Globals.AsSet($l));
  92.361 μs (411 allocations: 22.51 KiB)
```
This shows that when the list is expected to have duplicates, calling manually `GAP.Globals.AsSet` can solve the slow-down.

This approach works also for creating other containers. For example we can now call `collect`, which is faster than calling GAP's `Vector{T}` method:
```julia
julia> @btime Vector{Int}($l);
  364.568 μs (1982 allocations: 58.66 KiB)

julia> @btime collect(Int, $l);
  161.715 μs (3489 allocations: 109.28 KiB)
```

What this doesn't solve is the recursivity, i.e. you can call `Vector{Any}(x, recursive=true)` (similar for the proposed `Set` constructor in #537). But there is a workaround: you can lazily map `gap_to_julia` to the iterator. In julia 1.6, you can use `Iterators.map` (and in 1.5 and older, generator expressions can be used):
```julia
julia> lj = [rand(1:9, 2) for _=1:3]; l = GAP.julia_to_gap(lj, recursive=true)
GAP: [ [ 8, 7 ], [ 6, 6 ], [ 7, 2 ] ]

julia> Set(Iterators.map(x -> GAP.gap_to_julia(x, recursive=true), l))
Set{Vector{Any}} with 3 elements:
  Any[6, 6]
  Any[8, 7]
  Any[7, 2]
```
This is not strictly equivalent to using the `recursive` keyword from a dedicated `Set` constructor like in #537, as this doesn't use a `recursion_dict` at the top-level. But we could imagine an orthogonal approach to handle recursive conversion. Here is below a possible solution: a `Recursive` (poorly named) wrapper to a GAP iterable object, which caches the construction of a `recursion_dict` (implying that the performance is improved, and matches #537 at least in a couple of tests):
```julia
struct Recursive{T}
    x::GapObj
    d::IdDict
end

Recursive{T}(x) where {T} = Recursive{T}(x, IdDict())
Recursive(x) = Recursive{Any}(x, IdDict())

Base.length(r::Recursive) = length(r.x)

function Base.iterate(r::Recursive{T}, st...) where T
    res = iterate(r.x, st...)
    res === nothing && return nothing
    x, st = res
    gap_to_julia(T, x, r.d, recursive=true), st
end

Base.eltype(::Type{GAP.Recursive{T}}) where T = T
```
Demo:
```julia
julia> Set(GAP.Recursive{Vector{Int}}(l))
Set{Vector{Int64}} with 3 elements:
  [6, 6]
  [8, 7]
  [7, 2]

julia> Set(GAP.Recursive(l))
Set{Any} with 3 elements:
  Any[6, 6]
  Any[8, 7]
  Any[7, 2]
```
Defining iteration on `GapObj` doesn't seem ideal, as most objects are presumably not iterable, but it's good enough for me to just error when the object is not iterable and we try to iterate. On the performance side: if we add more supported objects to be iterated, the benchmarks shown above _might_ look less good (for example, I see that in the `gap_to_julia(::Type{<Vector}, ...` [method](https://github.com/oscar-system/GAP.jl/blob/d743256c8afe8ee1f7a883844d612bacecb99814/src/gap_to_julia.jl#L183), the `elmlist` method depends on the type of the object, so this might introduce overhead here).

### DIFF
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -288,7 +288,7 @@ function Base.iterate(obj::GapObj, (i, len)::Tuple{Int,Any})
     ElmList(obj, i), (i+1, len)
 end
 
-function Base.iterate(obj::GapObj, iter)
+function Base.iterate(obj::GapObj, iter::GapObj)
     if Globals.IsDoneIterator(iter)
         nothing
     else
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -274,13 +274,12 @@ Base.literal_pow(::typeof(^), x::GapObj, ::Val{-1}) = Globals.INV_MUT(x)
 # iteration
 
 function Base.iterate(obj::GapObj)
-    islist = Globals.IsList(obj)
-    islist || Globals.IsCollection(obj) ||
-        throw(ArgumentError("object cannot be iterated"))
-    if islist
+    if Globals.IsList(obj)
         iterate(obj, (1, Globals.Length(obj)))
-    else
+    elseif Globals.IsCollection(obj)
         iterate(obj, Globals.Iterator(obj))
+    else
+        throw(ArgumentError("object cannot be iterated"))
     end
 end
 
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -274,12 +274,26 @@ Base.literal_pow(::typeof(^), x::GapObj, ::Val{-1}) = Globals.INV_MUT(x)
 # iteration
 
 function Base.iterate(obj::GapObj)
-    Globals.IsList(obj) || Globals.IsCollection(obj) ||
+    islist = Globals.IsList(obj)
+    islist || Globals.IsCollection(obj) ||
         throw(ArgumentError("object cannot be iterated"))
-    iterate(obj, (1, Globals.Length(obj)))
+    if islist
+        iterate(obj, (1, Globals.Length(obj)))
+    else
+        iterate(obj, Globals.Iterator(obj))
+    end
 end
 
-function Base.iterate(obj::GapObj, (i, len))
+function Base.iterate(obj::GapObj, (i, len)::Tuple{Int,Any})
     i > len && return nothing
     ElmList(obj, i), (i+1, len)
+end
+
+function Base.iterate(obj::GapObj, iter)
+    if Globals.IsDoneIterator(iter)
+        nothing
+    else
+        x = Globals.NextIterator(iter)
+        (x, iter)
+    end
 end
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -271,3 +271,15 @@ Base.hash(::FFE, h::UInt) = h
 # only if `x` is a multiplicative element in the sense of GAP.
 Base.literal_pow(::typeof(^), x::GapObj, ::Val{-1}) = Globals.INV_MUT(x)
 
+# iteration
+
+function Base.iterate(obj::GapObj)
+    Globals.IsList(obj) || Globals.IsCollection(obj) ||
+        throw(ArgumentError("object cannot be iterated"))
+    iterate(obj, (1, Globals.Length(obj)))
+end
+
+function Base.iterate(obj::GapObj, (i, len))
+    i > len && return nothing
+    ElmList(obj, i), (i+1, len)
+end
--- a/test/adapter.jl
+++ b/test/adapter.jl
@@ -1,0 +1,9 @@
+@testset "iteration" begin
+    l = GAP.evalstr("[1, 2, 3]")
+    lj = collect(l)
+    @test lj isa Vector{Any}
+    @test lj == [1, 2, 3]
+    lj = collect(Int, l)
+    @test lj isa Vector{Int}
+    @test lj == [1, 2, 3]
+end
--- a/test/adapter.jl
+++ b/test/adapter.jl
@@ -6,4 +6,11 @@
     lj = collect(Int, l)
     @test lj isa Vector{Int}
     @test lj == [1, 2, 3]
+
+    s = GAP.Globals.SymmetricGroup(3)
+    xs = []
+    for x in s
+        push!(xs, x)
+    end
+    @test length(xs) == 6
 end


rfourquet:
Just a comment to clarify that the benchmark in the OP is not "fair" to the `Set` implementation, as it uses `recursive=true` by default, and not invoking keywords is probably what people would do in general... Here is the same benchmark without recursivity:
```julia
julia> @btime Set{Int}($l, recursive=false);
  167.061 μs (1257 allocations: 93.74 KiB)

julia> @btime Set{Any}($l, recursive=false);
  298.929 μs (3288 allocations: 125.48 KiB)

julia> lj = rand(1:99, 999); l = GAP.julia_to_gap(lj);

julia> @btime Set{Int}($l, recursive=false);
  87.090 μs (116 allocations: 15.66 KiB)
```
So this is closer to the timings we get with iteration. Also, it would probably be possible to avoid creating the `recursion_dict` in the `Set` constructor when the target type is not recursive.

src/adapter.jl
```diff
@@ -264,3 +264,16 @@ end
 # `false` if the default `hash` is used.
 Base.hash(::GapObj, h::UInt) = h
 Base.hash(::FFE, h::UInt) = h
+
+# iteration
+
+function Base.iterate(obj::GapObj)
+    Globals.IsList(obj) || Globals.IsCollection(obj) ||
+        throw(ArgumentError("object cannot be iterated"))
```
fingolfin, line 272:
```suggestion
    IsListOrCollection(obj) || throw(ArgumentError("object cannot be iterated"))
```

src/adapter.jl
```diff
@@ -264,3 +264,16 @@ end
...
+    iterate(obj, (1, Globals.Length(obj)))
```
fingolfin, line 273: That won't work for infinite obj, though, and even for finite ones can be very expensive as it forces a size computation. Granted, cases where this matter are more the exception than the norm.

src/adapter.jl
```diff
@@ -264,3 +264,16 @@ end
...
+end
+
+function Base.iterate(obj::GapObj, (i, len))
+    i > len && return nothing
+    ElmList(obj, i), (i+1, len)
```
fingolfin, line 278: ElmList is only suitable for lists, not for non-list collections. For those you really need an iterator.

For example, take `GAP.Globals.SymmetricGroup(3)`

fingolfin (changes requested):

fingolfin:
If `Vector{Int}(l)` is slower than `collect(Int, l)` doesn't that suggest a (most likely fixable) inefficiency in our `Vector` constructor, rather than indicating a true win for iterators?

rfourquet:
> If Vector{Int}(l) is slower than collect(Int, l) doesn't that suggest a (most likely fixable) inefficiency in our Vector constructor, rather than indicating a true win for iterators?

Indeed. Like I said in my previous comment (for `Set`), using `recursive=false` makes it way faster, like 8x, which suggests that the recursive case could be improved.

src/adapter.jl
rfourquet: Yes I'm awayre it's not complete. For now I just copied the method in the `Set` constructor from Thomas PR, which uses `ElmList`, and accepts only `Globals.IsList(obj) || Globals.IsCollection(obj) `. I didn't add for example ``Globals.IsVectorObj` as I didn't know yet how to construct and test such object.

src/adapter.jl
rfourquet: So yes, maybe we should define `Base.IteratorSize` to return `SizeUnknown()` (but I copied from Thomas' PR the iteration modality; I didn't know how to iterate otherwise).

ThomasBreuer:
If we are now talking about general `iterate` methods for iterating over GAP objects, the relevant parts from the discussion of pull request #323 should perhaps be revisited.
(And if the result is that we do not want such an `iterate` feature then we should document this.)

src/adapter.jl
fingolfin: so for Thomas' PR, we know that we need all objects eventually, so in the infinite case, we'd have a problem anyway (we still might want to catch that explicitly a la `Globals.IsFinite(l) || error("cannot convert infinite collections")`

src/adapter.jl
fingolfin: so the way to do general iteration is GAP is to use ... an iterator ;-). But that comes with an overhead, esp. since GAP cannot optimize the iterator object away, so it's desirable to use `ElmList` when possible (for `IsList`), but otherwise use the iterator.

In GAP one can write `for x in coll do ... od` which is syntactic sugar for roughly this:
```gap
  iter := Iterator(coll)
  while not IsDoneIterator(iter) do
    x := NextIterator(iter);
  od;
```

fingolfin:
Could you please rebase this?

src/adapter.jl
rfourquet: So I rebased, but still without including general iteration yet, I will look into that now.

test/adapter.jl
```diff
@@ -0,0 +1,16 @@
...
+    lj = collect(Int, l)
+    @test lj isa Vector{Int}
+    @test lj == [1, 2, 3]
+
+    s = GAP.Globals.SymmetricGroup(3)
+    xs = []
+    for x in s
+        push!(xs, x)
+    end
+    @test length(xs) == 6
```
rfourquet, line 15: So here, we can't do `collect(s)` because `collect` tries to get the length of `s`, which doesn't work. The correct lenght of `s` in this context would be to use `Globals.Order(s)` I believe. So should we make `length` return `Order` here, and similarly for other iterable GAP objects, or simply defining `Base.IteratorSize(x::GapObj) = Base.SizeUnknown()` ?

ThomasBreuer: From GAP's point of view, `Globals.Size` is the function that returns the number of elements in the collection.

(And from the viewpoint of what iterators are used for in GAP, if the code would be an example how to proceed then I would not collect the elements in a list but just increase a counter for each element: The main idea in GAP is to avoid creating a full elements list whenever it is likely that either one can stop after one has found a suitable element or at least one creates only one element at a time, which saves space.)

rfourquet: > From GAP's point of view, Globals.Size is the function that returns the number of elements in the collection.

Ah thank you! 

Concerning your comment inside parentheses, I think I understand the content, but maybe not the implications for this PR... are you referring to my test code above where I collect the iterators's elements in a list? Usually it's not necessary to "collect" elements of an iterator in a list, but inevitably we collect them anyway once in a while, if only at the REPL to "see" the content of the iterator, for small collections, so supporting `collect` would be good.

src/adapter.jl
```diff
@@ -271,3 +271,29 @@ Base.hash(::FFE, h::UInt) = h
...
+
+function Base.iterate(obj::GapObj)
+    islist = Globals.IsList(obj)
+    islist || Globals.IsCollection(obj) ||
+        throw(ArgumentError("object cannot be iterated"))
+    if islist
+        iterate(obj, (1, Globals.Length(obj)))
+    else
+        iterate(obj, Globals.Iterator(obj))
+    end
```
fingolfin, line 284: Seems a bit artificial to force the error to be at the start... how about this?
```suggestion
    if Globals.IsList(obj)
        iterate(obj, (1, Globals.Length(obj)))
    elseif Globals.IsCollection(obj)
        iterate(obj, Globals.Iterator(obj))
    else
        throw(ArgumentError("object cannot be iterated"))
    end
```

src/adapter.jl
fingolfin: Seems you've done it :-)

src/adapter.jl
```diff
@@ -271,3 +271,29 @@ Base.hash(::FFE, h::UInt) = h
...
+end
+
+function Base.iterate(obj::GapObj, (i, len)::Tuple{Int,Any})
+    i > len && return nothing
+    ElmList(obj, i), (i+1, len)
+end
+
+function Base.iterate(obj::GapObj, iter)
```
fingolfin, line 292: Perhaps this?
```suggestion
function Base.iterate(obj::GapObj, iter::GapObj)
```

fingolfin (approved):

PR #540, opened 2020-10-20
optimize Vector{T} constuctor by avoiding elmlist variable

rfourquet:
The `elmlist` variable was storing either the `ElmList` or the `Globals.ELM_LIST`
object, and was called later as `elmlist(obj, i)`. But unfortunately,
this is not (yet) a pattern that the Julia compiler optimizes well.
So instead, just write explicitly the calls to `ElmList` and
`ELM_LIST`, with a branch.

The timings for
```julia
lj = rand(1:99, 999); l = GAP.julia_to_gap(lj); @btime Vector{Int}($l);
```
look like
```
  337.507 μs (1982 allocations: 58.66 KiB) # master
  102.435 μs (1494 allocations: 35.42 KiB) # PR
  46.805 μs (980 allocations: 23.56 KiB) # master + PR 539
  31.162 μs (1491 allocations: 31.55 KiB)  # PR + PR 539
```
Together with PR 539, this is still an improvement, but involves more allocations, I'm not sure why. By putting the `if islist` test out of the loop (meaning duplicating the loop), the extra allocations disappear, and we get 
```
26.122 μs (492 allocations: 15.94 KiB)
```
But it's annoying to duplicate code like that, so I'd rather not do so too hastily.

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -187,11 +187,10 @@ function gap_to_julia(
     recursive = true,
 ) where {T}
     if Globals.IsList(obj)
-        elmlist = ElmList  # returns 'nothing' for holes in the list
+        islist = true
     elseif Globals.IsVectorObj(obj)
-        # vector objects aren't lists,
-        # but the function for accessing entries is `ELM_LIST`
-        elmlist = Globals.ELM_LIST
+        islist = false
+        ELM_LIST = Globals.ELM_LIST
     else
         throw(ConversionError(obj, Vector{T}))
     end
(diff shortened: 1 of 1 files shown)


ThomasBreuer:
This is interesting.
The point seems to be that `GAP.ElmList` is a Julia function whereas `GAP.Globals.ELM_LIST` is a `GAP.ForeignGAP.MPtr` object.
(I have looked at related places in the code, but I did not find other instances of this situation.)

src/gap_to_julia.jl
```diff
@@ -201,7 +200,13 @@ function gap_to_julia(
...
         recursion_dict[obj] = new_array
         for i = 1:len_list
-            current_obj = elmlist(obj, i)
+            if islist
+                current_obj = ElmList(obj, i)  # returns 'nothing' for holes in the list
+            else
+                # vector objects aren't lists,
+                # but the function for accessing entries is `ELM_LIST`
+                current_obj = ELM_LIST(obj, i)
+            end
```
fingolfin, line 209: I wonder if we could achieve the same (but perhaps more elegantly?) by moving this loop into a separate helper function which takes `elmlist` as an argument -- then at least in theory, Julia should be able to generate optimal code in both cases, too?

rfourquet:
> The point seems to be that GAP.ElmList is a Julia function whereas GAP.Globals.ELM_LIST is a GAP.ForeignGAP.MPtr object.

Yes; if both were functions, the situation would be the same because each function has its own type. But I agree it's tempting to factor out the code by putting the function in a variable; I'm not clear on whether this is an optimization that we can hope the compiler to implement any time soon...

fingolfin:
@rfourquet could you please rebase this?

PR #541, opened 2020-10-20
added `literal_pow` method for inverting

ThomasBreuer:
Thanks to @rfourquet for mentioning `literal_pow`.
It can be used for improving `x^-1` but not for improving `x^0` and `x^1`
(where `x` is a GAP object).

### DIFF
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -264,3 +264,10 @@ end
 # `false` if the default `hash` is used.
 Base.hash(::GapObj, h::UInt) = h
 Base.hash(::FFE, h::UInt) = h
+
+# The following bypasses GAP's redirection of `x^-1` to `INV_MUT(x)`.
+# Installing analogous methods for `x^0` and `x^1` would *not* be allowed,
+# these terms are equivalent to `ONE_MUT(x)` and `x`, respectively,
+# only if `x` is a multiplicative element in the sense of GAP.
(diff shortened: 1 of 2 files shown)


fingolfin:
There is a conflict which needs to be resolved (rebase on latest master?)

Also the PR description doesn't really make it clear to me what this about: x^-1 already works for me now. So perhaps this is an optimization?

ThomasBreuer:
I thought the term "improve" in the description would be clear enough; besides that, the comment in the code should explain what this is about.
To be even more explicit: My main motivation was to document that `literal_pow` can be used for improving `x^-1` but not for improving `x^0` and `x^1`, according to the situation in GAP.

fingolfin:
I guess my main issue is that I don't understand in which sense this "improves" anything. What does `x^-1` do right now?

ThomasBreuer:
@fingolfin See [the paragraph on literal_pow in the Julia documentation](https://docs.julialang.org/en/v1/base/math/#math-ops):
The Julia method for computing `x^y` of a GapObj `x` calls GAP's `x^y`; and if `y` is `-1` then GAP will then call `InverseSameMutability(x)`.
With the `literal_pow` method in question, `x^-1` goes directly to GAP's `InverseSameMutability(x)`.

fingolfin (approved):

PR #544, opened 2020-10-22
wrap GAP's IsMersenneTwister RNG

rfourquet:
This is a proof of concept of what I was mentioning at https://github.com/oscar-system/GAP.jl/pull/535#discussion_r510148942.

This gives you a Julia handle on GAP's `IsMersenneTwister` objects, named `GAP.MersenneTwister` here. A little demo:
```julia
julia> r = GAP.MersenneTwister()
GAP.MersenneTwister(GAP: <RandomSource in IsMersenneTwister>)

julia> rand(r, 1:9)
8

julia> s = copy(r); rand(r, 1:9) == rand(s, 1:9)
true

julia> rand(r, 1:9, 3)
3-element Vector{Int64}:
 6
 4
 5

julia> copy!(r, s); rand(r, 1:9, 3)
3-element Vector{Int64}:
 6
 4
 5

julia> Random.seed!(r, 123); Random.seed!(s, 123); rand(r, 1:9, 3) == rand(s, 1:9, 3)
true

julia> g = GAP.default_rng() # handle on Globals.GlobalMersenneTwister
GAP.MersenneTwister(GAP: <RandomSource in IsMersenneTwister>)


julia> rand(g, GAP.Globals.SymmetricGroup(9), 3) # ideally, `g` would be the default here, but there is no support (yet) from the `Random` module
3-element Vector{Any}:
 GAP: (1,4,8,9,6,7,3,2,5)
 GAP: (1,5,9,7)(3,6,4,8)
 GAP: (1,7,6,8)(2,5,9,4,3)
```
I didn't write tests yet, nor a wrapper for `IsGAPRandomSource`, but can do so if there is interest for continuing this PR.

### DIFF
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -265,6 +265,64 @@ end
 Base.hash(::GapObj, h::UInt) = h
 Base.hash(::FFE, h::UInt) = h
 
+### RNGs
+
+using Random: Random, AbstractRNG, rand
+
+abstract type AbstractGAPRNG <: AbstractRNG end
+
+struct MersenneTwisterState state end
+
+struct MersenneTwister <: AbstractGAPRNG
+    ptr::GapObj
+
+    function MersenneTwister(; state::MersenneTwisterState=MersenneTwisterState(nothing))
+        if state.state === true
+            new(Globals.GlobalMersenneTwister)
+        elseif state.state === nothing
+            new(Globals.RandomSource(Globals.IsMersenneTwister))
+        else
+            new(Globals.RandomSource(Globals.IsMersenneTwister, state.state))
+        end
+    end
+
+    MersenneTwister(seed::Integer) = new(Globals.RandomSource(Globals.IsMersenneTwister, seed))
+end
+
+default_rng() = MersenneTwister(state=MersenneTwisterState(true))
+
+function Random.seed!(rng::MersenneTwister, seed::Integer)
+    Globals.Reset(rng.ptr, seed)
+    rng
+end
+
+function Base.copy!(dst::MersenneTwister, src::MersenneTwister)
+    Globals.Reset(dst.ptr, Globals.State(src.ptr))
+    dst
+end
+
+Base.copy(rng::MersenneTwister) = MersenneTwister(state=MersenneTwisterState(Globals.State(rng.ptr)))
+
+## rand methods
+
+Random.rand(rng::AbstractGAPRNG, x::Random.SamplerTrivial{<:Obj}) = Globals.Random(rng.ptr, x[])
+
+Random.rand(rng::AbstractGAPRNG, x::Random.SamplerTrivial{<:AbstractUnitRange}) =
+    Globals.Random(rng.ptr, julia_to_gap(first(x[])), julia_to_gap(last(x[])))
+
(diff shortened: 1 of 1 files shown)


fingolfin:
I am not sure what the goal of this is... but in general, I'd prefer to keep code in GAP.jl to that what's strictly necessary to access GAP from Julia and vice versa, and some uncontroversial "sugar coating"; but higher level wrappers are something I'd rather see in Oscar.jl. I guess the RNG is a bit of a border case, but e.g. I wouldn't want GAP.jl to depend on AbstractAlgebra.jl.

Hence my question: is there a use case you have in mind for this? or some other compelling arguments for why this should be here, and not e.g. in Oscar.jl?

ThomasBreuer:
@fingolfin My understanding of @rfourquet 's idea is that Oscar would like to have access to `rand` functionality for objects from the involved systems, in a uniform way. In this sense, this pull request provides an interface between Julia's `rand` functions and GAP's `Random` functions. The point seems to be what the right place for this kind of interfaces is. Currently GAP.jl is "the" interface in the middle between Oscar and GAP, and thus one is tempted to put such code there.
One can argue that GAP.jl should provide only a low level interface, then we have to think about a suitable place for higher level code.

rfourquet:
So first, this was a "proof of concept" to help explaining what I had in mind in the comment I linked in the OP, where you were asking how to "capture the current RNG state". 
Otherwise, I agree with what @ThomasBreuer said above.

To give a concrete example, we have in Oscar.jl the following method: ` Base.rand(x::GAPGroup)`, which uses GAP's global RNG.It doesn't follow the `Random` API, so of course it doesn't mean it's not useful, but it's limited. One idea would be to
1. define this method as as expected by the Random API: ` Base.rand(rng::GAP.MersenneTwister, x::Random.SamplerTrivial{GAPGroup})`.
2. have something like https://github.com/JuliaLang/julia/pull/38043 merged in Julia so that `rand(gapgroup)` can fall back to `rand(GAP.GlobalMersenneTwister, x)`

1) alone already gives some advantage (e.g. you can create an array of thing with the usual `rand` syntax), but it forces to pass an explicit RNG. 
For 2) I don't how likely this is going to be merged in Base. If not, or in the meantime, it's still possible to keep the `Base.rand(x::GAPGroup)` definition around. 

I personally find that this code fits in GAP.jl, but I don't really mind. And anyway, in isolation, I don't know if it's very useful to merge this change; I think this should go in tandem with making (few) changes in Oscar.jl as highlighted above, i.e. allowing passing an explicit RNG to `rand` methods using GAP's random generation.

ThomasBreuer:
Meanwhile there is the directory `Oscar.jl/src/GAP`, intended for code that belongs to the interface between GAP and Oscar but on a higher level than the GAP-Julia interface provided by GAP.jl.
Would this be an appropriate place for the code from this pull request?

ThomasBreuer:
Taking the current discussion on using GAP and Julia random sources in both systems, and the use cases from GroupsCore.jl, I think that the changes from this pull request should become part of GAP.jl. @fingolfin Shall we merge the pull request?

fingolfin (approved):
@ThomasBreuer if you want to merge this, fine by me! I've just rebased it.

Issue #545, opened 2020-10-22
Regression through commit `fix conversion of strings containing null characters`


mohamed-barakat:
Commit d2f51deaa30f introduces the following behavior:
```
julia> using GAP

julia> GAP.julia_to_gap("𝓞(1)")
GAP: "𝓞"
```

This leads to this [issue](https://github.com/homalg-project/HigherHomologicalAlgebra/issues/14).

PR #547, opened 2020-10-22
fix string conversion

ThomasBreuer:
Use `sizeof` not `length` in order to get the number of bytes in a string.

This fixes the problem from issue #545.


mohamed-barakat (approved):

mohamed-barakat:
Thank you very much @ThomasBreuer.

Issue #518, opened 2020-09-16
ordering of initialization steps


ThomasBreuer:
GAP.jl's initialization depends on libgap's `GAP_Initialize`, which runs GAP's initialization.
In particular, autoloading GAP packages is part of this process.
Only afterwards, GAP.jl loads the GAP package JuliaInterface.

As soon as one wants to autoload JuliaInterface (not very likely) or a GAP package that depends on JuliaInterface (might become more likely), calling `using GAP` runs into the following problem.
```
WARNING: using GAP.GAP in module Main conflicts with an existing identifier.
ERROR: importing GAP into Main conflicts with an existing identifier
```
This happens after the execution of the `__init__` function of GAP.jl.

Do we have to forbid autoloading JuliaInterface and related packages?

ThomasBreuer:
Here is another instance of the same problem:
When one starts GAP such that the name of an input file belongs to the command line then this file will be read before the initialization process has loaded JuliaInterface.
Thus one cannot use the GAP-Julia integration in the input file.
(Explicitly loading JuliaInterface in the input file leads to the above error, and GAP crashes.)

fingolfin:
So we really should load JuliaInterface earlier.

We discussed this a bit in a Jitsi meeting. I stumbled over related issues when working on the `initialize` code -- part of the problem there was that GAP starts by reading `init.g`, which does a lot, including the processing of the command line arguments -- in fact we might end up quitting in there, if the user passed a file that quits.

One trick I employed to deal with this was to insert extra arguments to be passed to `GAP_Initialize`. We could try the same for loading JuliaInterface, and add the equivalent of `-c LoadPackage("JuliaInterface")` to the *start* of the `argv` vector. Of course that's yet another hack, but it should at least work right now. We can then work on a "proper" solution on the GAP side for future GAP releases.

ThomasBreuer:
@fingolfin I understand GAP's `lib/init.g` as follows.
1. The steps until the user preferences in `gap.ini` get evaluated seem to be harmless in the sense that the GAP-Julia integration is not affected.
2. Then `AutoloadPackages` is called. Here JuliaInterface may be requested (perhaps indirectly) for loading.
3. Then some more GAP internal initializations happen. The only part that  is relevant for Julia is that the `gaprc` file may already try to use the Julia integration (set some variables, read some user files, ...).
4. Then the files mentioned in the command line (`GAPInfo.InitFiles`) are read.

At the moment, I see the following two issues.

[...]

PR #548, opened 2020-10-30
try to load JuliaInterface earlier

ThomasBreuer:
such that it can be used by other autoloaded packages etc., see issue #518.

This is work in progress, see the discussion for issue #518.

### DIFF
--- a/lib/systemfile.g
+++ b/lib/systemfile.g
@@ -1,0 +1,27 @@
+(function()
+    local deps, i, standalone, initfiles, entry;
+
+    # Add JuliaInterface to the needed GAP packages, such that it is loaded
+    # early enough for being required by other GAP packages,
+    # and such that it is available when user files get read
+    # via the GAP command line.
+    deps:= SHALLOW_COPY_OBJ( GAPInfo.Dependencies.NeededOtherPackages );
+    APPEND_LIST_INTR( deps, [ [ "JuliaInterface", ">= 0.4.4" ] ] );
+    GAPInfo.Dependencies:= MakeImmutable( rec( NeededOtherPackages:= deps ) );
+
+    # Notify that GAP.jl initiates the loading process.
+    for i in [ 1 .. LEN_LIST( GAPInfo.SystemCommandLine )-1 ] do
+      if GAPInfo.SystemCommandLine[i] = "-c" then
+        if POSITION_SUBSTRING( GAPInfo.SystemCommandLine[i+1], "__JULIAINTERNAL_LOADED_FROM_JULIA_STAND_ALONE\", true", 0 ) <> fail then
+          standalone:= true;
+        elif POSITION_SUBSTRING( GAPInfo.SystemCommandLine[i+1], "__JULIAINTERNAL_LOADED_FROM_JULIA_STAND_ALONE\", false", 0 ) <> fail then
+          standalone:= false;
+        fi;
+      fi;
+    od;
+    if IsBound( standalone ) then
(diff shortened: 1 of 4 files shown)


lib/systemfile.g
```diff
@@ -0,0 +1,27 @@
...
+        elif POSITION_SUBSTRING( GAPInfo.SystemCommandLine[i+1], "__JULIAINTERNAL_LOADED_FROM_JULIA_STAND_ALONE\", false", 0 ) <> fail then
+          standalone:= false;
+        fi;
+      fi;
+    od;
+    if IsBound( standalone ) then
+      # Set the flag that notifies the 'JuliaInterface' package:
+      # GAP.jl has initiated to load GAP.
+      BIND_GLOBAL( "__JULIAINTERNAL_LOADED_FROM_JULIA", true );
+    fi;
```
fingolfin, line 26: All this just to set a GAP variable. I see three alternatives to this:
1. have two different `systemfile.g`, one for standalone mode, one for non-standalone ("embedded") mode
2. only in standalone mode, inject code just before the start of `init.g` which set `__JULIAINTERNAL_LOADED_FROM_JULIA` to `true`; two ways to achieve this come to mind:
    1. provide our own `init.g` and ensure it is visible in a GAPROOT before the one with the real `init.g`. Then have that custom `init.g` set `__JULIAINTERNAL_LOADED_FROM_JULIA` and read the "real" `init.g`.
    2. set the C kernel variable `SyLoadSystemInitFile` to 0 (zero). That ensures that `GAP_Initialize` returns *before* `init.g` is read (but well after the kernel itself is set up, so we can e.g. assign to GAP global variables). It also means that we then need to issue the equivalent of `READ("lib/init.g")`.

fingolfin: OK, I missed your comments on issue #518, otherwise my comment above would be a bit different :-). Let me first say that I think the idea of adding `--systemfile` is the right call, and I fully support it. I just think it's not mutually exclusive with inserting some code before reading `init.g`. I think the two would complement each other quite nicely. I might hack up the `SyLoadSystemInitFile` idea just to try it ...

In fact, I am actually think about combining the `SyLoadSystemInitFile` approach with the switch to GAP_jll anyway: the idea being that we move the definition of `GapObj` as well as the call to `GAP_Initialize` into `GAP_jll`, but with `SyLoadSystemInitFile` set to 0, so that it does not process `init.g`. 

By doing this, when `GAP.jl` is loaded, `ForeignPtr.MPtr` should already be defined, and it might be possible for us to do some things that are not possible right now, or at least do them in a somewhat cleaner fashion. Like writing things such as `const T_MACFLOAT = ValueGlobalVariable(:T_MACFLOAT)` should become possible at the top level of the package. I guess in a sense this would bring back the old `GAPTypes` package, at least in spirit, but also way beyond it.  (However, I need to think about whether and how this can be made to work while retaining support for loading GAP first, then loading Julia from GAP).

fingolfin:
Tests fails because `--systemfile` is not (yet) available, as we are compiling an outdated snapshot of GAP 4.11. We'll have to update to a newer one by (a) creating it and (b) editing `Artifacts.toml`...  Or alternatively by getting `GAP_jll` to work which makes this kind of update relatively easy (basically, update https://github.com/JuliaPackaging/Yggdrasil/blob/master/G/GAP/build_tarballs.jl to use a newer GAP commit and/or use different patch files in https://github.com/JuliaPackaging/Yggdrasil/tree/master/G/GAP/bundled/patches -- which reminds me, this kind of thing should also be documented...).

Anyway, we'll get there.

PR #554, opened 2020-11-04
Migrate CI from Travis to GitHub Actions

fingolfin:
One thing I don't know yet is how to setup Slack notifications into the build system.

UPDATE: one could probably use https://github.com/marketplace/actions/slack-notify-build for that but I won't look into that today.


fingolfin:
OK, doctest's are simply broken in various Julia versions. I'll refrain from trying to fix them all here, and instead move that to a separate PR

PR #556, opened 2020-11-04
Re-enable doctests in CI

fingolfin:
Currently these fail for various reasons, I don't have time to track that down, but didn't want to forget about it.

There are at least two options for us how to run the tests: either we run them along with the rest of the test suite; this way we get most coverage, but the downside is that we need to make sure the doctests pass in all tested Julia versions, which puts an extra maintenance burden on us. Or we could just run the doctests in one config, and make sure it works there. 

I've enabled both variants here, we can later remove the one we don't want.

One issue is that the printing of `Set` instances seems to vary between Julia versions or perhaps even in a single Julia version based on some random seed? (Perhaps @rfourquet knows more).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -107,12 +107,9 @@ jobs:
             using Pkg
             Pkg.develop(PackageSpec(path=pwd()))
             Pkg.instantiate()'
-      #- name: Run doctests
-      #  run: |
-      #    julia --project=docs --color=yes -e '
-      #      using Documenter, GAP
-      #      DocMeta.setdocmeta!(GAP, :DocTestSetup, :(using GAP); recursive = true)
-      #      doctest(GAP)'
+      - name: Run doctests
+        run: |
+          julia --project=docs --color=yes test/doctest.jl
       - name: Deploy documentation
         run: julia --project=docs --color=yes docs/make.jl
         env:
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -362,7 +362,7 @@ involves only native Julia objects such as integers and strings.
 Dealing with results containing GAP objects will be inefficient.
 
 # Examples
-```jldoctest
+```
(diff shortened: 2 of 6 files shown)


rfourquet:
> One issue is that the printing of Set instances seems to vary between Julia versions

Yes, `hash` is not guaranteed to be stable across Julia minor versions. 

> or perhaps even in a single Julia version based on some random seed?

But I'm pretty sure that no randomness is involved (of course, besides the fact that `objectid` can be used and can look random-ish, but it's not the case here in the failing tests apparently), and normally hash is stable for a given minor version.

fingolfin:
Thanks for the clarification @rfourquet. Based on that it'd seem best to me if we focused on just getting the doctests to pass in one Julia version, say the latest stable 1.x release, which is 1.5.2. I just pushed a commit which fixes doctests for me locally in Julia 1.5.2. Unfortunately, it still fails on GitHub Actions in a test involving `@gap (1,2)(3,4)`: for some reason, the error we expect it to trigger (and that I get on my laptop) is different from the error that actually is triggered. I have no idea why :-(.

ThomasBreuer:
@fingolfin I get the different error for `@gap (1,2)(3,4)` with Julia 1.3, but only after I had upgraded Documenter.jl to v0.25.3.
In the Julia session, I still get the expected error, but the error during doctests is different.
(I wanted to try the new Documenter.jl because it contains a fix concerning doctestfilters, which would be good for the GAP.jl tests, but now there is this new problem ...)

fingolfin:
Mostly works now, except with Julia nightlies: on macOS, those are known to be broken (see https://github.com/JuliaLang/julia/issues/38782); but on Linux, that's because they tweaked how types are printed: e.g. `Array{Int64,1}` became `Array{Int64, 1}`. I guess we can adjust the transformations in `test/doctest.jl` to deal with that? @ThomasBreuer ?

PR #557, opened 2020-11-04
JuliaInterface: remove include(Base.find_package("GAP"))

### LINKED ISSUE #419: problem with `import` statements in files loaded via `JuliaIncludeFile`
Currently I run into the following problem.
I start GAP, load its JuliaInterface package, and then call `JuliaIncludeFile` for a file containing a Julia module.
If this Julia module tries to import GAP (which makes sense if the Julia code wants to call GAP functions) and also some other Julia module then strange things happen.

**Case 1**:
The included file is `testGP.jl` and looks as follows.
```
module testGP
import GAP; println( "after import GAP" )
import Primes; println( "after import Primes" )
end
```
The GAP session looks as follows.
```
gap> LoadPackage( "juliai", false );;
gap> Julia.Base.sqrt(5);                 # the interface works
<Julia: 2.23606797749979>
gap> JuliaIncludeFile( "testGP.jl" );    # including seems to work
after import GAP
after import Primes
gap> Julia.Base.sqrt(5);                 # but now the interface is broken
Error, MethodError: Cannot `convert` an object of type Main.ForeignGAP.MPtr to an obj\
ect of type Main.GAP.GapObj
[...]
```

**Case 2**:
The included file is `testPG.jl` and looks as follows.
```
module testPG
import Primes; println( "after import Primes" )
import GAP; println( "after import GAP" )
end
```
The GAP session looks as follows.
```
gap> LoadPackage( "juliai", false );;
gap> Julia.Base.sqrt(5);                # the interface works
<Julia: 2.23606797749979>
gap> JuliaIncludeFile( "testPG.jl" );   # including runs into an error
after import Primes
Error, LoadError: InitError: MethodError: Cannot `convert` an object of type Main.For\
eignGAP.MPtr to an object of type Main.GAP.GapObj
[...]
```
If the Julia module imports just GAP or does not import GAP then the problem does not occur.
The effect is not changed when I replace `Primes` by `Nemo`.

(I am using Julia 1.4.0 with the GAP 4.11.0 which was installed by Julia using the current standard procedure, together with the master branch of GAP.jl, whose JuliaInterface was compiled relative to Julia's GAP 4.11.0.
GAP must be started with `-E`, otherwise command line editing does not work; as far as I understand, this is the problem described in issue #415.)

### LINKED ISSUE #461: a change due to `GAP.prompt()`
In a GAP session started from Julia with `GAP.prompt()`, one gets the following.
```
gap> JuliaModule( "GAP" );
<Julia: GAP>
```
In a GAP session started in the traditional way, one gets the following.
```
gap> JuliaModule( "GAP" );
<Julia: Main.GAP>
```
(I just stumbled over this because it occurs in one of the test files for JuliaInterface.)

fingolfin:
Fixes #419 
Fixes #461

### DIFF
--- a/pkg/JuliaInterface/gap/JuliaInterface.gd
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gd
@@ -283,7 +283,7 @@ DeclareGlobalFunction( "ImportJuliaModuleIntoGAP" );
 #!  see <Ref Func="ImportJuliaModuleIntoGAP"/>.
 #! @BeginExampleSession
 #! gap> gapmodule:= JuliaModule( "GAP" );
-#! <Julia: Main.GAP>
+#! <Julia: GAP>
 #! gap> gapmodule = JuliaPointer( Julia.GAP );
 #! true
 #! @EndExampleSession
--- a/pkg/JuliaInterface/read.g
+++ b/pkg/JuliaInterface/read.g
@@ -6,10 +6,6 @@
 ReadPackage( "JuliaInterface", "gap/JuliaInterface.gi");
 ReadPackage( "JuliaInterface", "gap/BindCFunction.gi" );
 
-if not IsBound( __JULIAINTERNAL_LOADED_FROM_JULIA ) then
-    JuliaEvalString( "Base.MainInclude.include(Base.find_package(\"GAP\"))" );
-fi;
-
 _JULIAINTERFACE_INTERNAL_INIT();
 
 ## The GAP module is also bound to the variable __JULIAGAPMODULE,
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -267,11 +267,11 @@ function __init__()
         ENV["TERMINFO_DIRS"] = joinpath(@__DIR__, "..", "deps", "usr", "share", "terminfo")
     end
 
-    ## Older versions of GAP need a pointer to the GAP.jl module during
-    ## initialization, but at this point Main.GAP is not yet bound. So instead
-    ## we assign this module to the name __JULIAGAPMODULE.
-    ## Newer versions of GAP won't need this; however, JuliaInterface still
-    ## uses it.
+    ## At this point, the GAP module has not been completely initialized, and
+    ## hence is not yet available under the global binding "GAP"; but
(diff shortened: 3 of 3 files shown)


fingolfin:
This fails the GAP based tests, of course, for the reasons @ThomasBreuer explained elsewhere (I shouldn't make PRs so late at night ;-).

Anyway, perhaps this can now fixed by combining it with PR #548 suitably, plus some extra work? Namely: merge the changes in PR #548 and this PR; then use the `--systemfile` to also disable (or rather, defer) processing of the command line arguments. Let GAP.jl's `__init__` end, now `GAP` is defined globally. Finally, process the command line arguments, by modifying the `gap.sh` produced by `deps/build.jl`. Which right now does this:
```
__GAP_ARGS__ = ARGS
using GAP
ccall(:Call0ArgsInNewReader, Cvoid, (Any,), GAP.Globals.SESSION)
```
We could now insert another call after the `using GAP`, something like `GAP.process_command_line_arguments_now()`.

ThomasBreuer (approved):
Thanks.
The combination of this change and the idea to read GAP's command line files from `gap.sh` (why didn't I get this idea?) solves our problems concerning the initialization.
I will modify #548 accordingly.

fingolfin:
> why didn't I get this idea?

I've been asking that myself about what you did 😄 . So I think this is simply a great example of how in a collaboration one sometimes can come up with better solutions than alone (or at least gets much faster to those).

fingolfin:
@ThomasBreuer huh, did you merge this on purpose?

ThomasBreuer:
@fingolfin Sorry.
I wanted to put my #548 on top of this. That was a bad idea.
(#548 has "Do not merge" for good reasons, since it needs a new GAP release.)
How to get out of this mess?

fingolfin:
I've reverted the offending commit via `git revert`. You'll have to be a bit careful in rebasing #548; one way to do it would be this, I guess:
```
git checkout master
git pull # to get my revert commit
git checkout TB_load_order_for_GAP
git rebase --onto master HEAD^^  # take the three top commits in your branch along (HEAD. HEAD^, HEAD^^)
```

PR #550, opened 2020-10-31
Switch to using GAP_jll

fingolfin:
Incomplete and with several issue, some really bad, but I wanted to have it here for reference

Closes #421
Closes #450 
Closes #468
Closes #548
Closes #564 

Resolves #518
Resolves #285
Fixes #465
Resolves #433 (well, arguably at least, via our custom gaproot)

PR #559, opened 2020-11-11
Improve _GAP_TO_JULIA

fingolfin:
Reimplement more of the julia_gap logic; this makes the cases we now handle on
the Julia side potentially a bit faster (??? untested) but most importantly it
means this code now works even if JuliaInterface is not yet loaded. On the
downside, if we end up ccalling julia_gap anyway, it performs a bunch of
checks a second time.


rfourquet:
> requested review from rfourquet 

I'm afraid this is a bit too advanced into GAP's internals for me to give useful feedback... 

The tests fail because 

> ERROR: LoadError: InitError: could not load library "libgap"
> libgap.so: cannot open shared object file: No such file or directory

fingolfin:
I'll just try to get this in as part of #550 instead of trying to figure out  how to make it work in isolation

PR #562, opened 2020-11-26
Remove JuliaBindCFunction

### LINKED ISSUE #349: problem with JuliaBindCFunction in JuliaInterface
When one uses the interface from the GAP side, the GAP function `JuliaBindCFunction` is intended to take a Julia function, create a C-compatible function using `@cfunction`, and then wrap this into a GAP "kernel" function.
Currently this is broken, the examples in `JuliaInterface/example/gap_function_call.gi` run into errors.
```
gap> fun:= JuliaEvalString( "function mysquare( n ) return n^2; end" );
<Julia: mysquare>
gap> cfun:= JuliaBindCFunction( "mysquare", [ "nn" ] );
function( nn ) ... end
gap> cfun( 2 );
Error, MethodError: no method matching ^(::Ptr{Nothing}, ::Int64)
[...]
```

fingolfin:
This was a very strange function, and was difficult to use correctly
without a crash, and also very limited in its applications.

It was mainly meant to provide a performance boost; but I think for
real world applications, we need a different approach which takes
Julia's multiple dispatch and argument type specialization properly
into account.

Fixes #349


ThomasBreuer (approved):
(I will change the `GASMAN` calls in the examples to `CollectGarbage` calls as soon as this function is available in the underlying GAP.)

PR #564, opened 2020-11-29
Change default install dir for GAP packages

fingolfin:
Right now, `GAP.Packages.install` installs GAP packages  by default into `GAPROOT * "/pkg"`, which means: into the GAP 4.11.0 artifact. This is bad because that artifact is supposed to be read-only, and we only succeed because Julia only makes the files in artifacts readonly, not the directories, so we can add new files, just modify existing ones. Still this is not a good thing.

This PR changes it to `~/.gap/pkg` -- I should probably say "back to" because I think it was there before and we changed it after some back and forth. 

I've now opened this PR so we can discuss this (again?).

See also https://github.com/gap-system/gap/issues/3917 where the suggestion is to use `~/.gap/v4.11/pkg`.

Another idea would be to put this into a Julia specific path, say `~/.julia/gap/pkg` or ~/.julia/gap/v4.11/pkg` -- but then this directory will never be cleaned unless we teach the user about it and ask them to clean it. (Once we drop compatibility with Julia versions before 1.5, we could switch to using https://github.com/JuliaPackaging/Scratch.jl which allows creating scratch spaces that are removed when the Julia package is removed -- i.e. users who delete GAP.jl get their disk space back eventually.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -79,7 +79,9 @@ export LoadPackageAndExposeGlobals
 
 module Packages
 
-import ...GAP: Globals, julia_to_gap, GAPROOT
+import ...GAP: Globals, julia_to_gap
+
+const DEFAULT_PKGDIR = joinpath(get(ENV, "HOME", ""), ".gap", "pkg")
 
 """
     load(spec::String, version::String = ""; install = false)
@@ -129,7 +131,7 @@ containing a package, or the URL of a `PackageInfo.g` file.
 The function uses [the function `InstallPackage` from GAP's package
 `PackageManager`](GAP_ref(PackageManager:InstallPackage)).
 """
-function install(spec::String; interactive::Bool = true, pkgdir::AbstractString = GAPROOT * "/pkg")
+function install(spec::String; interactive::Bool = true, pkgdir::AbstractString = DEFAULT_PKGDIR)
     res = load("PackageManager")
     @assert res
     # FIXME: crude hack to allow PackageManager 1.0 shipped with GAP
@@ -158,7 +160,7 @@ containing a package, or the URL of a `PackageInfo.g` file.
 The function uses [the function `UpdatePackage` from GAP's package
 `PackageManager`](GAP_ref(PackageManager:UpdatePackage)).
 """
-function update(spec::String; interactive::Bool = true, pkgdir::AbstractString = GAPROOT * "/pkg")
+function update(spec::String; interactive::Bool = true, pkgdir::AbstractString = DEFAULT_PKGDIR)
     res = load("PackageManager")
     @assert res
 
@@ -181,7 +183,7 @@ Return `true` if the removal was successful, and `false` otherwise.
 The function uses [the function `RemovePackage` from GAP's package
 `PackageManager`](GAP_ref(PackageManager:RemovePackage)).
 """
-function remove(spec::String; interactive::Bool = true, pkgdir::AbstractString = GAPROOT * "/pkg")
+function remove(spec::String; interactive::Bool = true, pkgdir::AbstractString = DEFAULT_PKGDIR)
     res = load("PackageManager")
     @assert res
 


mohamed-barakat (changes requested):
This change will cause me several troubles. I also know of others who dislike the idea for the following reason: I would expect the packages in `~/.gap/pkg` to be compiled against my "default" GAP. Currently, the GAP in `GAP.jl` is not my default GAP. I can even imagine a scenario where another package in Julia having `GAP.jl` as a secondary dependency (making even the expert user unaware of it) and this `GAP.jl` starts meddling around in such an important subdirectory which is crucial for the correct functionality of the default GAP. That said, I would be in favor of a `GAP.jl`-specific solution. The one I like most is `~/.julia/gap/v4.11/pkg`.

fingolfin:
First off, let me emphasis that I by no means want to push hard for `~/.gap/pkg`; rather, I want to thoroughly explore the problem scape.

As it is, I am confused as to why there would be a problem with both "regular GAP" and "julia GAP" using the same dir -- they already put their binaries into separate locations, after all, and I can use the same copy of e.g. IO in both regular GAP and Julia GAP just fine:
- `bin/x86_64-apple-darwin18.7.0-default64-kv8/io.so` is used by regular GAP on my macOS 10.14.6 mac
- `bin/x86_64-apple-darwin14-julia64-kv7/io.so` is used by the Julia GAP.

i should also point out that the "important directory" `~/.gap` for me is almost empty and only contains the files `gap.ini` and `history` -- I never use it to install packages. And I suspect many user won't use it for that either; they will either only use the packages bundled with GAP; or install further packages into GAP's `pkg` dir (as the installation instructions for most packages describe it that way). Only with `PackageManager` does this kinda change, and it might be argued that this is a problem with PackageManager...

mohamed-barakat:
> First off, let me emphasis that I by no means want to push hard for `~/.gap/pkg`; rather, I want to thoroughly explore the problem scape.

Good :) I am not also very much against it if there are alternatives :)
 
> As it is, I am confused as to why there would be a problem with both "regular GAP" and "julia GAP" using the same dir -- they already put their binaries into separate locations, after all, and I can use the same copy of e.g. IO in both regular GAP and Julia GAP just fine:
> 
> * `bin/x86_64-apple-darwin18.7.0-default64-kv8/io.so` is used by regular GAP on my macOS 10.14.6 mac
> * `bin/x86_64-apple-darwin14-julia64-kv7/io.so` is used by the Julia GAP.

I was unaware that a package compiled under the control of GAP.jl gets a different binary path under `bin`. This would indeed resolve the issue of packages with a binary part.
 
> i should also point out that the "important directory" `~/.gap` for me is almost empty and only contains the files `gap.ini` and `history` -- I never use it to install packages. And I suspect many user won't use it for that either; they will either only use the packages bundled with GAP; or install further packages into GAP's `pkg` dir (as the installation instructions for most packages describe it that way). Only with `PackageManager` does this kinda change, and it might be argued that this is a problem with PackageManager...

Maybe normal users would prefer the `pkg` directory of the default GAP as they are using packages developed by others. I am thinking about developers who need a directory to place their developed package or distributed packages they contribute to. At least in my case, I want to have a directory that is independent of any GAP installation. I don't remember when I switched to `.gap/pkg` but maybe, as you said, it was the influence of `PackageManager`.

I would be extremely happy if GAP would by default look for packages in the directory `~/.gap/dev`, very much like the way Julia behaves. This would resolve for me all issues.

ThomasBreuer:
I interpret Mohamed's argument as follows:
There is a GAP installation (used for serious research), with a particular collection of packages, some of them in `~/.gap/pkg`.
Then one starts to use Oscar, perhaps without noticing that this causes some new versions of GAP packages to be downloaded behind the scenes.
If these new versions are stored in `~/.gap/pkg` then the effect is that also the installed GAP will use the newer versions whenever this is possible; in particular for packages without binaries, it is probable that newer versions will be preferred and loadable.
In other words, one gets into the mess of "automatic updates", which means that one cannot be sure that yesterday's results are compatible with today's results.

Concerning my personal use of `~/.gap/pkg`:
I store certain development versions of packages there, and use the command line option `-r` for disabling these versions.

PR #565, opened 2020-12-01
Provide Oscar CI workflows for GitHub actions.

rbehrends:
These workflows allow you to run the Oscar CI test suite (entirely or in part) via manually triggered GitHub actions.

Note that the workflows have to be on the master branch for that to work.

See also https://github.com/oscar-system/oscar-ci-action, which contains GitHub actions used by these workflows.

### DIFF
--- a/.github/workflows/PackageCI.yml
+++ b/.github/workflows/PackageCI.yml
@@ -1,0 +1,23 @@
+name: "Oscar CI (package tests)"
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/PackageCI.yml
```diff
@@ -0,0 +1,23 @@
+name: "Oscar CI (package tests)"
```
rfourquet, line 1: Besides the name, this file looks identical to the previous one ("FullCI.yml"), did I miss something?

rbehrends: This is a copy and paste error in `FullCI.yml`. Basically, I've copied the block without adjusting the `test` parameter.

It should be fixed now.

PR #566, opened 2020-12-04
try to make the Julia help available in a GAP session

ThomasBreuer:
For that, notify a virtual GAP help book called `Julia`,
extract the Julia help about the Julia object in question
(including the relevant ANSI escape sequences),
and feed it into the data structures expected by the GAP help system.

For example, the following requests in a GAP session will include entries from the Julia help.
```
?Julia:size
?Julia:GAP.julia_to_gap
?Julia:GAP.julia_help_string
```
One problem is that the Julia help system is *case sensitive*, but the GAP help system turns the given string to lowercase in the beginning (in a call to `SIMPLE_STRING` in GAP's `HELP` function).
Note that Julia does not understand the request `?base.sqrt`, one has to enter `?Base.sqrt`.
I think the solution to this problem would be to change the GAP help system such that the custom handler functions get access to the original search string, either by not changing the string to lowercase (and then adjust the functions that expect lowercase) or by storing the original search string in a global variable or global option; what do you think?

Another problem (or rather a feature) is that help requests to a Julia object work only if the Julia module in question has been loaded. (Note that GAP can load the documentation of a GAP package also when the package itself is not loaded.)

### DIFF
--- a/pkg/JuliaInterface/gap/juliahelp.g
+++ b/pkg/JuliaInterface/gap/juliahelp.g
@@ -1,0 +1,82 @@
+##  Make the Julia help available in the GAP session,
+##  via a (virtual) help book called `Julia`,
+##  which uses a custom handler called `juliahelpformat`.
+##  This works only for the `"text"` help in the terminal,
+##  we cannot access PDF and HTML formats.
+##
+##  The idea is to enter `?Julia:sqrt` or `?Base.isdefined`,
+##  to get one entry for the book `Julia` in the list of choices
+##  if Julia knows documentation for the topic,
+##  and to show *all* matching docstrings from Julia when this entry
+##  is chosen from the menu (since this is the way how Julia behaves).
+
+HELP_ADD_BOOK( "Julia", "dummy entry for accessing Julia documentation",
+    DirectoriesPackageLibrary( "JuliaInterface", "help" )[1] );
+
+HELP_BOOK_HANDLER.juliahelpformat:= rec(
+    # see "The Help Book Handler"
+    ReadSix:= stream -> rec( bookname:= "Julia", entries:= [] ),
+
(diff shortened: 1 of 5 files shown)


fingolfin:
> I think the solution to this problem would be to change the GAP help system such that the custom handler functions get access to the original search string, either by not changing the string to lowercase (and then adjust the functions that expect lowercase) or by storing the original search string in a global variable or global option; what do you think?

I agree that the original string should be made available to help handler. That requires a change to GAP, and we have to be careful to make it so that we don't break existing help handlers.

One option to do this would be to pass both the "simplified" string as well as the original through to `HELP_SHOW_MATCHES` and `HELP_GET_MATCHES`. Then in the help book handler records, allow for an optional member `WantsOriginalSearchString` (or so); if that is bound to `true`, pass the original string to `SearchMatches`, otherwise the simplified one.

ThomasBreuer:
The code can be made safe when pull request gap-system/gap/pull/4212 is available.
Currently my idea is as follows:
- The Julia help is shown only if the help request in GAP has the form `?Julia:<topic>`, where `<topic>` is case sensitive.
That is, `?Julia:size` and `?Julia:Base.size` will work, but entering `?size` will *not* list the entry for Julia's `size`in the list of matches (because the input does not specify the "book" `Julia`), and `?Julia:Size` will *not* show a match (because there is no `Size` in Julia).
- No "search alternatives" (like `centralizer` matching `centraliser`) are supported in help requests for Julia.
- No requests `??Julia:<topic>` are supported. The description of the GAP help system says that `??` triggers a search in the index entries of the book(s), which we do not have for Julia.

ThomasBreuer:
@fingolfin Now a new GAP_jll is available that contains the changes from gap-system/gap/pull/4212.
Can we merge this pull request?

PR #569, opened 2020-12-08
README.md: clarify that GAP.jl is low-level

fingolfin:
There recently was some confusion about this. Point it out explicitly (people still won't read it, but at least we can
now point them at the README ;-)


fingolfin:
Oops, I didn't mean to merge this -- will undo

ThomasBreuer (approved):
The term "This package" is irritating, since the previous paragraph is about Julia and GAP packages in the repository.
Perhaps write "This code" instead?

fingolfin:
@ThomasBreuer sorry, I accidentally merged this PR, wanted to merge #568 instead and got confused in the browser windows *sigh*😔. Anyway, I'll address your remark in a separate PR.

PR #570, opened 2020-12-08
Tweak README.md

fingolfin:
@ThomasBreuer how about this? Feel free to suggest something different and/or propose further edits.


ThomasBreuer (approved):
Thanks.

PR #576, opened 2020-12-08
CI: add Coveralls integration

fingolfin:


fingolfin:
There is an issue: it also processes coverage data for files in `$HOME/.julia`, so e.g. the GAP library; of course we don't want that. This distorts the coverage reported by coveralls quite a bit.

May need to use some tool or awk script to filter out those paths from the `lcov.info` file.

PR #577, opened 2020-12-08
Prefer constructors over gap_to_julia

fingolfin:


ThomasBreuer (approved):
Good.

I think that also the example in `README.md` can be changed, in the sense that
"converting back to Julia is done using `GAP.gap_to_julia` or suitable constructors".

And there is one more occurrence in `src/julia_to_gap.jl`, we can replace `gap_to_julia(Globals.RecNames(obj))` by `Vector{String}(Globals.RecNames(obj))`.

fingolfin:
@ThomasBreuer thanks, good feedback. I adjusted a couple more places, too. Please have another look.

ThomasBreuer (approved):
Thanks.
The new changes deal with the `julia_to_gap` direction.
From this viewpoint, should we perhaps regard `julia_to_gap` as an internal function, and `GapObj` as the recommended constructor?
Then `julia_to_gap` could become undocumented, and the documentation of `GapObj` should be extended.

fingolfin:
I would support the idea of making `julia_to_gap` unofficial. But not in this PR ;-)

PR #584, opened 2020-12-11
Test GAP.Packages.install/remove in a temporary directory

### LINKED ISSUE #574: Don't install&remove GAP "io" package globally as part of the test suite
We currently do this in the test suite:
```julia
    @test GAP.Packages.install("io", interactive = false)
    @test GAP.Packages.remove("io", interactive = false)
```
... which is inconvenient as it means that after running the test suite, the IO package is removed...

Better would be to reactivate the commented out variant coming afterwards:
```
#    pkgdir = mktempdir()
#    @test GAP.Packages.install("io", interactive = false, pkgdir = pkgdir)
#    @test GAP.Packages.remove("io", interactive = false, pkgdir = pkgdir)
```
I think this fails because the installed package can not be actually loaded; so to "fix" this, one probably needs to add the temporary `pkgdir` to the list of directories GAP searches for packages, e.g. via `ExtendRootDirectories`

fingolfin:
Fixes #574

### DIFF
--- a/test/packages.jl
+++ b/test/packages.jl
@@ -4,12 +4,12 @@
     @test ! GAP.Packages.load("no such package")
     @test ! GAP.Packages.load("no such package", install = true)
 
-    @test GAP.Packages.install("io", interactive = false)
-    @test GAP.Packages.remove("io", interactive = false)
-
-#    pkgdir = mktempdir()
-#    @test GAP.Packages.install("io", interactive = false, pkgdir = pkgdir)
-#    @test GAP.Packages.remove("io", interactive = false, pkgdir = pkgdir)
+    tmproot = mktempdir()
+    pkgdir = joinpath(tmproot, "pkg")
+    mkpath(pkgdir)
+    GAP.Globals.ExtendRootDirectories(GapObj([tmproot * "/"]; recursive=true))
+    @test GAP.Packages.install("io", interactive = false, pkgdir = pkgdir)
+    @test GAP.Packages.remove("io", interactive = false, pkgdir = pkgdir)
 
 end
 


fingolfin:
This works locally on my computer (????) but not in the CI tests, and also not on an Ubuntu server I tried.

Looking at it, the IO package is compiled fine (and I separately verified that the `io.so` loadable module is functional. But for some reason, GAP fails to ever "see" this additional copy of the IO package in a temp directory. Instead it insists on wanting to use the copy in `~/.julia/artifacts/848f0f318405266896d24d11e4baf5cf7bf3f203/gap-4.11.0/`. Even if I put the temp dir first in the list of GAPROOTS, this does not change.

So this looks as if it is a problem in the GAP library itself, and/or in PackageManager...

ThomasBreuer:
I have tried the test locally, and `GAP.Packages.install("io", interactive = false, pkgdir = pkgdir)` returns `true` --this seems to be the problem.
In my local setup, version 4.7.1 of `io` is already loaded before the test. Then the above command first downloads and unpacks version 4.7.2;
Afterwards `PKGMAN_CheckPackage` is called, which runs the compilation script.
In the final check, `TestPackageAvailability(info.PackageName, info.Version)` returns `fail` but since `io` is already loaded (here the version number is *not* tested!) the function returns `true`.

With the following input, Julia installs `io` in the temporary directory, and the package can be loaded afterwards.

```
__GAP_ARGS__ = [ "-r", "-A" ]; # make sure that io is not loaded before the test
using GAP
tmproot = mktempdir()
pkgdir = joinpath(tmproot, "pkg")
mkpath(pkgdir)
GAP.Globals.ExtendRootDirectories(GapObj([tmproot * "/"]; recursive=true))
GAP.Packages.install("io", interactive = false, pkgdir = pkgdir) # yields true
GAP.Packages.load("io") # yields true
GAP.Globals.GAPInfo.PackagesLoaded.io # belongs to version 4.7.2, installed in the temp. dir.
```

(Changing the list `GAPInfo.RootPaths` by hand is dangerous, `ExtendRootDirectories` takes care about the caches for `DirectoriesLibrary`.)

PR #594, opened 2020-12-22
CI: remove version restriction for AbstractAlgebra

fingolfin:
This reverts commit 65ee87cbdb79c325cf9b3fe1372ab9696cab3eab.

But it does not (yet) adjust the tests: IMHO the new printing is worse in some cases, and so I wonder if we ought to discuss this with AA/Nemo people. But to do that, one should extract pure Julia versions of those tests, etc. etc. which I don't have time for right now.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -60,7 +60,6 @@ jobs:
         run: |
           julia -e '
             using Pkg
-            Pkg.add(PackageSpec(name="AbstractAlgebra", version="0.10"))
             Pkg.add(["Singular", "Nemo"])
             '
       - name: "GAP tests"


fingolfin:
Hrm, I don't understand why tests pass in Julia 1.3 yet fail in 1.4 and 1.5. I checked and the exact same list of packages is installed in both:
```
[c3fe647b] + AbstractAlgebra v0.11.2
[e21ec000] + Antic_jll v0.2.2+1
[d9960996] + Arb_jll v2.18.1+1
[56f22d72] + Artifacts v1.3.0
[b99e7846] + BinaryProvider v0.5.10
[e66e0078] + CompilerSupportLibraries_jll v0.3.4+0
[1f15a43c] + CxxWrap v0.11.0
[e134572f] + FLINT_jll v2.6.3+1
[e8aa6df9] + GLPK_jll v4.65.0+1
[781609d7] + GMP_jll v6.1.2+6
[692b3bcd] + JLLWrappers v1.1.3
[472f376f] + LoadFlint v0.3.4
[3a97d323] + MPFR_jll v4.0.2+4
[1914dd2f] + MacroTools v0.5.6
[2edaba10] + Nemo v0.18.5
[fb686558] + RandomExtensions v0.4.3
[bcd08a7b] + Singular v0.4.2
[43d676ae] + Singular_jll v4.1.3+6
[f07e07eb] + cddlib_jll v0.94.10+1
[1493ae25] + lib4ti2_jll v1.6.9+2
[3eaa8342] + libcxxwrap_julia_jll v0.8.3+0
[ae4fbd8f] + libsingular_julia_jll v0.2.0+0
```

fingolfin:
@ThomasBreuer can you have a look at the diffs in the tests here? In particular, this one:
```
########> Diff in /Users/runner/work/GAP.jl/GAP.jl/pkg/JuliaExperimental/tst/context.tst:159
# Input is:
vec:= GAPToNemo( c, [ a+1, a-1 ] );
# Expected output:
<<Julia: [(a+1) (a-1)]>>
# But found:
<<Julia: [a + 1 a - 1]>>
########
########> Diff in /Users/runner/work/GAP.jl/GAP.jl/pkg/JuliaExperimental/tst/context.tst:166
# Input is:
mat:= GAPToNemo( c, [ [ a, a+1 ], [ 2*a, a^2+1] ] );
# Expected output:
<<Julia: [a (a+1); 2*a 0]>>
# But found:
<<Julia: [a a + 1; 2*a 0]>>
```

One way to address this (while letting the tests pass with both old and new AA versions) would be to simply suppress the output of those two input lines. Would that be OK?

ThomasBreuer:
@fingolfin Yes, I will make the tests pass.

(I think the former output format was definitely better --it is hard to distinguish `[a^2 -a - 1]` from `[a^2 - a -1]` even if one knows the (current) rules.)

PR #600, opened 2021-01-06
Fix potential GC related crashes

fingolfin:
When we ccall a function to which we pass pointers (in)to Julia objects, if
that function may end up making memory allocations, in general we must use
GC.@preserve on the Julia objects we passed in to ensure they are not garbage
collected. This is not needed if the Julia object is in fact a GAP object (as
our stack scanning logic takes care of those), or if other Julia code later on
is guaranteed to access the object in danger of being GCed.

This could result in a crash in practice when using `evalstr`; the string
containing GAP code could be garbage collected while it was still being
executed.

In the same vein, call `JL_GC_PUSH1(&args)` in the C function `call_gap_func`.

This is unfortunately most likely *not* the same as the GC related crash @fieker reported in December

### DIFF
--- a/pkg/JuliaInterface/src/calls.c
+++ b/pkg/JuliaInterface/src/calls.c
@@ -24,6 +24,7 @@ jl_value_t * call_gap_func(Obj func, jl_value_t * args)
     size_t len = jl_nfields(args);
     Obj    return_value = NULL;
     BEGIN_GAP_SYNC();
+    JL_GC_PUSH1(&args);
     if (IS_FUNC(func) && len <= 6) {
         switch (len) {
         case 0:
@@ -73,6 +74,7 @@ jl_value_t * call_gap_func(Obj func, jl_value_t * args)
         }
         return_value = CallFuncList(func, arg_list);
     }
+    JL_GC_POP();
     END_GAP_SYNC();
     if (return_value == NULL) {
         return jl_nothing;
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -81,7 +81,7 @@ function initialize(argv::Array{String,1})
         # Tell GAP to show some traceback on errors.
         append!(argv, ["--alwaystrace"])
     end
-    ccall(
+    GC.@preserve argv ccall(
         (:GAP_Initialize, libgap),
         Cvoid,
         (Int32, Ptr{Ptr{UInt8}}, Ptr{Cvoid}, Ptr{Cvoid}, Cuint),
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -39,8 +39,7 @@ end
 
 
 function evalstr_ex(cmd::String)
-    res = ccall((:GAP_EvalString, libgap), GapObj, (Ptr{UInt8},), cmd)
-    return res
+    return GC.@preserve cmd ccall((:GAP_EvalString, libgap), GapObj, (Ptr{UInt8},), cmd)
 end
 
 """
@@ -79,7 +78,7 @@ end
 # Retrieve the value of a global GAP variable given its name. This function
 # returns a raw Ptr value, and should only be called by plumbing code.
 function _ValueGlobalVariable(name::Union{AbstractString,Symbol})
-    return ccall((:GAP_ValueGlobalVariable, libgap), Ptr{Cvoid}, (Ptr{UInt8},), name)
+    return GC.@preserve name ccall((:GAP_ValueGlobalVariable, libgap), Ptr{Cvoid}, (Ptr{UInt8},), name)
 end
 
 function ValueGlobalVariable(name::Union{AbstractString,Symbol})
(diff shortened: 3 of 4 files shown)


rbehrends (approved):
Looks good. While the overhead associated with `JL_GC_PUSH1()` and `JL_GC_POP()` is non-trivial (they have to call `jl_get_ptls_states()`), it should be negligible in the context of calling a GAP function, and is unavoidable, anyway.

rfourquet:
Just a little pedantic remark :)

> This is not needed [...] if other Julia code later on is guaranteed to access the object in danger of being GCed.

I'm no expert at all on this, but Yu Yichao has warned occasionally against such interpretations of when `GC.@preserve` is needed; while this is generally true in practice and might remain so for the foreseable future, they advise to stick to what is said in the docstring, which doesn't mention the case where "Julia code later on accesses the object" or where "that function may end up making memory allocations", given in particular that 

> `@preserve` should generally not have any performance impact in typical use cases [...]

fingolfin:
@rfourquet thanks for the remark! What I don't understand, though: with that strict interpretation, then one should/must use `GC.@preserve` *always* when a Julia object is passed to a `ccall`? Julia itself doesn't seem to do that (and one wonders: if it is indeed free, then why doesn't it do it "automatically" in those cases?)

fingolfin:
BTW, I wrote "memory allocation" but had in mind "anything potentially triggering a GC", I'll fix that

rfourquet:
> then one should/must use GC.@preserve always when a Julia object is passed to a `ccall`?

Only when the `ccall` uses memory (or other resources) of an object without Julia being able to see it. E.g. an array can be passed in lieu of a pointer without `@preserve`, because `ccall` knows about the array, and converts via `unsafe_convert`/`cconvert`. But when passing directly to `ccall` a pointer to a Julia array, `ccall` doesn't know the connection between the pointer it received and the source array, which then needs to be "preserved".

ThomasBreuer:
To me, it looks strange that using `ccall` requires `GC.@preserve`, in particular the documentation of `ccall` at https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/ does not mention this.
(I have found the discussion at https://discourse.julialang.org/t/on-the-garbage-collection/35695, is there a better reference?)

fingolfin:
@rfourquet that does not match what we saw: we passed a Julia string into a function causing a GC and it crashed eventually; that triggered the creation of this PR. This was the problematic code:
```
function evalstr_ex(cmd::String)
     res = ccall((:GAP_EvalString, libgap), GapObj, (Ptr{UInt8},), cmd)
     return res
end
```
Here `cmd` got GCed prematurely.

rfourquet:
> we passed a Julia string into a function causing a GC and it crashed eventually

I don't understand why, this gets out of my depth :shrug: 

> Here cmd got GCed prematurely.

I would consider this a bug in Julia.

> it looks strange that using ccall requires GC.@preserve

It took me some time to understand the concept of `GC.@preserve` (it was using `guile` scheme, where there is a similar concept called `remember_upto_here`) .
The typical example is passing an array to a C function:
```julia
v = [1, 2, 3]
p = pointer(v)
ccall(..., (Ptr{Int},), p) # or even regular julia code, like `x = unsafe_load(p)`
```
As far as Julia can see, `v` is not needed -- and therefore could be GC'ed -- between the second and 3rd lines, e.g. right before the `ccall`, in which case `ccall` will presumably access invalid memory. Yichao then says (in discussion you linked to) that we shouldn't even get concerned about the GC:

> the GC has little to do with the need to preserve an object (well, or only partially)

Also, using `GC.@preserve v expr` only preserves `v` within `expr`, not before, and using `v` after the unsafe code doesn't mean that `@preserve` can be omitted.

A good tldr from Yichao:

> [`@preserve`] is needed whenever you are using manipulating what belong to an object without using an reference to that object.

fingolfin:
Interesting, https://docs.julialang.org/en/v1.5/base/base/#Base.GC.@preserve has been substantially extended compared to v1.4 (which apparently was the last time I read this part of the docs).

Of course it'd be nice if we didn't need those `GC.@preserve` calls (we then also get rid of the new `JL_GC_PUSH1()` and `JL_GC_POP()` this PR inserts).

So perhaps it is indeed a bug in Julia -- we have a 100% reliable reproducer, after all, which went away with that `GC.@preserve` inserted (in Julia 1.5.3).

fingolfin:
The docs are still terribly vague, though: in the example where they claim no `GC.@preserve` is needed, the Julia string is passed in as a `Cstring` while we use `Ptr{UInt8}`. Maybe that's a relevant difference?

I tried the repro code by @rbehrends at https://gist.github.com/rbehrends/be0e03df33dbe88bea083613941eee6b and I can still repro a crash with it on GAP.jl master. But, interestingly, when I replace the `Ptr{Uint8}` by `Cstring`, it does seem to fix the issue as well.

So perhaps not a bug in Julia after all, just a case of really fuzzy documentation? Hmmm

Issue #580, opened 2020-12-09
problem with handling of GAP errors


ThomasBreuer:
The following error message is fine.
```julia
julia> using GAP
[...]

julia> GAP.evalstr( "1/0" )
ERROR: Error thrown by GAP: Error, Rational operations: <divisor> must not be zero
[...]
```

The following behaviour is not what one would expect.
```julia
julia> using GAP
[...]

julia> GAP.evalstr( "\"" )

julia> GAP.evalstr( "1/0" )
ERROR: Error thrown by GAP: Syntax error: String must end with " before end of file in stream:1

Error, Rational operations: <divisor> must not be zero
[...]
```
There is no error message after the first bad input.
But when the second bad input gets evaluated, one gets the message that belongs to the first call, together with the error message for the second one.

ThomasBreuer:
I guess the problem is related to the one described in issue #516.

ThomasBreuer:
done via #605

Issue #596, opened 2020-12-23
Calling GapObj on Tuples yields Arrays


SyxP:
See attached code.

```
julia> GAP.evalstr("(1,2,3)")
GAP: (1,2,3)

julia> GapObj((1,2,3))
GAP: [ 1, 2, 3 ]
```

fingolfin:
There are no tuples in GAP.

Your first input returns a GAP permutation.

Your second input returns a GAP list (roughly corresponds to a Julia Vector/Array).

fingolfin:
I don't see an issue here, so I am going to close this. But feel free to provide additional information, then we can reopen.

SyxP:
Ah the difficulty I faced was that when I tried to replicate the GAP code `Group((1,2,3))`, I tried `GAP.Globals.Group(GapObj((1,2,3)))` which didn't work, but required `GAP.Globals.Group(Gap.evalstr("(1,2,3)")`, perhaps not a bug but perhaps should be at least noted somewhere. I think this is likely the first things people might try.

Issue #516, opened 2020-09-07
GAP.Globals.Display once a GAP error has occured


zickgraf:
Consider the following code and output:
```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.1 (2020-08-25)
 _/ |\__'_|_|_|\__'_|  |  
|__/                   |

julia> using GAP
 ┌───────┐   GAP 4.11.0 of 29-Feb-2020
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-julia64-kv7-v1.5
 Configuration:  gmp 6.2.0, Julia GC, Julia 1.5.1, readline
 Loading the library and packages ...
 Packages:   GAPDoc 1.dev, PrimGrp 3.4.0, SmallGrp 1.4.1, TransGrp 2.0.5
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'

julia> GAP.Globals.Display( g"test1" )
test1

julia> GAP.Display( g"test2" )
test2

julia> GAP.evalstr( "error" )
ERROR: Error thrown by GAP: Error, Variable: 'error' must have a value

Stacktrace:
 [1] error_handler() at .julia/packages/GAP/tkivN/src/GAP.jl:157
 [2] evalstr_ex at .julia/packages/GAP/tkivN/src/ccalls.jl:36 [inlined]
 [3] evalstr(::String) at .julia/packages/GAP/tkivN/src/ccalls.jl:63
 [4] top-level scope at REPL[4]:1
 [5] run_repl(::REPL.AbstractREPL, ::Any) at /build/julia/src/julia-1.5.1/usr/share/julia/stdlib/v1.5/REPL/src/REPL.jl:288

julia> GAP.Globals.Display( g"test3" )

julia> GAP.Display( g"test4" )
test4

julia>

```
The string `test3` is not printed. In fact, nothing seems to be printed using `GAP.Globals.Display` after an error has occured. `GAP.Display` seems to work fine.

A related question we are interested in: Which stream is `Display` outputting to? We would like to turn the print formatting off for this stream as we did in #513.

ThomasBreuer:
@zickgraf Thanks for this report.

[...]

PR #605, opened 2021-01-06
catch GAP errors in `GAP.evalstr`

ThomasBreuer:
The right way to do this is to check the return value of `evalstr_ex`,
and to call `GAP.error_handler` if necessary,
as was proposed in the discussion of gap-system/gap/issues/4199.

This is intended to resolve issue #580.

(The change does of course *not* affect the problems described in issue #596.
It is also not related to issue #516.)

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -47,7 +47,8 @@ end
     evalstr(cmd::String)
 
 Let GAP execute the command(s) given by `cmd`;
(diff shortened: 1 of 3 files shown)


fingolfin (approved):

ThomasBreuer:
(The test failures are due to changed output formats in Julia nightly.)

PR #607, opened 2021-01-08
added a description how doctestfilters work

ThomasBreuer:
This should better have been part of PR #604, but this is already merged.

### DIFF
--- a/src/doctestfilters.jl
+++ b/src/doctestfilters.jl
@@ -1,11 +1,18 @@
 # Define some replacements which are applied in `docs/make.jl` and `test/testmanual.jl`.
+# The idea is that the `jldoctest`ed examples in the source files
+# have the format that fits to the "current" Julia,
+# and that other Julia versions (such as Julia nightly) may show different formats.
+# Each regular expression in `GAP_doctestfilters` gets applied to both the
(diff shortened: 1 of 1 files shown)


src/doctestfilters.jl
```diff
@@ -1,11 +1,19 @@
 # Define some replacements which are applied in `docs/make.jl` and `test/testmanual.jl`.
+# The idea is that the `jldoctest`ed examples in the source files
+# have the format that fits to the "current" Julia,
+# and that newer Julia versions (Julia nightly) may show a different format.
```
rfourquet, line 4: Maybe also "older" versions?
```suggestion
# and that older or newer Julia versions (Julia nightly) may show a different format.
```

src/doctestfilters.jl
```diff
@@ -1,11 +1,19 @@
...
+# Each replacement in `GAP_doctestfilters` gets applied to both the
```
rfourquet, line 5: A regex in itself is not a replacement, so maybe clarify that each regex in `GAP_doctestfilters` defines implicitly a replacement: text matching any of the regexes is replaced by the empty string.

ThomasBreuer: Thanks. Yes, it is better to say what actually happens.

PR #617, opened 2021-02-11
improvements to the JuliaInterface manual

ThomasBreuer:
... such that we get a meaningful version for the GAP Days.

The details:

- extended `PackageInfo.g`:
  - added myself to the involved persons,
  - added `Keywords` entry,
  - added `AutoDoc` component,
    with the components `Abstract` and `Acknowledgements`,
- added the introductory chapter (aims, installation),
- clarified/corrected the description of `ImportJuliaModuleIntoGAP`,
- clarified the description of the different kinds of function calls,
- explained the defaults of `GAPToJulia` and `JuliaToGAP` w.r.t. recursion,
- removed those open items that have meanwhile been dealt with,
- added a description of the user preference `IncludeJuliaStartupFile`,
- added a few examples.

Now there are no visible `TODO` markers anymore, but I have still some questions:

- The chunk `IncludeJuliaStartupFile` is requested in `gap/JuliaInterface.gd`, *behind* the other two sections of the first chapter, but it appears as the *first* section.
  Am I misusing AutoDoc?
- There had been an open item "Should we restrict the automatic FFE conversion to `IsInternalRep`?"
  However, in fact we do not provide a conversion of non-internal FFEs at all.
  For example, `GAPToJulia( Z(101, 10) )` yields an error.
- Currently the cross-references to `MatElm` and `SetMatElm` in the GAP Reference Manual do not work, because GAP 4.11.0 provides these operations but no documentation for them.
  Can we expect this documentation in GAP soon, or shall we reformulate the cross-references?
- Also cross-references to the GAP `dev` manual will not work if users do not have this manual installed (which is usually the case).
- The `AutoDoc` manual contains a section on copyright and license; shall we add this also for `JuliaInterface`? (If yes with which copyright?)

### DIFF
--- a/pkg/JuliaInterface/PackageInfo.g
+++ b/pkg/JuliaInterface/PackageInfo.g
@@ -16,6 +16,24 @@ License := "GPL-2.0-or-later",
 
 Persons := [
   rec(
+    LastName := "Breuer",
+    FirstNames := "Thomas",
+    IsAuthor := true,
+    IsMaintainer := true,
+    Email := "sam@math.rwth-aachen.de",
+    WWWHome := "http://www.math.rwth-aachen.de/~Thomas.Breuer",
+    PostalAddress := Concatenation( [
+                       "Thomas Breuer\n",
+                       "Lehrstuhl für Algebra und Zahlentheorie\n",
+                       "RWTH Aachen\n",
+                       "Pontdriesch 14/16\n",
+                       "52062 Aachen\n",
+                       "Germany"
+      ] ),
+    Place := "Aachen",
+    Institution := "RWTH Aachen",
+  ),
+  rec(
     IsAuthor := true,
     IsMaintainer := true,
     FirstNames := "Sebastian",
@@ -117,7 +135,22 @@ end,
 
 TestFile := "tst/testall.g",
 
-#Keywords := [ "TODO" ],
+Keywords := [ "GAP-Julia integration" ],
 
+AutoDoc := rec(
+  TitlePage := rec(
+    Abstract := Concatenation(
+      "The &GAP; package <Package>JuliaInterface</Package> is part of ",
+      "a bidirectional interface between &GAP; and &Julia;.\n"
+    ),
+    Acknowledgements := Concatenation(
+      "The development of this &GAP; package has been supported ",
+      "by the <URL><Link>https://www.computeralgebra.de/sfb/</Link>",
+      "<LinkText>SFB-TRR 195 ",
+      "<Q>Symbolic Tools in Mathematics and their Applications</Q>",
+      "</LinkText></URL> (from 2017 until 2020).\n",
+      "<P/>\n"
+    ),
+  ),
+),
 ));
-
--- a/pkg/JuliaInterface/PackageInfo.g
+++ b/pkg/JuliaInterface/PackageInfo.g
@@ -148,7 +148,7 @@ AutoDoc := rec(
       "by the <URL><Link>https://www.computeralgebra.de/sfb/</Link>",
       "<LinkText>SFB-TRR 195 ",
       "<Q>Symbolic Tools in Mathematics and their Applications</Q>",
-      "</LinkText></URL> (from 2017 until 2020).\n",
+      "</LinkText></URL> (from 2017 until 2021).\n",
       "<P/>\n"
     ),
   ),
--- a/pkg/JuliaInterface/PackageInfo.g
+++ b/pkg/JuliaInterface/PackageInfo.g
@@ -35,7 +35,7 @@ Persons := [
   ),
   rec(
     IsAuthor := true,
-    IsMaintainer := true,
+    IsMaintainer := false,
     FirstNames := "Sebastian",
     LastName := "Gutsche",
     WWWHome := "https://sebasguts.github.io",
--- a/pkg/JuliaInterface/gap/JuliaInterface.gd
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gd
@@ -533,8 +533,10 @@ DeclareGlobalFunction( "CallJuliaFunctionWithKeywordArguments" );
 #!    access to and assignment of entries of arrays, via
 #!    <Ref Oper="\[\]" BookName="ref"/>,
 #!    <Ref Oper="\[\]\:\=" BookName="ref"/>,
-#!    <Ref Oper="MatElm" BookName="ref"/>, and
-#!    <Ref Oper="SetMatElm" BookName="ref"/>,
+#!    <!-- <Ref Oper="MatElm" BookName="ref"/>, and
+#!    <Ref Oper="SetMatElm" BookName="ref"/>, -->
+#!    and the (up to &GAP; 4.11 undocumented) operations <C>MatElm</C> and
+#!    <C>SetMatElm</C>,
 #!    delegating to
 #!    <C>Julia.Base.getindex</C> and
 #!    <C>Julia.Base.setindex</C>,
@@ -593,5 +595,6 @@ DeclareGlobalFunction( "CallJuliaFunctionWithKeywordArguments" );
 #! gap> m + m;
 #! <Julia: [2 4; 6 8]>
 #! @EndExampleSession
+#TODO: add the cross-references to MatElm, SetMatElm when they are documented
 
 #! @InsertChunk JuliaHelpInGAP
--- a/pkg/JuliaInterface/gap/JuliaInterface.gd
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gd
@@ -17,7 +17,7 @@
 #!  in a &Julia; session,
 #!  to access &Julia; objects and to call &Julia; functions
 #!  in a &GAP; session,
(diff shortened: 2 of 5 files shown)


fingolfin:
Thanks @ThomasBreuer 

> Now there are no visible `TODO` markers anymore, but I have still some questions:

OK...
 
> * The chunk `IncludeJuliaStartupFile` is requested in `gap/JuliaInterface.gd`, _behind_ the other two sections of the first chapter, but it appears as the _first_ section.
>   Am I misusing AutoDoc?

I will have to take a look to see what's going on

> * There had been an open item "Should we restrict the automatic FFE conversion to `IsInternalRep`?"
>   However, in fact we do not provide a conversion of non-internal FFEs at all.
>   For example, `GAPToJulia( Z(101, 10) )` yields an error.

What is the question?

> * Currently the cross-references to `MatElm` and `SetMatElm` in the GAP Reference Manual do not work, because GAP 4.11.0 provides these operations but no documentation for them.
>   Can we expect this documentation in GAP soon, or shall we reformulate the cross-references?

They have documentation in GAP 4.12, but not in 4.11; I don't think there are plans to backport this. However, I hope we'll release GAP 4.12 soon (perhaps next week, but more likely we'll do 4.11.1 and then 4.12.0 in a couple weeks)

For now, we could comment out our refs and add a TODO to add them back once they are in GAP?

> * Also cross-references to the GAP `dev` manual will not work if users do not have this manual installed (which is usually the case).

I wasn't even aware we had such a ref, but now I see it. But we have it for immediate ints and FFEs... So perhaps for now comment it out? For small ints, we could also reference the start of the `Integers` chapter in the GAP ref manual, which discusses this briefly. Anyway it's a low-level technical details. Perhaps future GAP versions will have a public dev manual again (but then it probably should be revised first?)

> * The `AutoDoc` manual contains a section on copyright and license; shall we add this also for `JuliaInterface`? (If yes with which copyright?)

I'd punt this for now, perhaps open an issue so we can discuss this in the future (e.g. next week).

pkg/JuliaInterface/PackageInfo.g
```diff
@@ -117,7 +135,22 @@ end,
...
+    Abstract := Concatenation(
+      "The &GAP; package <Package>JuliaInterface</Package> is part of ",
+      "a bidirectional interface between &GAP; and &Julia;.\n"
+    ),
+    Acknowledgements := Concatenation(
+      "The development of this &GAP; package has been supported ",
+      "by the <URL><Link>https://www.computeralgebra.de/sfb/</Link>",
+      "<LinkText>SFB-TRR 195 ",
+      "<Q>Symbolic Tools in Mathematics and their Applications</Q>",
+      "</LinkText></URL> (from 2017 until 2020).\n",
```
fingolfin, line 151:
```suggestion
      "</LinkText></URL> (from 2017 until 2021).\n",
```

pkg/JuliaInterface/init.g
```diff
@@ -4,22 +4,46 @@
...
-##  <F>~/.julia/config/startup.jl</F>
-##  gets imported into &Julia; after startup.
-##  If the value is a nonempty string that is the name of a directory
-##  then the <F>startup.jl</F> file in this directory gets imported into
-##  &Julia;.
-##  Otherwise no <F>startup.jl</F> file will be imported automatically,
-##  since apparently <C>libjulia</C> does not import such a file.
+#! @BeginChunk IncludeJuliaStartupFile
+#! @Chapter Introduction to <Package>JuliaInterface</Package>
+#! @Section User preferences in the <Package>JuliaInterface</Package> package
```
fingolfin, line 9: Try removing these lines to see if it helps with the chunk issue.

If not, let's merge this PR anyway, we can figure this out later (perhaps log an issue so we don't forget)
```suggestion
```

fingolfin (approved):

ThomasBreuer:
@fingolfin Concerning the comment

> > * There had been an open item "Should we restrict the automatic FFE conversion to `IsInternalRep`?"
> >   However, in fact we do not provide a conversion of non-internal FFEs at all.
> >   For example, `GAPToJulia( Z(101, 10) )` yields an error.
> 
> What is the question?

the situation is that we provide neither an automatic nor a non-automatic conversion for non-internal FFEs.
Thus I was irritated by the question from the open item. I have removed the open item, if this is o.k. then there is no question left here.

pkg/JuliaInterface/PackageInfo.g
ThomasBreuer: Yes.

pkg/JuliaInterface/init.g
ThomasBreuer: Yes, this helps.
(Strange, I remember that I got error messages because of the missing "context" in similar situations.)

fingolfin:
> the situation is that we provide neither an automatic nor a non-automatic conversion for non-internal FFEs.
> Thus I was irritated by the question from the open item. I have removed the open item, if this is o.k. then there is no question left here.

it is OK :-)

pkg/JuliaInterface/init.g
```diff
@@ -4,22 +4,44 @@
...
-##  since apparently <C>libjulia</C> does not import such a file.
+#! @BeginChunk IncludeJuliaStartupFile
+#!  <Subsection Label="subsect:IncludeJuliaStartupFile">
+#!  <Heading>User preference <C>IncludeJuliaStartupFile</C></Heading>
+#!  <Index Key="IncludeJuliaStartupFile"><C>IncludeJuliaStartupFile</C></Index>
+#!
+#!  When one starts an interactive &Julia; session,
+#!  the &Julia; startup file <F>~/.julia/config/startup.jl</F> gets
+#!  included automatically by default,
+#!  see the section <Q>The Julia REPL</Q> in the &Julia; documentation.
```
fingolfin, line 15: Perhaps this could be turned into an actual link to https://docs.julialang.org/en/v1/stdlib/REPL/?

ThomasBreuer: Yes.

fingolfin:
I just realized that we never really build the documentation for `JuliaInterface`, do we? We probably should (HTML+TXT only, no PDF, so that user's don't need LaTeX). But how?? We have to be careful to do it only after GAP actually runs, so we can't do it during precompilation as part of `regenerate_gaproot()` (I think).

Next best I can think of is to *delete* any generated documentation files in `regenerate_gaproot()`, and then regenerate them in `__init__` if we notice they are missing (this way, we always regenerate them if needed).

Another solution would be to be slightly nasty and commit the relevant files (*.html, *.txt, *.js, *.css, ...) into the repository... 

Or if JuliaInterface was built into a JLL (see issue #619), then that JLL could contain the generated documentation, too.

fingolfin:
I guess my last comment should be turned into an issue of its own

pkg/JuliaInterface/PackageInfo.g
```diff
@@ -15,6 +15,24 @@ Date := "07/01/2021", # dd/mm/yyyy format
...
+                       "Pontdriesch 14/16\n",
+                       "52062 Aachen\n",
+                       "Germany"
+      ] ),
+    Place := "Aachen",
+    Institution := "RWTH Aachen",
+  ),
   rec(
     IsAuthor := true,
     IsMaintainer := true,
```
mohamed-barakat, line 38:
```suggestion
    IsMaintainer := false,
```

ThomasBreuer: @mohamed-barakat Perhaps I am misunderstanding what "maintainer" means; I thought that an author is somebody who has contributed to the package, perhaps some time ago, and that a maintainer is somebody who is still working on it.

fingolfin: And Sebastian is no longer working on it

ThomasBreuer: Ah, now I understand. Sorry for the confusion.

pkg/JuliaInterface/gap/JuliaInterface.gd
```diff
@@ -11,7 +11,86 @@
...
 #!  a bidirectional interface between &GAP; and &Julia;.
 #!
-#!  TODO: state the aims, describe the installation
+#! @Section Aims of the <Package>JuliaInterface</Package> package
+#!  The low level interface between &GAP; and &Julia; allows one
+#!  to access &GAP; objects and to call &GAP; functions
+#!  in a &Julia; session,
+#!  to access &Julia; objects and to call &Julia; functions
+#!  in a &GAP; session,
+#!  and to convert low level data such as integers, Booleans, strings,
```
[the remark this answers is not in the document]
ThomasBreuer: Hm, there are several occurrences of "Boolean".
I think the main point is that one of the two variants is used consistently.

(Writing "Sylow p-subgroup" with capital first letter seems to be used throughout the literature, "abelian group" with small first letter is used in the majority of cases I know, and for "Boolean" vs. "boolean" I do not have a feeling which variant should be preferred.
See for example https://english.stackexchange.com/questions/4481/should-the-word-boolean-be-capitalized .)

PR #620, opened 2021-02-12
Remove references to ForeignGAP

fingolfin:
These are now gone in the latest master branch. Instead we have `GAP_jll.MPtr` and `GAP.GapObj` is an alias for that (this is slightly annoying, actually, it'd be great if we could "rename" it, but that's for another day to worry about.

Perhaps instead of `GAP_jll.MPtr` this PR should use its synonym `GAP.GapObj`. However, Julia prints that as `GAP_jll.MPtr`, so I think for the user `GAP_jll.MPtr` might be better?

```
julia> GapObj
GAP_jll.MPtr

```

### DIFF
--- a/pkg/JuliaInterface/example/function_perform.jl
+++ b/pkg/JuliaInterface/example/function_perform.jl
@@ -1,6 +1,6 @@
 module GapFunctionPerform
 
-import Main.ForeignGAP: MPtr
+import GAP_jll: MPtr
 
 function typed_func(a::MPtr, b::MPtr)
     return a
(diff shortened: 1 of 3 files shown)


ThomasBreuer (approved):
Converning `GapObj` vs. `GAP_jll.MPtr`:
We use `GapObj` in several places in the documentation of GAP.jl, and advertise `GapObj` as a constructor for conversions to GAP.
I think `GapObj` looks nicer, but users will anyhow have to know about `GAP_jll.MPtr` since this is what Julia prints.

fingolfin:
Perhaps then we should use `GapObj`, *but* add something somewhere which explains the relation? But I am in a meeting now. In any case, this PR makes things better, albeit not perfect...

Perhaps you can merge this PR and then (if you feel like it / have time) do such a change?

ThomasBreuer:
@fingolfin the suggested change is in pull request #622.

PR #630, opened 2021-03-03
Don't unwrap IS_JULIA_FUNC objects automatically

### LINKED ISSUE #629: How to get on the Julia side a GAP object wrapping a Julia function?
I ran into this question in the context of issue oscar-system/Oscar.jl/issues/108:

In a GAP session, I can fetch the Julia function `sqrt` via `JuliaFunction( "sqrt" )`; in fact, this yields a GAP object in the filter `IsFunction`.
In a Julia session, this object would be interesting in situations where a GAP operation gets called which expects one of its arguments to have the filter `IsFunction`. However, `GAP.Globals.JuliaFunction(g"sqrt")` yields not the wrapped GAP object but the Julia function itself, which is not in GAP's `IsFunction`.

How can I construct the `IsFunction` object on the Julia side?

(The above description should become part of the documentation for GAP.jl and JuliaInterface.)

fingolfin:
... when moving them to the Julia side.

This way, Julia code can store them and pass them later on to GAP code again,
e.g. as action function for G-set related computations.

This also fits better into our philosophy regarding conversions: the principle
of only doing necessary or "obvious" AND cheap conversions. But this automatic
unwrapping was neither necessary (as long as we provide an explicit unwrap
function on the Julia side), nor necessarily cheap (as discarding the wrapper
meant that one may have to recreate many such wrappers). Moreover, it impeded
the usefulness of these wrappers.

Resolves #629
```
julia> GAP.Globals.JuliaFunction(g"sqrt")
GAP: <Julia: sqrt>
```

TODO:
- add a Julia function for unwrapping a wrapped Julia function
  -  the hardest problem here is to determine what to call it :-) so suggestions welcome
- add a Julia function for testing if a given GAP object is a Julia function (just calling `GAP.Globals.IS_JULIA_FUNC` should be sufficient unless speed is of utmost concern)
- add tests to cover this
- possibly update any relevant documentation, if there is any?

### DIFF
--- a/pkg/JuliaInterface/gap/JuliaInterface.gi
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gi
@@ -191,8 +191,13 @@ end );
 
 
 InstallGlobalFunction( JuliaTypeInfo,
-    juliaobj -> JuliaToGAP( IsString,
-                    Julia.Base.string( Julia.Core.typeof( juliaobj ) ) ) );
+    function( juliaobj )
+    if IsFunction( juliaobj ) then
+      juliaobj:= Julia.GAP.UnwrapJuliaFunc( juliaobj );
+    fi;
+    return JuliaToGAP( IsString,
+                       Julia.Base.string( Julia.Core.typeof( juliaobj ) ) );
+end );
 
 
 InstallGlobalFunction( JuliaModule,
--- a/pkg/JuliaInterface/gap/calls.gi
+++ b/pkg/JuliaInterface/gap/calls.gi
@@ -26,6 +26,9 @@ InstallGlobalFunction( CallJuliaFunctionWithCatch,
     local res;
 
     args := GAPToJulia( _JL_Vector_Any, args, false );
+    if IsFunction( julia_obj ) then
+      julia_obj:= Julia.GAP.UnwrapJuliaFunc( julia_obj );
+    fi;
     res:= Julia.GAP.call_with_catch( julia_obj, args );
     if res[1] then
       return rec( ok:= true, value:= res[2] );
--- a/pkg/JuliaInterface/src/convert.c
+++ b/pkg/JuliaInterface/src/convert.c
@@ -19,9 +19,6 @@ jl_value_t * julia_gap(Obj obj)
     if (IS_JULIA_OBJ(obj)) {
         return GET_JULIA_OBJ(obj);
     }
-    if (IS_JULIA_FUNC(obj)) {
-        return GET_JULIA_FUNC(obj);
-    }
     if (obj == True) {
         return jl_true;
     }
--- a/pkg/JuliaInterface/tst/convert.tst
+++ b/pkg/JuliaInterface/tst/convert.tst
@@ -183,7 +183,8 @@ gap> JuliaToGAP( IsString, emptystring );
 ""
 
 ##  'GAPToJulia' for Julia functions (inside arrays)
-gap> parse:= JuliaFunction( "parse", "Base" );;
+gap> parse:= JuliaFunction( "parse", "Base" );
+<Julia: parse>
 gap> list:= GAPToJulia( JuliaEvalString( "Array{Any,1}"), [ 1, parse, 3 ] );
 <Julia: Any[1, parse, 3]>
 gap> list2:= JuliaToGAP( IsList, list );
--- a/pkg/JuliaInterface/tst/wrapper.tst
+++ b/pkg/JuliaInterface/tst/wrapper.tst
@@ -18,7 +18,7 @@ Error, MethodError: objects of type BigInt are not callable
 ## wrap a Julia function
 gap> f := Objectify(type, rec());;
 gap> SetJuliaPointer(f, Julia.Base.sqrt);
-gap> Julia.Base.typeof(f);
+gap> Julia.Base.typeof(Julia.GAP.UnwrapJuliaFunc(f));
 <Julia: typeof(sqrt)>
 gap> f(4);
 <Julia: 2.0>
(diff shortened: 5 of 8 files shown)


ThomasBreuer (approved):
Thanks. Yes, I think this is better now.
(After some experiments, I noticed that `GAP.julia_to_gap(sqrt)` returned the wrapped GAP function already before the change, so I could have used that. Sorry for missing this.)

Now one could argue that on the GAP side, `Julia.sqrt` could be a not wrapped Julia function (but of course callable on the GAP side; `JuliaToGAP( IsFunction, Julia.sqrt )` could then do the conversion/wrapping), but this is perhaps a bad idea.

fingolfin:
> After some experiments, I noticed that GAP.julia_to_gap(sqrt) returned the wrapped GAP function already before the change, so I could have used that. Sorry for missing this.

No, don't be sorry, I genuinely think this change is the way to go anyway

fingolfin:
This crashes and I do not yet understand why... Below a relevant excerpt of the crash backtrace. It dies in the garbage collector, but note that also the Julia type inference code shows up substantially in the full log. Both may be red herrings, too...
```
    1314 ms (181 ms GC) and 140KB allocated for import.tst
testing: .../GAP.jl/pkg/JuliaInterface/tst/juliainterface01.tst
# line 140 of 171 (81%)
signal (11): Segmentation fault: 11
in expression starting at .../GAP.jl/gaproot/v1.5/bin/gap.sh:17
gc_mark_loop at /Users/julia/buildbot/worker/package_macos64/build/src/gc.c:2397
_jl_gc_collect at /Users/julia/buildbot/worker/package_macos64/build/src/gc.c:2902
jl_gc_collect at /Users/julia/buildbot/worker/package_macos64/build/src/gc.c:3108
maybe_collect at /Users/julia/buildbot/worker/package_macos64/build/src/gc.c:827 [inlined]
jl_gc_pool_alloc at /Users/julia/buildbot/worker/package_macos64/build/src/gc.c:1142
jl_gc_alloc_ at /Users/julia/buildbot/worker/package_macos64/build/src/./julia_internal.h:277 [inlined]
jl_gc_alloc at /Users/julia/buildbot/worker/package_macos64/build/src/gc.c:3150
jl_alloc_svec_uninit at /Users/julia/buildbot/worker/package_macos64/build/src/simplevector.c:60
jl_alloc_svec at /Users/julia/buildbot/worker/package_macos64/build/src/simplevector.c:69
save_env at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:145
forall_exists_subtype at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:1442
subtype_in_env_ at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:1898 [inlined]
subtype_in_env at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:1903 [inlined]
intersect_invariant at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:2811
intersect at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:3057
intersect_tuple at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:2705 [inlined]
intersect at /Users/julia/buildbot/worker/package_macos64/build/src/subtype.c:3021
[...]
typeinf at ./compiler/typeinfer.jl:12
typeinf_ext at ./compiler/typeinfer.jl:570
typeinf_ext at ./compiler/typeinfer.jl:601
jfptr_typeinf_ext_21449.clone_1 at .../lib/julia/sys.dylib (unknown line)
jl_apply at /Users/julia/buildbot/worker/package_macos64/build/src/./julia.h:1690 [inlined]
jl_type_infer at /Users/julia/buildbot/worker/package_macos64/build/src/gf.c:296
jl_generate_fptr at /Users/julia/buildbot/worker/package_macos64/build/src/jitlayers.cpp:290
jl_compile_method_internal at /Users/julia/buildbot/worker/package_macos64/build/src/gf.c:1964
_jl_invoke at /Users/julia/buildbot/worker/package_macos64/build/src/gf.c:2224 [inlined]
jl_apply_generic at /Users/julia/buildbot/worker/package_macos64/build/src/gf.c:2398
jl_apply at /Users/julia/buildbot/worker/package_macos64/build/src/./julia.h:1690 [inlined]
jl_call2 at /Users/julia/buildbot/worker/package_macos64/build/src/jlapi.c:247
DoCallJuliaFunc2Arg at .../GAP.jl/pkg/JuliaInterface/bin/x86_64-apple-darwin14-julia64-kv7/JuliaInterface.so (unknown line)
IntrFuncCallEnd at /Users/mhorn/.julia/artifacts/0703db94dbc1f41724968540c81a95cfa19e701b/lib/libgap.0.dylib (unknown line)
EvalRef at /Users/mhorn/.julia/artifacts/0703db94dbc1f41724968540c81a95cfa19e701b/lib/libgap.0.dylib (unknown line)
...
```

ThomasBreuer:
@fingolfin I cannot reproduce a crash, but `Test( "juliainterface01.tst" )` gets stuck.
The line in question calls `CallJuliaFunctionWithCatch( Julia.sqrt, [ -1 ] )`, and calling this interactively runs into an error.
(Should we be worried about this?)

The problem can be solved either by calling `UnwrapJuliaFunc` in `call_with_catch` (like in `kwarg_wrapper`),
or by calling `Julia.GAP.UnwrapJuliaFunc` already on the GAP side.
(I think I prefer the latter.)

Afterwards, the tests in this file show a difference at `Print( fun );`, and replacing this by `Print( fun, "\n" );` makes this difference disappear. (I do not understand why the old version of the test file had been accepted before.)

It is perhaps confusing that the wrapped and the unwrapped version are `View`ed equally:
```
gap> gfun:= Julia.sqrt;
<Julia: sqrt>
gap> IsFunction( gfun );
true
gap> jfun:= Julia.GAP.UnwrapJuliaFunc( gfun );
<Julia: sqrt>
gap> IsFunction( jfun );
false
```

fingolfin:
@ThomasBreuer how shall we proceed with this? Some things that perhaps should be done:
- `View` wrapped and the unwrapped functions differently?
- document (un)wrapping of functions explicitly, perhaps in a section on "performance hints"
- ... ?

ThomasBreuer:
I have created the new issue #645 that takes care of the open items. Now the pull request can be merged.

PR #640, opened 2021-04-13
Align banner printing code to other Oscar packages

fingolfin:
Rename GAP_SHOW_BANNER to GAP_PRINT_BANNER which matches the names
used in Nemo, Hecke, Singular.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -204,8 +204,8 @@ function __init__()
         append!(cmdline_options, ["--nointeract"])
     end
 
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
Is only the name change for the `ENV` variable intended, or also the changed behaviour?
In the old version, the result in case of the available `ENV` variable depends only on its value.
In the new version, also the other conditions are considered.

ThomasBreuer (approved):

PR #642, opened 2021-04-13
support using Julia's random sources in GAP

ThomasBreuer:
- `GAP.wrap_rng` creates a GAP random source that uses the given Julia rng
- cache the pairs of Julia/GAP random sources in a dictionary
- on the GAP side, provide `State`, `Reset`, `Init` methods as specified by the GAP documentation

an open question:

- Why is there no `WeakKeyIdDict` in Julia?
  (Perhaps @rfourquet knows about the background.)
  ~~I think it would be enough to store an `IdDict` in the field `ht` of a `WeakKeyDict`, but one can do better than duplicate `base/weakkeydict.jl`.~~
  No, the situation seems to be more tricky ...)

### DIFF
--- a/pkg/JuliaInterface/gap/adapter.gi
+++ b/pkg/JuliaInterface/gap/adapter.gi
@@ -138,8 +138,9 @@ InstallMethod( Init,
     [ "IsRandomSourceJulia", "IsObject" ],
     function( rng, seed )
     ImportJuliaModuleIntoGAP( "Random" );
-    if IsInt( seed ) and 0 <= seed then
-      # This means a prescribed seed.
+    if IsInt( seed ) then
+      # This means a prescribed seed.  (Negative seeds are not supported.)
+      seed:= AbsInt( seed );
       if HasJuliaPointer( rng ) then
         Julia.Random.seed\!( JuliaPointer( rng ), GAPToJulia( seed ) );
       else
--- a/pkg/JuliaInterface/gap/adapter.gi
+++ b/pkg/JuliaInterface/gap/adapter.gi
@@ -139,7 +139,7 @@ InstallMethod( Init,
     function( rng, seed )
     ImportJuliaModuleIntoGAP( "Random" );
     if IsInt( seed ) then
-      # This means a prescribed seed.  (Negative seeds are not supported.)
+      # This means a prescribed seed.
       seed:= AbsInt( seed );
       if HasJuliaPointer( rng ) then
         Julia.Random.seed\!( JuliaPointer( rng ), GAPToJulia( seed ) );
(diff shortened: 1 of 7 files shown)


ThomasBreuer:
@fingolfin How shall we proceed with this --first settle gap-system/gap/issues/4379 and if necessary adjust the case of the Julia random number generators to the situation in GAP, or merge the current pull request (w.r.t. the current interface in GAP) and adjust to changes in GAP when this becomes necessary?

pkg/JuliaInterface/gap/adapter.gi
```diff
@@ -117,6 +117,101 @@ InstallOtherMethod( \.\:\=,
...
+# Then the pair given by 'rng' and its 'JuliaPointer' value may be cached
+# in the Julia session.
+# Thus we cannot replace the 'JuliaPointer' value by a copy.
+# Instead we change the value in place.
+InstallMethod( Init,
+    [ "IsRandomSourceJulia", "IsObject" ],
+    function( rng, seed )
+    ImportJuliaModuleIntoGAP( "Random" );
+    if IsInt( seed ) then
+      # This means a prescribed seed.  (Negative seeds are not supported.)
```
fingolfin, line 142:
```suggestion
      # This means a prescribed seed.
```

pkg/JuliaInterface/gap/adapter.gi
```diff
@@ -117,6 +117,101 @@ InstallOtherMethod( \.\:\=,
...
+# Instead we change the value in place.
+InstallMethod( Reset,
+    [ "IsRandomSourceJulia and HasJuliaPointer", "IsObject" ],
+    function( rng, seed )
+    local old;
+
+    old:= State( rng );
+    ImportJuliaModuleIntoGAP( "Random" );
+    if IsInt( seed ) and 0 <= seed then
+      # This means a prescribed seed.
```
fingolfin, line 184:
```suggestion
    if IsInt( seed ) then
      # This means a prescribed seed.
      seed:= AbsInt( seed );
```

pkg/JuliaInterface/gap/adapter.gi
```diff
@@ -117,6 +117,101 @@ InstallOtherMethod( \.\:\=,
...
+      Error( "<seed> must be a nonnegative integer ",
+             "or a Julia random number generator" );
+    fi;
+
+    return old;
+    end );
+
+InstallMethod( Random,
+    [ "IsRandomSourceJulia and HasJuliaPointer", "IsDenseList" ],
+    { rng, list } -> list[ Random( rng, 1, Length( list ) ) ] );
```
fingolfin, line 200: Is this really necessary? Doesn't the following GAP library method already take care of this?
```gap
InstallMethod(Random, [IsRandomSource, IsList], function(rs, list)
  return list[Random(rs, 1, Length(list))];
end);
```

fingolfin (approved):
Seems OK to me, except for a minor inconsistency that I commented on.

Once that's fixed, this could be merged from my POV.

pkg/JuliaInterface/gap/adapter.gi
ThomasBreuer: Yes, after today's discussion how to change the setup in GAP, this method does not make sense.

Note that the mentioned  library method is wrong: If the list is not dense then one runs into an error.
In the (intermediate) setup of my pull request for gap-system where non-dense lists were treated "correctly", an additional method for dense list would make sense; but admittedly this method should have been added to GAP itself.

PR #647, opened 2021-05-05
new macro @gapattribute

ThomasBreuer:
This is intended to address oscar-system/Oscar.jl/issues/128. ~~but it does not work as expected~~
~~Apparently something is wrong with the code: When I try to call the macro inside an Oscar source file then Julia crashes with a segmentation fault.~~

~~However, the following works.~~
Here is an example how to use the new macro.

1. Create a file `testmodule.jl` with a Julia module that calls the macro in the way we want to use it.
```
module enclose
  
import GAP
import Oscar
GAP.@gapattribute isfinite(G::Oscar.GAPGroup) = GAP.Globals.IsFinite(G.X)
GAP.@gapattribute issolvable(G::Oscar.GAPGroup) = GAP.Globals.IsSolvableGroup(G.X)
end # module
```
2. Use the functions created by the macro.
```
julia> using Oscar
[...]

julia> include( "testmodule.jl" )
Main.enclose

julia> g = symmetric_group(5);

julia> Main.enclose.hasissolvable(g)
false

julia> Main.enclose.issolvable(g)
false

julia> Main.enclose.hasissolvable(g)
true

julia> g = symmetric_group(5);

julia> Main.enclose.hasissolvable(g)
false

julia> Main.enclose.setissolvable(g, false)

julia> Main.enclose.hasissolvable(g)
true
```

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -50,6 +50,7 @@ GapObj
 @gap
 @g_str
 @gapwrap
+@gapattribute
 ```
 
 ## Conversions
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -154,7 +154,7 @@ macro gapwrap(ex)
     # take the body of the function
     body = def_dict[:body]
 
-    # find, record and substitute all occurrences of GAP.Global.*
+    # find, record and substitute all occurrences of GAP.Globals.*
     symdict = IdDict{Symbol,Symbol}()
     body = MacroTools.postwalk(body) do x
         MacroTools.@capture(x, GAP.Globals.sym_) || return x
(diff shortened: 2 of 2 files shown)


ThomasBreuer:
With the second attempt (after a discussion with @fingolfin), the segmentation faults are avoided.
On the one hand, the new code is better because the aspects addressed by the two macros `@gapwrap` and `@gapattribute` are now separated; for the given situation, this is fine.
On the other hand, I find it strange that the "easy metaprogramming approach" from my first attempt runs into segmentation fault problems.

fingolfin:
Note that with https://github.com/JuliaLang/julia/pull/40852 at least there will be a somewhat helpful (?) error message instead of a crash when using GAP objects on the top level of package modules (where the precompilation code ends up trying to serialize them).

fingolfin (approved):
Looks good to me

PR #656, opened 2021-06-29
Run CI tests against Julia 1.7.x

fingolfin:


fingolfin:
The failure is caused by https://github.com/JuliaLang/julia/pull/41308 missing in 1.7.0-beta2 ; it will be in beta3, see https://github.com/JuliaLang/julia/pull/41382

fingolfin:
I am actually surprised that the tests pass -- they don't pass for me locally, due to breaking ABI changes in Julia 1.7, so https://github.com/JuliaPackaging/Yggdrasil/pull/3384 should be merged (but it isn't at the time I am writing this). Huh????

PR #658, opened 2021-07-05
changed the definition of `UnwrapJuliaFunc`

ThomasBreuer:
Up to now, we defined `UnwrapJuliaFunc(x::Function) = x` and then
added a method for `GapObj`.
This is not good enough, since there are callable Julia objects
which do not have the type `Function`.

Concretely:
`Polymake.polytope.Polytope` is a callable Julia object
which is not a `Function` and which admits keyword arguments.
When we call it from GAP via `CallJuliaFunctionWithKeywordArguments`,
we get an error with the old version.

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -133,8 +133,25 @@ NEW_MACFLOAT(x::Float64) = ccall((:NEW_MACFLOAT, libgap), GapObj, (Cdouble,), x)
 ValueMacFloat(x::GapObj) = ccall((:GAP_ValueMacFloat, libgap), Cdouble, (Any,), x)
 CharWithValue(x::Cuchar) = ccall((:GAP_CharWithValue, libgap), GapObj, (Cuchar,), x)
 
+# `WrapJuliaFunc` and `UnwrapJuliaFunc` are intended to create a GAP function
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
OK, fine.

But what about the other direction? Should `WrapJuliaFunc` also be adapted?

ThomasBreuer:
Concerning the other direction (`WrapJuliaFunction`), I think this would make sense:

Then `WrapJuliaFunction` maps a Julia `Function` to a `GapObj` that is a GAP function, and maps other Julia objects (assuming they are callable) to themselves, thus calling them as a function on the GAP side does the same as calling them in Julia.
And `UnwrapJuliaFunction` reverses this, it extracts the underlying Julia `Function` from a `GapObj`, and is the identity on other Julia objects.

Of course, the situation is not completely symmetric: When one calls `WrapJuliaFunction` with a `GapObj` x, say, then x is returned, and calling `UnwrapJuliaFunction` on x does in general not return x.
If x is a wrapped Julia function then it gets unwrapped, and otherwise Julia is likely to crash.

fingolfin:
OK this now has a conflict which has to be resolved

Issue #660, opened 2021-07-07
Find a way so that Julia prints `GapObj` as `GAP.GapObj` not as `GAP_jll.MPtr`


fingolfin:
Perhaps this just a matter of installing a suitable `show` method (or so), or perhaps we can hack the name somewhere possibly even via a call to the Julia API... 

Low priority, but would improve user friendliness.

PR #665, opened 2021-07-14
`GapObj` now prints itself as `GapObj`

ThomasBreuer:
This addresses issue #660.
It is an ugly hack, and I do not know whether it will have unwanted side-effects.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -229,6 +229,10 @@ function __init__()
             ),
         )
     end
+
+    # hack: Let `GAP_jll.MPtr` print itself as `GAP.GapObj`.
+    GAP_jll.MPtr.name.module = GAP
+    GAP_jll.MPtr.name.name = :GapObj
 end
 
 """
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -229,10 +229,6 @@ function __init__()
             ),
         )
     end
-
-    # hack: Let `GAP_jll.MPtr` print itself as `GAP.GapObj`.
-    GAP_jll.MPtr.name.module = GAP
-    GAP_jll.MPtr.name.name = :GapObj
 end
 
 """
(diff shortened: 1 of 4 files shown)


src/GAP.jl
```diff
@@ -229,6 +229,10 @@ function __init__()
             ),
         )
     end
+
+    # hack: Let `GAP_jll.MPtr` print itself as `GAP.GapObj`.
+    GAP_jll.MPtr.name.module = GAP
+    GAP_jll.MPtr.name.name = :GapObj
```
rfourquet, line 235: Maybe defining `Base.show(io::IO, ::Type{GAP_jll.MPtr}) = print(io, "GapObj")` would be slightly less hacky, would it good enough?

ThomasBreuer: Thanks @rfourquet .
After `import GAP`, one can access `GAP.GapObj` but not `GapObj`.
After `using GAP`, one can access also `GapObj`.
How can I achieve with your approach that "GAP.GapObj" is printed in the first case, and "GAP" in the latter?

rfourquet: Taking inspiration from `Base.show_type_name`:
```julia
function Base.show(io::IO, ::Type{GapObj})
    from = get(io, :module, Main)
    if from === nothing || !Base.isvisible(:GapObj, GAP, from)
        show(io, GAP)
        print(io, ".")
    end
    print(io, "GapObj")
end
```

rfourquet: Possibly wrapping the printing of the prefix within `if !(get(io, :compact, false)::Bool)` like in `Base.show_type_name`.

fingolfin: Thank you @rfourquet

ThomasBreuer: Thanks @rfourquet , I have updated my proposal.

fingolfin (approved):

PR #667, opened 2021-07-16
Some helper scripts for setting up JLL overrides

fingolfin:
This is unfinished, but since I've had﻿it sitting around for
two weeks now, I thought it best to put it in a PR so I don't
forget it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /gaproot/
 /gh-pages/
-/Manifest.toml
+Manifest.toml
 /tags
 *.log
+
+proj*
--- a/etc/README.md
+++ b/etc/README.md
@@ -26,7 +26,17 @@ The latter works because we have set up a GitHub Action for
 make the release automatically.
 
 
-## Using GAP.jl with a different version of GAP than what GAP_jll provides
+## Updating `GAP_jll` resp. `GAP_lib_jll`
+
+To do this, open a pull request for <https://github.com/JuliaPackaging/Yggdrasil>.
+The files to edit there are in `G/GAP` (produces a JLL with compiled code for
+the GAP kernel) and `G/GAP_lib` (produces a JLL with the GAP library code, no
+binaries). Usually, those two JLLs should be updated simultaneously; but if
+one knows that only the kernel resp. only the GAP library, one can also
+deviate from this.
+
+
+## Using GAP.jl with a different version of GAP than what `GAP_jll` provides
 
 This can be useful for various reasons e.g.,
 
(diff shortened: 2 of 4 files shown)


ThomasBreuer (approved):
This works.
I found only the following items for improving the instructions how to use the override.
- Better state explicitly that the last argument for `etc/setup_override_dir.jl` must be an absolute path.
- If the target directory for `etc/setup_override_dir.jl` exists already, it has to be removed first. This could be checked before configuring and compiling.

Irritating:
After calling `etc/run_with_override.jl` and `using GAP`, the GAP banner shows `GAPInfo.BuildDateTime` in its first line. This date is not the one of the current build process but the date for an older version.

fingolfin:
@ThomasBreuer wrote:

> This works.
> I found only the following items for improving the instructions how to use the override.
> 
> * Better state explicitly that the last argument for `jul` must be an absolute path.

Actually, the script should just turn this into an absolute path. Will do.

> * If the target directory for `etc/setup_override_dir.jl` exists already, it has to be removed first. This could be checked before configuring and compiling.

Indeed, there is even a TODO comment for that. I'll resolve that now, too.

> 
> Irritating:
> After calling `etc/run_with_override.jl` and `using GAP`, the GAP banner shows `GAPInfo.BuildDateTime` in its first line. This date is not the one of the current build process but the date for an older version.

As discussed in private chat, this seems to be a defect in the GAP build system. Will take care of that, too.

fingolfin:
@ThomasBreuer I've taken care of all three points: added a commit to this PR which validates the paths and converts them into absolute ones. And opened a PR for the GAP build system, see https://github.com/gap-system/gap/pull/4620

ThomasBreuer:
@fingolfin I have applied gap-system/gap#4620; building GAP works afterwards.
Then I have checked out the current version of #667, and when I call
`julia --proj=override-1.5 etc/setup_override_dir.jl ~/gapmaster/gap/ /tmp/gap_jll_override-1.5`,
I get the message

``
[ Info: Building GAP in /tmp/jl_xaRr9f
GAP_BUILD_VERSION = 4.12dev
/home/sam/gapmaster/gap/cnf/gap-version-gen.sh: line 45: cnf/GAP-VERSION-FILE: No such file or directory
/home/sam/gapmaster/gap/cnf/gap-version-gen.sh: line 46: cnf/GAP-VERSION-FILE: No such file or directory
make: *** No rule to make target 'cnf/GAP-VERSION-FILE', needed by 'build/gap_version.c'.  Stop.
make: *** Waiting for unfinished jobs....
    * new C compiler flags
    * new preprocessor flags
    * new C++ compiler flags
ERROR: LoadError: failed process: Process(`make -j8 build/gap-nocomp`, ProcessExited(2)) [2]
``

fingolfin:
I have updated https://github.com/gap-system/gap/pull/4620 to hopefully fix that error.

PR #672, opened 2021-07-26
Replace Array by Vector or Matrix where appropriate

fingolfin:
This was done with the following commands:

    git ls-files | xargs perl -pi -e 's/Array\{([^{}]+(\{((?>[^{}]+)|(?2))*\})?),\s*2\s*\}/Matrix{\1}/g'
    git ls-files | xargs perl -pi -e 's/Array\{([^{}]+(\{((?>[^{}]+)|(?2))*\})?),\s*1\s*\}/Vector{\1}/g'
    git ls-files | xargs perl -pi -e 's/Array\{([^{}]+(\{((?>[^{}]+)|(?2))*\})?),\s*1\s*\}/Vector{\1}/g'

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -249,7 +249,7 @@ julia> val = GAP.evalstr("[ true, false, true ]")
 GAP: [ true, false, true ]
 
 julia> BitArray{1}(val)
-3-element BitArray{1}:
+3-element BitVector:
  1
  0
  1
@@ -479,7 +479,7 @@ julia> r = StepRange{Int8,Int8}(val)
 1:2:11
 
 julia> typeof(r)
-StepRange{Int8,Int8}
+StepRange{Int8, Int8}
 
 ```
 """ StepRange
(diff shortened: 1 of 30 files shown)


src/constructors.jl
```diff
@@ -249,7 +249,7 @@ julia> val = GAP.evalstr("[ true, false, true ]")
 GAP: [ true, false, true ]
 
 julia> BitArray{1}(val)
-3-element BitArray{1}:
+3-element BitVector:
```
ThomasBreuer, line 252: Strange: This is the only place where `BitArray{1}` got changed to `BitVector`. Why did the replacements find this one but not the others?

ThomasBreuer (changes requested):
If I understand the changes right then the Julia output got adjusted after the regular expressions had been applied; otherwise I do not understand the changed spacing in `Dict` and `StepRange` outputs.
I think we should also replace `BitArray{1}` by `BitVector`, but this should then be all, according to the doctestfilters we use.

src/constructors.jl
fingolfin: Just to record in public what I mentioned in private chat: this second commit was made by running `doctest(GAP, fix=true)`.

fingolfin:
I've added a commit which replace BitArray by BitVector

ThomasBreuer (approved):

Issue #611, opened 2021-01-22
Better gap.sh support


fingolfin:
Right now we have this code in `src/setup.jl` which generates a `gap.sh` so that people can invoke GAP "as usual" (although in the background this starts Julia load GAP.jl and then opens a GAP prompt).
```julia
        ##
        ## Create Project.toml & Manifest.toml for use by gap.sh
        ##
        @info "Generating custom Julia project ..."
        relative_pkgdir = joinpath("..", "..")
        @assert abspath(gaproot_mutable, relative_pkgdir) == gaproot_gapjl
        run(pipeline(`$(Base.julia_cmd()) --startup-file=no --project=$(gaproot_mutable) -e "using Pkg; Pkg.develop(PackageSpec(path=\"$(relative_pkgdir)\")); Pkg.resolve()"`))
```

However, this slows down package precompilation and still is inflexible: in order to use that `gap.sh`, people need to either launch Julia, load GAP.jl, call `gap_exe` then execute the resulting path (thus loading GAP *twice*, for twice the overhead); or they need to add a symlink somewhere, which they then need to potentially update all the time.

So I think we need something better, I am just not yet quite sure *what*... Here's one idea: We add a function which uses can call and which places a "fake gap.sh" in a place of their choice, along with the  `Project.toml` as generated by the code I quoted above (perhaps hiding it away somewhere?). Then it is less black magic and more explicit. And we can remove anything related to it from precompilation.

Thoughts? Alternative ideas? @mohamed-barakat @ThomasBreuer

ThomasBreuer:
Up to now, I have used the symlink variant, that is, I have created an `alias` pointing to the `gap.sh` in question.
The path encodes the location of `GAP.jl` (which does not change except if the user decides to change it) and the Julia version number (o.k., this changes sometimes, but not very often).

The classical GAP user gets the `gap.sh` by calling `./configure; make` after unpacking the GAP archive.
Users of GAP via Julia will install GAP by first installing and starting Julia and then calling something like `using Pkg; Pkg.add("GAP")`. Afterwards they can start GAP inside Julia via `GAP.prompt()`. I think that explicitly calling another Julia function in order to create `gap.sh`, as an optional part of the GAP installation, would be fine.

mohamed-barakat:
> The path encodes the location of `GAP.jl` (which does not change except if the user decides to change it)

[...]

PR #678, opened 2021-08-17
Move code generating gap.sh into its own function

fingolfin:
The next step would be to not create gap.sh by default anymore, but instead turn this into a user visible function. Then user's can create any number of `gap.sh` scripts they like, in a place of their choosing (but note that it'll be accompanied by a `Project.toml` and `Manifest.toml` in the same dir; that needs to be documented, too).

See also issue #611

CC @mohamed-barakat

### DIFF
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -243,22 +243,55 @@ function regenerate_gaproot(gaproot_mutable)
         # create a `pkg` symlink to the GAP packages artifact
         force_symlink(artifact"gap_packages", "pkg")
 
-        ##
-        ## Create Project.toml & Manifest.toml for use by gap.sh
-        ##
-        @info "Generating custom Julia project ..."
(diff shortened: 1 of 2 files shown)


src/setup.jl
```diff
@@ -243,22 +243,43 @@ function regenerate_gaproot(gaproot_mutable)
...
+    ## Create custom gap.sh
+    ##
+    create_gap_sh(joinpath(gaproot_mutable, "bin"))
+
+    return gaproot_mutable
+end # function
 
-    end # cd
+
+function create_gap_sh(dstdir)
```
fingolfin, line 267: Ultimately, a docstring needs to be added and perhaps a better name be selected.

Also, if users are supposed to call this, perhaps some sanity tests should be performed? 

Should we require `dstdir` to exist? Or do the opposite and demand that it must not exist? Or create it if missing?

PR #683, opened 2021-08-19
Don't create `gap.sh` anymore and remove `GAP.gap_exe();` instead users can call `GAP.create_gap_sh(path)`

fingolfin:
This patch stops us from creating a `gap.sh` during precompilation; instead we instruct users to use `create_gap_sh()`. This also means removing `gap_exe()`.

But this is a breaking change for anybody using `gap_exe()` (I suspect @mohamed-barakat is one of these people). So, to allow for a clean transition, first we'll release GAP.jl 0.6.2 with `create_gap_sh` in it, at which point people can convert their code to use that. Then when we release 0.7, we can merge this PR shortly before.

UPDATE: Should also get rid of `src/obsolete.jl` for 0.7.0

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -61,8 +61,8 @@ jobs:
         run: julia --color=yes -e 'using Pkg ; Pkg.add(["Singular", "Nemo"])'
       - name: "GAP tests"
         run: |
-          gap_sh=$(julia --project=. -e 'import GAP; print(GAP.gap_exe())');
-          export GAP="$gap_sh -A --quitonbreak --norepl"
+          julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")'
+          export GAP="/tmp/gap.sh -A --quitonbreak --norepl"
           etc/ci_test.sh
 
       - name: "Process Julia code coverage"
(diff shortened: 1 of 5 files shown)


mohamed-barakat (approved):
Looks good. I am not using `gap_exe()` :)

.github/workflows/CI.yml
```diff
@@ -61,9 +61,13 @@ jobs:
         run: julia --color=yes -e 'using Pkg ; Pkg.add(["Singular", "Nemo"])'
       - name: "GAP tests"
         run: |
-          gap_sh=$(julia --project=. -e 'import GAP; print(GAP.gap_exe())');
-          export GAP="$gap_sh -A --quitonbreak --norepl"
+          echo "Check 0"
+          $(julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")')
```
benlorenz, line 65:
```suggestion
          julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")'
```

.github/workflows/CI.yml
```diff
@@ -111,9 +114,9 @@ jobs:
           julia --project=docs --color=yes test/doctest.jl
       - name: "Build GAP manual"
         run: |
-          gap_sh=$(julia --project=. -e 'import GAP; print(GAP.gap_exe())');
+          $(julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")')
```
benlorenz, line 117:
```suggestion
          julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")'
```

.github/workflows/CI.yml
fingolfin: Thanks, that was it!!

PR #695, opened 2021-09-14
Override PKGMAN_DownloadURL using Downloads.download

### LINKED ISSUE #692: Override GAP function `PKGMAN_DownloadURL` to use Julia for downloading?
The problem with `PKGMAN_DownloadURL` is that it does different things for different systems (e.g. it might use curl or wget or curlInterface or maybe something else), and if those run into errors, it can be hard to reproduce and debug. If we instead use Julia's `download()` we should get more consistent behavior. That is, assuming `download` can do what we need.

lgoe-ac:
closes #692

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsch
 version = "0.6.2"
 
 [deps]
+Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 GAP_jll = "5cd7a574-2c56-5be2-91dc-c8bc375b9ddf"
 GAP_lib_jll = "de1ad85e-c930-5cd4-919d-ccd3fcafd1a3"
 GMP_jll = "781609d7-10c4-51f6-84f2-b8444358ff6d"
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsch
 version = "0.6.2"
 
 [deps]
+Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
```
fingolfin, line 7: Perhaps also add a compat entry?

fingolfin:
Great, thanks!

Project.toml
lgoe-ac: oh yes, I forgot that. Then just "1.5.1" (i.e. the current version)? Otherwise, I do not really know how to forcefully downgrade my environment for testing

fingolfin: Fine by me

lgoe-ac: Apparently, I have been using 1.4.2 the entire time. Thus this is now in the compat entry. (Why i cannot upgrade to the current 1.5.1 is another story)

fingolfin:
Which Julia version do you use? This may restrict the version of Download.jl, as it is a stdlib in newer Julia versions

lgoe-ac:
```julia
julia> versioninfo()
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

fingolfin:
Yeah and its a stdlib in 1.6, so that fits

PR #696, opened 2021-09-14
Remove LoadPackageAndExposeGlobals

fingolfin:
This function always was very peculiar and for special interests only.
It is also very easily abused resp. used in a problematic way.

I believe that any code which uses this right now should be rewritten
(and I'll be happy to help with that by suggesting ways to do it).
Alternatively, one can of course also copy the code deleted by this
patch into one's own code.

Indeed, right now I just know about `CapAndHomalg.jl` and `HomalgProject.jl` using this, by @mohamed-barakat . Now, both of them use `LoadPackageAndExposeGlobals` to import things into `Main`. That's a very dangerous thing for a package to do. It would be less problematic if the functions were at least imported into a fresh module of their own. In retrospect, I think it was a mistake that we ever allowed to importing things into arbitrary modules; because that suggests it is OK to do so, and then it gets used.

@mohamed-barakat you only use the second method, and only in a special way. I think something like this would do the job for you (but again, I would recommend *against* putting these things into `Main` -- instead, I would import them into e.g. the module `CapAndHomalg` and then `export` it from there. This way. The user can at least just whether to `import` or `using` your package. Also, I think you do *not* want to silently ignore it when one of the methods is not imported, do you?! Otherwise, modify the loop `for sym in new_symbols` to check for it

```julia
function LoadPackageAndExposeGlobals(package::String)
    old_gvar_list = Globals.ShallowCopy(Globals.NamesGVars())
    if !GAP.Packages.load(package)
        error("cannot load package $package")
    end
    new_gvar_list = Globals.Difference(Globals.NamesGVars(), old_gvar_list)
    new_symbols = Vector{Symbol}(new_gvar_list)
    for sym in new_symbols
        # FIXME: do not import into Main, use some other module
        # Note that we must specify `GAP.` here.
        Main.eval(:($(sym) = GAP.Globals.$(sym)))
        #Main.eval(:(export $(sym)))
    end
end
```


mohamed-barakat:
Thank you very much for your suggestion. I should have for `CapAndHomalg` a substitute along these lines by the beginning of next week.

fingolfin:
@mohamed-barakat OK, I'll merge this now but the next version will be 0.7, so due to semver, you can keep using 0.6 (or 0.5 or...) until you are read to switch!

mohamed-barakat:
Yes, good idea. Thank you very much.

PR #700, opened 2021-09-16
Explicit types in ConversionError & default gap_to_julia

fingolfin:
I also wonder if the default `gap_to_julia` method should be changed to call `convert` ?  But that's another question. CC @thofma @ThomasBreuer

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -1,8 +1,8 @@
 
 ## Show a specific error on conversion failure.
 struct ConversionError <: Base.Exception
-    obj::Any
-    jl_type::Any
+    obj::Obj
+    jl_type::Union{String,Type}
 end
 
 Base.showerror(io::IO, e::ConversionError) =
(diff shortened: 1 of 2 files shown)


src/gap_to_julia.jl
```diff
@@ -64,13 +64,10 @@ function gap_to_julia(t::T, x::Any) where {T<:Type}
...
         throw(ErrorException(
-            "Don't know how to convert value of type " *
-            string(typeof(x)) *
-            " to type " *
-            string(t),
+            "Don't know how to convert value of type $(typeof(x)) to type $(t)"
         ))
     end
-    return x
+    return x::T
```
thofma, line 70: There is no `T` anymore.

Do you want this to be fast, that is, do you want it to specialize for all `T`?

fingolfin: Is that you telling me very nicely that I shouldn't be an idiot and use `@nospecialize`? ;-) Because I think that sounds like a good suggestion... Thanks!

PR #704, opened 2021-09-16
Remove convert methods

fingolfin:
This may break some existing code, so should only go to 0.7

CC @thofma @ThomasBreuer


pkg/JuliaExperimental/gap/numfield.g
```diff
@@ -288,8 +288,7 @@ InstallMethod( ContextGAPNemo,
     if Length( indetnames ) = 1 then
       names:= GAPToJulia( indetnames[1] );
     else
-      names:= Julia.Base.convert( JuliaEvalString( "Vector{String}" ),
-                                  GAPToJulia( indetnames ) );
+      names:= JuliaEvalString( "Vector{String}" )( GAPToJulia( indetnames ) );
```
fingolfin, line 291: Off-topic remark: For production code, the value `JuliaEvalString( "Vector{String}" )` should probably be computed only once and stored in a global variable.

pkg/JuliaExperimental/ipynb/GAPJulia_Singular_car.ipynb
```diff
@@ -101,8 +101,7 @@
    ],
    "source": [
     "indetnames:= ConvertedToJulia( [ \"x0\", \"x1\", \"x2\", \"x3\" ] );\n",
```
fingolfin, line 103: I think `ConvertedToJulia` is not needed, here?!

pkg/JuliaExperimental/ipynb/GAPJulia_Singular_car.ipynb
```diff
@@ -404,9 +403,7 @@
    ],
    "source": [
     "arr1:= ConvertedToJulia( [ x0, x3, x1, x2, x3, x0 ] );;\n",
```
fingolfin, line 405: Again, I think `ConvertedToJulia` is not needed, here?!

pkg/JuliaExperimental/gap/numfield.g
ThomasBreuer: We have already quite a few of these Julia types in global GAP variables (`_JL_Vector_Bool`, `_JL_Vector_Any`, `_JL_Dict_Symbol`), this pattern can be continued.
In order to (at least) avoid the `JuliaEvalString` call, we could also provide a more programmatic approach: If GAP.jl defines `typeconstruct(T1::Type, T2::Type) = T1{T2}` then one can write `Julia.GAP.typeconstruct( Julia.Vector, Julia.Int )` in a GAP session, and gets the Julia type `Vector{Int64}`; would something in this direction be useful?

pkg/JuliaExperimental/ipynb/GAPJulia_Singular_car.ipynb
ThomasBreuer: Yes, `ConvertedToJulia` and `ConvertedFromJulia` are no longer available and are not needed.
(And also some outputs look meanwhile different from what is shown in the notebook.)

I had not looked at this notebook for a long time because of the problems with the Jupyter kernel for GAP.
Now I have tried GAP's JupyterKernel again. Without Julia connection, it seems to work (which was a surprise for me).
When I try to install JupyterKernel for a GAP that is running under Julia then Julia crashes when GAP loads ZeroMQInterface.

ThomasBreuer (approved):
I have checked that my Julia code outside the GAP interface does not call `convert` explicitly. As I wrote before, my understanding had been that `convert` was supported mainly because Julia offered some benefits. Thus let us try to get along without `convert`.

pkg/JuliaExperimental/gap/numfield.g
fingolfin: yeah, that would be useful. Ideally it should also deal with multiple args as in  e.g. `Dict{String, Int}` or `Array{Int, 3}`. 

I actually thought about installing a method for `ELMS_LIST` / `\{\}` to make this convenient to access from GAP code (so you could write `Julia.Vector{Julia.String}` or `Julia.Dict{[Julia.String, Julia.Int]}` but currently GAP insists on the argument being a list of positive integers. I actually have a PR in the works for GAP to lift that limitation, but I am not yet sure we really want it (but I'd like to experiment with it, and it might actually be interesting for other applications in GAP, so I figured "why not try it").

pkg/JuliaExperimental/ipynb/GAPJulia_Singular_car.ipynb
fingolfin: AFAIK the GAP JupyterKernel uses `fork`, so I am surprised it doesn't crash earlier (it definitely used to).

On the GAP slack, just this week a student from St Andrews turned up who reported that he has been hired to work on the JupyterKernel package. So here's hoping...

PR #707, opened 2021-09-17
Make use of @gapwrap

fingolfin:
... on most of the GAP functions we access from GAP.jl itself. This is
intended to be an optimization: caching the value of `GAP.Globals.FOO`
should speed up a bunch of common operations.

Also use this opportunity to add type annotations for the return values of a
bunch of them, to improve type stability and thus further optimize things.


src/ccalls.jl
```diff
@@ -73,13 +73,13 @@ function evalstr(cmd::String)
       global last_error
       # HACK HACK HACK: if there is an error string on the GAP side, call
       # error_handler to copy it into `last_error`
-      if !GAP.Globals.IsEmpty(Globals._JULIAINTERFACE_ERROR_BUFFER)
+      if !Wrappers.IsEmpty(GAP.Globals._JULIAINTERFACE_ERROR_BUFFER)
```
fingolfin, line 76:
```suggestion
      if !Wrappers.IsEmpty(Globals._JULIAINTERFACE_ERROR_BUFFER)
```

PR #711, opened 2021-09-18
Add GAP.Obj conversion constructor, overhaul GapObj constructor

fingolfin:
- `GapObj` now only returns objects of type `GapObj` for better type
  stability and general sanity (a constructor `T(x)` should always
  return an object of type `T`)
- To partially fill the void this leaves, add a similar conversion
  constructor for `GAP.Obj`, which can produce anything within that type
  union.
- Add a docstring for `GAP.Obj` and enhance that for `GapObj`


docs/src/index.md
```diff
@@ -102,7 +103,7 @@ julia_to_gap
 ### Constructor Methods for GAP-to-Julia Conversions
 
 (For Julia-to-GAP conversions,
-one can use [`GapObj`](@ref) as a constructor.)
+one can use [`GapObj`](@ref) and [`GAP.Obj`](@ref) as a constructor.)
```
ThomasBreuer, line 106: perhaps "as constructors"?

ThomasBreuer (approved):
This looks good.
(There is one typo.)
What is the reason to include `Nothing` in `GAP.Obj`: Can a native GAP object have this type?
(And do we want `GAP.julia_to_gap(nothing)` to return `nothing`?)

The test failures seem to be due to a changed format of the `TypeError` message since Julia 1.5.
Since we want to require at least Julia 1.6, this is not a problem.

PR #713, opened 2021-09-18
Remove IsArgumentForJuliaFunction

fingolfin:
We don't use it anywhere, the documentation referencing it is wrong
(a Julia function can take *any* GAP object as argument and likewise
return any GAP object), and in general I don't understand what one
might use it for.

@ThomasBreuer you originally added it, perhaps you remember better than me what this was/is about?


ThomasBreuer (approved):
The idea was as follows: If we have the two worlds of GAP and Julia, and JuliaInterface provides an interface between them then it makes sense to ask whether an object in a GAP session belongs to the Julia world in the sense that one can use it as an argument of a Julia function --if not then one has to convert it to the Julia world.
Of course this distinction does not make sense if one writes Julia code that deals with GAP objects (and already the interface does this). Probably the idea was bad, or at least the name of the function was inappropriate.

PR #714, opened 2021-09-22
Rewrite @gapattribute macro to allow for docstrings

fingolfin:
Now function defined with `@gapattribute` can have a docstring.
This is done using the `Base.@__doc__` macro. However, I actually
ended up rewriting `@gapattribute` almost from scratch.

First off, this merges `gapwrap_fun` back into `@gapwrap`.
Also `@gapwrap` now uses the `__source__` macro argument to set the
location data for the method it generates.

`@gapattribute` is changed to use a much more direct approach to
producing the getter/tester/setter methods: before, we assembled these
"by hand" using MacroTools, and used `Core.eval` to inject them into the
surrounding module, which of course made attaching a docstring
impossible. Now, we simply write down a `quote` expression which looks
virtually like the code one would write to produce the effect of
`@gapattribute` manually, except with a few key pieces replaced by
placeholders. This means we can now "call" other macros like `@gapwrap`
or `Base.@__doc__`. As a bonus, we can now easily attach docstrings to
the generated tester and setter methods. Location data is also correct
thanks to `Meta.replace_sourceloc!` and the above change where
`@gapwrap` now sues `__source__`

The resulting code is more powerful yet IMHO actually easier to understand.


src/macros.jl
```diff
@@ -186,17 +151,60 @@ macro gapwrap(ex)
...
+function replace_sourceloc!(sourceloc, @nospecialize(ex))
+    if ex isa Expr
+        if ex.head == :macrocall
+            ex.args[2] = sourceloc
+        end
+        map!(e -> replace_sourceloc!(sourceloc, e), ex.args, ex.args)
+    end
+    return ex
+end
+
```
fingolfin, line 204: Reminder: If/once PR #723 is merged, get rid of this...

ThomasBreuer (approved):
Very nice.
(Concerning the understandability of the code, I would say that I understand both the old and the new version only in the sense that I can convince myself that the code does what one wants, and that it looks reasonable.)

PR #746, opened 2021-10-29
OscarCI: add OscarDevTools based workflow to run Oscar tests

benlorenz:
This might have given an early warning for the failures in https://github.com/oscar-system/Oscar.jl/issues/774, lets try what happens here.

### DIFF
--- a/OscarCI.toml
+++ b/OscarCI.toml
@@ -1,0 +1,16 @@
+title = "metadata for oscar CI run"
+
+[env]
+os = [ "ubuntu-latest", "macos-latest" ]
+julia-version = [ "~1.6.0-0", "~1.7.0-0"]
+# branches = [ "release", "<matching>" ]
+
+[pkgs]
+  [pkgs.Oscar]
+  test = true
+
+#[include]
+#  [include.julia]
+#  Oscar = "master"
+#  julia-version = "nightly"
+#  os = "ubuntu-latest"
--- a/OscarCI.toml
+++ b/OscarCI.toml
@@ -1,7 +1,7 @@
 title = "metadata for oscar CI run"
 
 [env]
-os = [ "ubuntu-latest"= ]
+os = [ "ubuntu-latest" ]
 julia-version = [ "~1.6.0-0"]
 # branches = [ "release", "<matching>" ]
 
--- a/OscarCI.toml
+++ b/OscarCI.toml
@@ -1,8 +1,8 @@
 title = "metadata for oscar CI run"
 
 [env]
-os = [ "ubuntu-latest", "macos-latest" ]
-julia-version = [ "~1.6.0-0", "~1.7.0-0"]
+os = [ "ubuntu-latest"= ]
+julia-version = [ "~1.6.0-0"]
 # branches = [ "release", "<matching>" ]
 
 [pkgs]
--- a/.github/workflows/oscar.yml
+++ b/.github/workflows/oscar.yml
@@ -1,0 +1,90 @@
+name: OscarCI
+
+on:
+   pull_request:
+     branches:
+       - master
+
+jobs:
+  generatematrix:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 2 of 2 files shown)


benlorenz:
For `[Oscar]#master` we do get the SIGKILL after the Transitivity tests from Oscar.jl: https://github.com/oscar-system/GAP.jl/runs/4050585502?check_suite_focus=true#step:7:426

`[Oscar]#release` is expected to fail as the current Oscar release does not support GAP.jl 0.7. Interestingly there does seem to be a valid version set which is compatible with GAP 0.7.2-DEV, but not with any recent julia version ([log](https://github.com/oscar-system/GAP.jl/runs/4050585632?check_suite_focus=true)):
```
  [c3fe647b] + AbstractAlgebra v0.9.2
  [9e28174c] + BinDeps v1.0.2
  [b99e7846] + BinaryProvider v0.5.10
  [631607c0] + CMake v1.2.0
  [1f15a43c] + CxxWrap v0.9.1
  [3e1990a7] + Hecke v0.8.4
  [682c06a0] + JSON v0.21.2
  [472f376f] + LoadFlint v0.1.3
  [2edaba10] + Nemo v0.17.5
  [f1435218] + Oscar v0.2.0
  [69de0a69] + Parsers v2.1.1
  [d720cf60] + Polymake v0.3.3
  [ae029012] + Requires v1.1.3
  [bcd08a7b] + Singular v0.2.2
  [30578b45] + URIParser v0.4.1
```

fingolfin:
Yeah, Oscar.jl 0.2.0 did not yet use GAP.jl, so it is "compatible". Pretty much useless to have this, though.

fingolfin:
Running tests with Oscar 0.2.0 is pretty much useless. It would be great if the test could be skipped in that case.

fingolfin:
Annoying about this is that it increases the CI turnaround times from <5 minutes to ~40 minutes... even if we speed up the Oscar test suite it'll be slow.

OscarCI.toml
```diff
@@ -0,0 +1,16 @@
+title = "metadata for oscar CI run"
+
+[env]
+os = [ "ubuntu-latest", "macos-latest" ]
+julia-version = [ "~1.6.0-0", "~1.7.0-0"]
```
fingolfin, line 5:
```suggestion
os = [ "ubuntu-latest"= ]
julia-version = [ "~1.6.0-0"]
```

OscarCI.toml
```diff
@@ -0,0 +1,16 @@
+title = "metadata for oscar CI run"
+
+[env]
+os = [ "ubuntu-latest"= ]
```
fingolfin, line 4:
```suggestion
os = [ "ubuntu-latest" ]
```

benlorenz:
> Running tests with Oscar 0.2.0 is pretty much useless. It would be great if the test could be skipped in that case.

Unfortunately that is non-trivial with github actions, once a job is running it cannot be skipped anymore. We could ignore the failure and skip the test-run, but the status would still be either "success" or "fail" which both is misleading.

benlorenz:
> Annoying about this is that it increases the CI turnaround times from <5 minutes to ~40 minutes... even if we speed up the Oscar test suite it'll be slow.

Just running them on `master` instead of per PR would also be an option to reduce the impact, then one can/should check the status before triggering a release.

fingolfin:
Re time for running the Oscar test suite: I hope it will actually go back down to 20 (from 40) minutes soon, which seems less outlandish

Re not being able to skip a test once it is running: I understand this is not currently possible, but if the test was simply marked as "passing" in that case (meaning: no Oscar.jl version >= 0.3 (or 0.6, doesn't matter) can be loaded together with the GAP.jl dev version being tested), I actually think that'd be appropriate and exactly what I want: it'd mean that releasing that GAP won't break any currently released Oscar version; all I care about in that case is that the Oscar dev version isn't broken (which the other test is checking already).

PR #755, opened 2021-11-10
Detect if there is an exception while exiting

fingolfin:
... and if so, ensure a non-zero exit code, so that our CI catches it

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,10 @@
   enter `GAP.Globals.MTX.Is` into the REPL and press the TAB key twice, you
   should be offered a list of members of the record `GAP.Globals.MTX` whose
   name starts with `Is`.
+- Fix a bug where running GAP through Julia via a `gap.sh` wrapper created
+  using `GAP.create_gap_sh` could produce an error (specifically when a `QUIT`
+  statement is encountered while processing a GAP file by passing its path as
+  argument to `gap.sh`)
 
 ## Version 0.7.1 (released 2021-10-29)
 
--- a/pkg/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/JuliaInterface/src/JuliaInterface.c
@@ -151,6 +151,12 @@ jl_function_t * get_function_from_obj_or_string(Obj func)
 }
 
 
+void ResetUserHasQUIT(void)
+{
+    STATE(UserHasQUIT) = 0;
+}
+
+
 /*
  * Returns the function from the Object <func>
  * or the function with name <func> from
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -223,7 +223,14 @@ function __init__()
     end
 
     # ensure GAP exit handler is run when we exit
-    Base.atexit(function() GAP.Globals.PROGRAM_CLEAN_UP() end)
+    Base.atexit(function()
+                    try
+                        GAP.Globals.PROGRAM_CLEAN_UP()
+                    catch e
+                        showerror(stderr, e, catch_backtrace())
+                        exit(1) # signal error
+                    end
+                end)
 
     # Respect the -q flag
     isquiet = Bool(Base.JLOptions().quiet)
--- a/src/prompt.jl
+++ b/src/prompt.jl
@@ -59,6 +59,17 @@ function run_session()
         ccall((:Call0ArgsInNewReader, GAP_jll.libgap), Cvoid, (Any,), Globals.SESSION)
     end
 
+    # Reset the GAP kernel variable `UserHasQUIT` so that GAP's exit handlers
+    # can run. This is necessary if the user passed a file on the command line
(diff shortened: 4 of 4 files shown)


fingolfin:
TO reproduce: I created a GAP wrapper script via `julia --project=p/1.6 -e 'import GAP; GAP.create_gap_sh("/tmp")'` (the project is how I work, you can adjust that bit).

Then I run `/tmp/gap.sh -A ; echo $?`  -- if I do this and just quit via Ctrl-D, `quit;` or `QUIT;` it works fine.

But if I do this, I get an error:
```
$ echo "QUIT;" > foo.g
$ /tmp/gap.sh -b -A foo.g ; echo $?
About to read command line files
starting a session
about to exit with status 0
Error thrown by GAP:
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] ThrowObserver(depth::Int32)
    @ GAP ~/Projekte/OSCAR/GAP.jl/src/GAP.jl:51
  [3] _call_gap_func(func::GapObj)
    @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:250
  [4] call_gap_func_nokw
    @ ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:224 [inlined]
  [5] call_gap_func(::GapObj; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:213
  [6] call_gap_func
    @ ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:206 [inlined]
  [7] #_#10
    @ ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:231 [inlined]
  [8] (::GapObj)()
    @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:231
  [9] (::GAP.var"#1#4")()
    @ GAP ~/Projekte/OSCAR/GAP.jl/src/GAP.jl:228
 [10] _atexit()
    @ Base ./initdefs.jl:343
 [11] exit(n::Int64)
    @ Base ./initdefs.jl:28
 [12] top-level scope
    @ /tmp/gap.sh:15
 [13] include(mod::Module, _path::String)
    @ Base ./Base.jl:386
 [14] exec_options(opts::Base.JLOptions)
    @ Base ./client.jl:285
 [15] _start()
    @ Base ./client.jl:4851
```

fingolfin:
OK I understand a bit better now what's going on: in my example, it turns out that in regular GAP, `PROGRAM_CLEAN_UP();` is *not* run at the end. Because we exit from inside `ProcessInitFiles`, before `SESSION` is called....

To expand on this, here is what happens if none of the files passed via command line uses `QUIT`: then `ProcessInitFiles` finishes, GAP proceeds to process `init.g` and eventually invokes `SESSION()`. This invokes GAP's `Shell` command, which eventually ends due to `quit` or `QUIT` being called; the latter sets `STATE(UserHasQUIT) = 1`. The crucial point is that at the end, `Shell` resets `STATE(UserHasQUIT)` to 0. Which means that GAP continues to process `init.g`, and so it executes the line after `SESSION();`, which is `PROGRAM_CLEAN_UP();`

If however one of the command line files does `QUIT`, it also sets `STATE(UserHasQUIT) = 1`. But now we are *not* inside a `Shell`, so this never is reset, and so GAP aborts processing `init.g`. I think technically this is a bug in GAP, as I am pretty sure we really would prefer if `PROGRAM_CLEAN_UP();` was called then, too (@ThomasBreuer: agreed?). So we should report and fix it there.

Anyway, in GAP.jl, we always invoke `PROGRAM_CLEAN_UP();`. But in the above situation, `STATE(UserHasQUIT)` still is non-zero. Unfortunately that means the next time a function call is executed, we run into this kernel code:
```C
    if (STATE(UserHasQuit) || STATE(UserHasQUIT)) {
        // the function must have called READ() and the user quit from
        // a break loop inside it
        GAP_THROW();
    }
```
The `GAP_THROW();` then invokes our `ThrowObserver` and so we end up getting a (false) report about an error being thrown.

So one immediate fix would be to reset `UserHasQUIT` to 0.

ThomasBreuer:
Concerning the question from

> If however one of the command line files does QUIT, it also sets STATE(UserHasQUIT) = 1. But now we are not inside a Shell, so this never is reset, and so GAP aborts processing init.g. I think technically this is a bug in GAP, as I am pretty sure we really would prefer if PROGRAM_CLEAN_UP(); was called then, too (@ThomasBreuer: agreed?). So we should report and fix it there.

Yes, I agree:
This behaviour should not depend on whether the input file is read from the command line or later in the GAP session. (For example, `/tmp/gap.sh -b -A < foo.g > /dev/null; echo $?` does not run into the errors.)

PR #757, opened 2021-11-11
Hack the typename of GAP.GapObj

### LINKED ISSUE #756: `GAP_jll.MPtr` is still visible
Apparently pull request #665 was not good enough:
```
julia> L = [ GAP.evalstr( "()" ) ]
1-element Vector{GAP.GapObj}:
 GAP: ()

julia> repr(L)
"GAP_jll.MPtr[GAP: ()]"

julia> show(L)
GAP_jll.MPtr[GAP: ()]
```
Background: `Base.show_vector` calls `Base.typeinfo_prefix(stdout, L)`, which then eventually fetches `GAP.GapObj.name`.
(The original idea in #665 had been to modify this name, but this is of course an ugly hack.)

Note that `Base.repr` is used by `View` and `Print` on the GAP side.
```
julia> GAP.prompt()
gap> Julia.L;
<Julia: GAP_jll.MPtr[GAP: ()]>
gap> Print( Julia.L );
GAP_jll.MPtr[GAP: ()]
```

fingolfin:
... to ensure uniform printing

Resolves #756

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -207,6 +207,11 @@ function __init__()
         windows_error()
     end
 
+    # HACK HACK HACK: ensure the type GapObj (which is just an alias
+    # for `GAP_jll.MPtr`) is printed as ours
+    GapObj.name.module = GAP
+    GapObj.name.name = :GapObj
+
     # regenerate GAPROOT if it was removed
     if !isdir(GAPROOT) || isempty(readdir(GAPROOT))
         Setup.regenerate_gaproot(GAPROOT)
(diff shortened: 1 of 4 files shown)


ThomasBreuer (approved):
The fix is essentially what the first commit for pull request #665 had proposed. Apparently the Julia gurus don't like it, but I see no better way to solve the problem.

fingolfin:
Thanks @ThomasBreuer I wasn't aware (resp. forgot) about that, thanks for the reminder. So let's go this way for now, there is no real good alternative for now.

PR #761, opened 2021-11-23
Update docstrings for GAP.Packages functions

fingolfin:


src/packages.jl
```diff
@@ -87,11 +86,10 @@ function install(spec::String; interactive::Bool = true, pkgdir::AbstractString
 end
 
 """
-    update(spec::String; quiet::Bool = false)
+    update(spec::String; interactive::Bool = true, pkgdir::AbstractString = DEFAULT_PKGDIR, quiet::Bool = false)
```
ThomasBreuer, line 89: Perhaps write `GAP.Packages.DEFAULT_PKGDIR`, since the variable is not exported?

src/packages.jl
```diff
@@ -123,19 +121,19 @@ end
 # thus nec. to start Julia anew
 
 """
-    remove(spec::String; quiet::Bool = false)
+    remove(spec::String; interactive::Bool = true, pkgdir::AbstractString = DEFAULT_PKGDIR, quiet::Bool = false)
```
ThomasBreuer, line 124: Perhaps write `GAP.Packages.DEFAULT_PKGDIR`, since the variable is not exported?

ThomasBreuer (approved):
Yes, documenting all arguments is an improvement.
If `interactive` gets documented, should we say that it means to pass this value to the `PackageManager` function in question?
(The meaning is quite subtle, `InstallPackage` updates an installed package only if `interactive` is `true` and the user confirms that the package shall get updated.)

fingolfin:
@ThomasBreuer updated, please have another look

ThomasBreuer:
@fingolfin Very good, thanks.

PR #763, opened 2021-12-02
Remove `Base.big(obj::GapObj)`

fingolfin:


CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 0.7.3-DEV (released 2021-XYZ)
 
+- Remove `Base.big(obj::GapObj)`: nothing was using it and it does not really
+  fit conceptually into this package.
```
ThomasBreuer, line 6: This depends on what GAP.jl means. If it is a service package for Oscar then `big` is not interesting, if it is a low level interface between GAP and Julia then the situation is perhaps different.
(I am not arguing against removing the `big` method.)

PR #767, opened 2021-12-29
Restore compatibility with Julia nightlies

fingolfin:


fingolfin:
See also https://github.com/gap-system/gap/pull/4722

PR #768, opened 2022-01-01
Better banner suppression logic

### LINKED ISSUE #765: Silence banner of GAP.jl when loaded from another package
I am developing a package which uses GAP.jl. I would like to silence the banner of GAP from within this package, and I do not know how to do that. For example, writing in the REPL
```
julia> ENV["GAP_PRINT_BANNER"] = "false"
"false"

julia> using MyPackage
```
works as intended. However, I cannot seem to do the same when writing it inside the package.

How can I silence the banner from within my package?

fingolfin:
Borrowed from Hecke

Fixes #765 (hopefully)

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -236,10 +236,14 @@ function __init__()
                     end
                 end)
 
+    # Check if were loaded from another package
+    bt = Base.process_backtrace(Base.backtrace())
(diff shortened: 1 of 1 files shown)


albinahlback:
Yes, it works! Thank you, Max!

PR #773, opened 2022-01-20
Improve conversion code allow return type inference

fingolfin:
This in turn helps code using the conversion functions


fingolfin:
@ThomasBreuer if you have any concerns about this, let me know. Otherwise I'll merge it soon. I have a follow up PR in the works which resolves some issues with conversion -- this PR here is mainly about type stability and thus performance *of the calling code*. I'll deal with improving performance here separately.

ThomasBreuer (approved):
Looks good.

PR #774, opened 2022-01-21
fixed `@gapattribute`

ThomasBreuer:
In the construction of setter and tester, we have to distinguish the Julia object (called `juliaarg`) and the corresponding GAP object (called `gaparg`).
The two names are equal in the test example, but in typical Oscar situations,
the GAP object corresponding to the Julia object `G` is `G.X`.

Perhaps the problem has not been observed before because setters and testers are not really used in Oscar.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -288,7 +288,7 @@ macro gapattribute(ex)
 
         Return `true` if the value for `$($julianame)(x)` has already been computed.
         """
-        @gapwrap $testername(x) = GAP.Globals.$gaptester(x)::Bool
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
Yes, I will do that.

fingolfin (approved):

fingolfin:
Tests fail :-(

ThomasBreuer:
This is funny:
The tests themselves passed, but they failed when they were enclosed in a `@testset`.
Moving an assignment in front of `@testset` seems to work.

PR #777, opened 2022-01-24
WIP: rewrite `gap_to_julia`

fingolfin:
TODO:
- add more test cases esp. for the various fixes
  e.g. converting the same (identical) objects to multiple
  output types should be working correctly now
- identify issues this patch resolves
- update docs
- update CHANGES,md
- ...


src/gap_to_julia.jl
```diff
@@ -8,24 +7,60 @@ end
...
 ## Conversion from GAP to Julia
 """
-    gap_to_julia(type, x, recursion_dict::Union{Nothing,RecDict}=nothing; recursive::Bool=true)
+    gap_to_julia(type::Type, x::Any; recursive::Bool = true)
+
+FIXME: update this text
+TODO: document that `type` must be a `Any` or a concrete type, and likewise
+for its type parameters. This means that `Tuple{Any}` is not allowed, nor
+are unions. Indeed, for a union, how should we determine which of the various
+union types should be the end result of the conversion?
```
ThomasBreuer, line 46: I do not understand this statement.
Currently we have
```
julia> GAP.gap_to_julia(GAP.evalstr("[1,Z(2),true]"))
3-element Vector{Any}:
    1
     GAP: Z(2)^0
 true
```
and
```
julia> Vector{GAP.Obj}(GAP.evalstr("[1,Z(2),true]"))
3-element Vector{Union{Bool, GAP.FFE, GAP.GapInt}}:
    1
     GAP: Z(2)^0
 true
```
Shall the latter be forbidden?

src/gap_to_julia.jl
```diff
@@ -64,269 +99,327 @@ julia> GAP.gap_to_julia( Matrix{Int}, val )
...
+_needs_conversion_tracking(::Type{T}) where T = !isbitstype(T)
+
+# the only way conversion produces a GapObj is if the input was a GapObj, and
+# then we just use the same GapObj, so duplicates are automatically taken care of
+# and we don't need tracking
+_needs_conversion_tracking(::Type{GapObj}) = false
+
+# Union types need tracking if any of the types the union was formed over need it
+#_needs_conversion_tracking(T::Union) = any(_needs_conversion_tracking, Base.uniontypes(T))
+# TODO: alternatively just handle GAP.GapInt, GAP.Obj
```
ThomasBreuer, line 211: As ar as I understand, in line 46 above, it is stated that union types are not allowed.

src/gap_to_julia.jl
fingolfin: So we already discussed this on gather, but I want to record what I said here for others / myself in the future:

Indeed the text is not phrased correctly. What I wanted to express is that I don't see how to handle a request like this properly:
`Vector{Union{String,Int,Symbol}}(GAP.evalstr("[1, \"x\"]"))`

In principle this could be dealt with, in that there are sensible conversion results of the requested type for this, namely `[1, "x"]` and `[1, :x]`. But then we need to decide which of them to return? Perhaps do it arbitrarily? Or introduce some kind of priority between certain types (e.g "if both `Symbol` and `String` are requested and possible, prefer `Symbol`"), but then how to extend that to all cases (We could be asked to produce a `Vector{Union{Vector{Vector{Int}}, Matrix{Int}, fmpz_mat, fmpq_mat}` ...).

In contrast, conversion to `Vector{GAP.Obj}`  above works, despite `GAP.Obj` being a union type, because all the elements of the GAP list already had a type in that Union. So in that case the "trivial conversions" rule applies, i.e., asking to convert an object `x::T` to type `S` always succeeds and returns `x` whenever `S` is either `T` or a supertype of `T`.

PR #781, opened 2022-02-09
Add `quiet` kwarg to `Packages.load`

fingolfin:
This is useful for debugging among other things.

Also reformat some docstrings so that they render nicer in the HTML
manual (and don't require the user to scroll horizontally to read them)

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -21,7 +21,7 @@ function init_packagemanager()
 end
 
 """
-    load(spec::String, version::String = ""; install = false)
+    load(spec::String, version::String = ""; install::Bool = false, quiet::Bool = true)
 
 Try to load the newest installed version of the GAP package with name `spec`.
 Return `true` if this is successful, and `false` otherwise.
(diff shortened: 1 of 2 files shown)


src/packages.jl
```diff
@@ -32,19 +32,22 @@ the package banner is not printed.
 If `install` is set to `true` and the required GAP package is not yet
 installed then [`install`](@ref) is called first, in order to install
 the newest released version of the package.
+
+If `quiet` is set to `true` then package banners are shown for all packages
```
ThomasBreuer, line 36: "is set to `false`" ?

src/packages.jl
```diff
@@ -32,19 +32,22 @@ the package banner is not printed.
...
     gspec = GapObj(spec)
     gversion = GapObj(version)
-    loaded = Globals.LoadPackage(gspec, gversion, false)
+    loaded = Globals.LoadPackage(gspec, gversion, !quiet)
     if loaded == true
         return true
     elseif install == true
         # Try to install the package, without showing messages.
-        if Packages.install(spec; interactive = false, quiet = true)
+        if Packages.install(spec; interactive = false, quiet)
```
ThomasBreuer, line 48: `Packages.install(spec; interactive = false, quiet = quiet)`?

ThomasBreuer (commented):
just two small comments

src/packages.jl
fingolfin: Indeed.

src/packages.jl
fingolfin: That's equivalent; Julia 1.6 allows replacing kwargs of the form `foo = foo` by just `foo`

ThomasBreuer: O.k.
(I do not like this syntax. The left hand side of `foo = foo` is a key, the right hand side is a variable.
When one copies such a code line into another context and adjusts the variable name then one gets into trouble.)

PR #782, opened 2022-02-09
Replace some calls to low-level GAP functions

fingolfin:
... by high-level synonyms, which are easier to understand, and also
stable under changes to the GAP kernel.

### DIFF
--- a/docs/src/other.md
+++ b/docs/src/other.md
@@ -40,9 +40,10 @@ GAP operations.
 |--------------|----------|
 | `length`     | `Length` |
 | `in`         | `\in`    |
-| `zero`       | `ZERO`   |
-| `one`        | `ONE`    |
-| `-` (unary)  | `AINV`   |
+| `zero`       | `ZeroSameMutability`   |
+| `one`        | `OneSameMutability`    |
+| `-` (unary)  | `AdditiveInverseSameMutability`   |
+| `inv`        | `InverseSameMutability`    |
 | `+`          | `SUM`    |
 | `-` (binary) | `DIFF`   |
 | `*`          | `PROD`   |
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -217,10 +217,10 @@ Base.hasproperty(x::GapObj, f::Union{AbstractString,Int64}) =
     Wrappers.ISB_REC(x, RNamObj(f))
 
 #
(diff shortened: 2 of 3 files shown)


docs/src/other.md
```diff
@@ -40,9 +40,10 @@ GAP operations.
 |--------------|----------|
 | `length`     | `Length` |
 | `in`         | `\in`    |
-| `zero`       | `ZERO`   |
-| `one`        | `ONE`    |
-| `-` (unary)  | `AINV`   |
+| `zero`       | `ZeroSameMutability`   |
+| `one`        | `OneSameMutability`    |
+| `-` (unary)  | `AdditiveInverseSameMutability`   |
+| `inv`        | `InverseSameMutability`    |
```
fingolfin, line 46: ... but perhaps we should instead switch the mapping:
```suggestion
| `zero`       | `Zero`   |
| `one`        | `One`    |
| `-` (unary)  | `AdditiveInverse`   |
| `inv`        | `Inverse`    |
```
As that probably avoids further pain points. I am not sure if there is any good reason for mapping to the "same mutability" variants? Maybe @ThomasBreuer can think of any?

If we do such a switch, then I'd consider it a breaking change (i.e. for GAP.jl 0.8) as it at least in principle could break some existing code...

ThomasBreuer: The "same mutability" variants had been chosen for the sake of consistency:
GAP's binary arithmetic operations return immutable results iff both arguments are immutable, and mutable results if not.
Thus -if we have to choose one behaviour for a unary operation- it is reasonable to define the same also for the unary arithmetic operations.

ThomasBreuer: (I had yesterday tried to add a comment here, but now I do not find it anymore. Strange. Anyhow, here it is again.)

In GAP, binary arithmetic operations return immutable results if and only if the two arguments are immutable.
Thus it is consistent if unary arithmetic operations return a result with the same mutability as the argument.
Of course, GAP has several variants of these unary operations, regarding the mutability of the result, but here we have to choose one behaviour.

fingolfin: OK, fair enough.

So, shall we merge this as-is?

ThomasBreuer (approved):

docs/src/other.md
ThomasBreuer: Yes.

Issue #801, opened 2022-04-11
Update `GAP_pkg_juliainterface` for latest Julia 1.9


fingolfin:
In tests against Julia nightly, we now get an error `undefined symbol: jl_set_global`. That probably means we need to update `libjulia_jll` and then rebuild `GAP_pkg_juliainterface`.

ThomasBreuer:
I had thought that we can solve this problem by providing a GAP function `JuliaSetVal` that calls `setglobal!` if this is available, and calls `ccall(:jl_set_global, ...)` otherwise.
However, this seems to be tricky.
All combinations of GAP code and Julia code which I tried end up in a crash of Julia 1.6.5 or Julia 1.9.0-DEV, depending on which one had precompiled GAP.jl before.
I understand that the code that decides which variant shall be used cannot be precompiled.
Trying to define the Julia function in question inside the `__init__` function of GAP.jl was not successful.
Making the decision at runtime in GAP code does not work because I cannot call `ccall` from GAP: `ccall` is a keyword not a function.
(Well, using `JuliaEvalString` *would* work, if one binds the intended value to some global GAP variable in advance. Is this perhaps the simplest solution?)

fingolfin:
> crashing [...] depending on which one had precompiled GAP.jl before.

This sounds like issue #691. It would go away once we update the JLL to match any changes we may make in the C sources of JuliaInterface. However, we should also try to fix it "properly", as this is of course highly annoying.

If my guess is right, then to fix the crashes you "just" need to delete the compiled version of JuliaInterface before switching to another Julia version. In other words: your code may have been perfectly fine, the problem is in the compiled C code.

That said, it is difficult to say what exactly is the problem without seeing the code. So feel free to open a draft PR with what you have, and we can then discuss more there.

fingolfin:
Resolve by #811

PR #802, opened 2022-04-13
replace the GAP kernel function `JuliaSetVal` ...

ThomasBreuer:
... by a GAP function.
This addresses issue #801.

### DIFF
--- a/pkg/JuliaInterface/gap/utils.gi
+++ b/pkg/JuliaInterface/gap/utils.gi
@@ -4,6 +4,25 @@
 ##
 #############################################################################
 
+BindGlobal( "JuliaSetVal",
+function( name, val )
+  if not IsString( name ) then
+    Error( "JuliaSetVal: <name> must be a string (not the value '", name, "')" );
+  elif Julia.isdefined( Julia.Core,
+         Julia.Symbol( GAPToJulia( "setglobal!" ) ) ) then
+    # Julia 1.9 introduces `setglobal!`.
+    JuliaFunction( "setglobal!" )( Julia.Main,
+        Julia.Symbol( GAPToJulia( name ) ), val );
+  else
+    # `jl_set_global` is available up to Julia 1.8.
+    GAPInfo.JuliaValue:= val;
+    JuliaEvalString( Concatenation(
+        "ccall(:jl_set_global, Cvoid, (Any, Any, Any), Main, Symbol(\"", name,
+        "\"), GAP.Globals.GAPInfo.JuliaValue)" ) );
+    Unbind( GAPInfo.JuliaValue );
+  fi;
+end );
+
 ## Create a record from key value lists
 BindGlobal( "CreateRecFromKeyValuePairList",
   function( keys, vals )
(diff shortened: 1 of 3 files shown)


pkg/JuliaInterface/gap/utils.gi
```diff
@@ -4,6 +4,25 @@
...
+        Julia.Symbol( GAPToJulia( name ) ), val );
+  else
+    # `jl_set_global` is available up to Julia 1.8.
+    GAPInfo.JuliaValue:= val;
+    JuliaEvalString( Concatenation(
+        "ccall(:jl_set_global, Cvoid, (Any, Any, Any), Main, Symbol(\"", name,
+        "\"), GAP.Globals.GAPInfo.JuliaValue)" ) );
+    Unbind( GAPInfo.JuliaValue );
+  fi;
+end );
```
fingolfin, line 24: I would suggest to instead define a Julia helper function which is called here. That then can use `@static if` so that only the version of the code that works in the given Julia version is ever compiled. Something like this ought to work:
```julia
function _setglobal!(M::Module, name::Symbol, val::Any)
  @static if isdefined(Core,:setglobal!)
    Core.setglobal!(M, name, val)
  else
    ccall(:jl_set_global, Cvoid, (Any, Any, Any), M, Symbol(name), val)
    return val
  end
end
```

pkg/JuliaInterface/gap/utils.gi
```diff
@@ -4,6 +4,16 @@
 ##
 #############################################################################
 
+BindGlobal( "JuliaSetVal",
+function( name, val )
+  if not IsString( name ) then
+    Error( "JuliaSetVal: <name> must be a string (not the value '", name, "')" );
+  else
+    JuliaFunction( "_setglobal!", "GAP" )( Julia.Main,
+        Julia.Symbol( GAPToJulia( name ) ), val );
```
fingolfin, line 13: This also works, as we've installed a `Symbol` constructor for GAP strings:
```suggestion
    JuliaFunction( "_setglobal!", "GAP" )( Julia.Main,
        Julia.Symbol( name ), val );
```

We also have a dedicated function for this, which is faster; not that it matters much here, it is mostly for `_JL_RNAM_TO_JULIA_SYMBOL`.
```suggestion
    JuliaFunction( "_setglobal!", "GAP" )( Julia.Main,
        JuliaSymbol( name ), val );
```

fingolfin (approved):

pkg/JuliaInterface/gap/utils.gi
```diff
@@ -4,6 +4,16 @@
 ##
 #############################################################################
 
+BindGlobal( "JuliaSetVal",
+function( name, val )
+  if not IsString( name ) then
+    Error( "JuliaSetVal: <name> must be a string (not the value '", name, "')" );
+  else
+    JuliaFunction( "_setglobal!", "GAP" )( Julia.Main,
```
fingolfin, line 12: Slightly more efficient:
```suggestion
    Julia.GAP.("_setglobal!")( Julia.Main,
```
Even better: rename that function to e.g. `_setglobal` (without the "!"), then we can write
```suggestion
    Julia.GAP._setglobal( Julia.Main,
```
I guess at this point one could also put this call into a single line :-)

ThomasBreuer:
Yes.
My original idea was to move everything to the GAP side and to call the Julia functions at runtime, thus avoiding the question whether something gets precompiled or not, but as far as I understand, `@static` is the right answer to this problem.

PR #803, opened 2022-04-14
Allow `Julia.MODULE.x := 1` in GAP code, remove `JuliaSetVal`

### LINKED ISSUE #798: Allow assigning to Julia module variables, at least in Julia's `Main` module?
Consider this:
```
julia> GAP.prompt()
gap> Julia.y := 1;
Error, Manual assignment to module is not allowed at /Users/mhorn/Projekte/OSCAR/GAP.jl/pkg/JuliaInterface/gap/JuliaInterface.gi:73 called from
<function ".:= [ IsJuliaModule, IsPosInt, IsObject ]">( <arguments> )
 called from read-eval loop at *stdin*:1
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk>
```

It would be nice if that just worked.

In Julia 1.9, this can actually be done using `setproperty!`:
```
julia> setproperty!(Main, :x, 1)
1

julia> x
1

julia> module Foo const x = 1 end
Main.Foo

julia> Foo.x
1

julia> Foo.y = 2
2

julia> Foo.y
2
```

For older Julia versions, we may be able to at least hack something for `Julia.Main`

fingolfin:
Resolves #798

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 - Upgrade to newer GAP snapshot
 - Give helpful error if `GAP_jll` is not available
+- Allow assigning Julia variables in arbitrary Julia modules from GAP.
+  For example `Julia.x := 1` sets the variable `x` in the `Main` module,
+  as does `Julia.Main.x := 1`.
 - Change `@gapattribute` to use prefixes `has_` and `set_` instead of
   `has` and `set`. So for example for `@gapattribute isfinite(obj) = ...`
   the tester function now is `has_isfinite` instead of `hasisfinite` and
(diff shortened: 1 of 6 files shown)


ThomasBreuer (approved):
Technically, this solution is simpler than what we had before.
I am not sure whether assigning variables in other Julia modules than `Main` is a good idea.

fingolfin:
> Technically, this solution is simpler than what we had before.
> I am not sure whether assigning variables in other Julia modules than `Main` is a good idea.

I think it's a very bad idea in general! Alas, the Julia devs decided to allow it via the same `SomeModule.var = value` syntax in >= 1.9, too, so ... *shrug*. I mean, we *could* artificially restrict it to only allow assigning in `Main`, but that would require extra checks, and so all in all I decided to just go with the flow and imitate what Julia does / will do

ThomasBreuer:
@fingolfin This argument strikes: If the Julia people provide the tool to set variables in arbitrary modules, we should not be concerned with adding tests that restrict this.

PR #805, opened 2022-04-14
Remove OverrideTypeNameAndModule hack

fingolfin:
This means that type `GapObj` is no longer printed as `GapObj` resp.
`GAP.GapObj` (depending on context) but rather as `GAP_jll.GapObj`.

But it avoids a deep hack in the Julia kernel that could cause lots
of pain for us in the future.

Closes #804 (for which this PR is an alternative)

### DIFF
--- a/pkg/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/JuliaInterface/src/JuliaInterface.c
@@ -157,35 +157,6 @@ void ResetUserHasQUIT(void)
 }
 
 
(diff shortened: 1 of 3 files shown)


fingolfin:
I discussed this with @ThomasBreuer and we'll go this way: it's just not worth the ongoing hassle to try to keep this "feature" (i.e., slightly "nicer" (???)) printing of the `GapObj` type working.

PR #806, opened 2022-04-20
Update to latest GAP_jll / GAP_lib_jll

fingolfin:
This includes https://github.com/gap-system/gap/pull/4863 which removes the problematic `waitpid` calls. Let's see if that helps

CC @benlorenz

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,8 +17,8 @@ REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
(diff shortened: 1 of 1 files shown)


benlorenz:
The GAP.jl tests already ran successfully on macos nightly here (with the new jll versions): https://github.com/oscar-system/GAP.jl/runs/6092924646?check_suite_focus=true#step:6:308
The Oscar tests won't run because it is not compatible with GAP.jl 0.8 (yet?).

fingolfin:
Great!

Yes, 0.8 is incompatible. So the plan is to release it soon and then submit an Oscar.jl PR to match, and finally make a new Oscar release

PR #820, opened 2022-09-02
install a `GAP.Globals.Download` method if possible

ThomasBreuer:
If GAP knows the global list `Download_Methods` then add a new `Download` method in the first place, such that Julia's `Downloads.download` is used.
This way, we do no longer rely on whether or not some GAP package or external program is available that provides download functionality.

This pull request should not be merged at the moment because it relies on gap-packages/utils/pull/48, which may change before it gets merged.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -35,13 +35,18 @@ function init_packagemanager()
            :name => "via Julia's Downloads.download",
            :isAvailable => Globals.ReturnTrue,
            :download => function(url, opt)
-             # exception handling is omitted by concept, i.e. errors occuring during the download are shown to the user
-             if hasproperty(opt, :target)
-               Downloads.download(String(url), String(opt.target))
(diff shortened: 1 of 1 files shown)


src/packages.jl
```diff
@@ -18,6 +22,38 @@ function init_packagemanager()
...
+           :download => function(url, opt)
+             try
+               if hasproperty(opt, :target)
+                 Downloads.download(String(url), String(opt.target))
+                 return GapObj(Dict{Symbol, Any}(:success => true), recursive=true)
+               else
+                 buffer = Downloads.download(String(url), IOBuffer())
+                 return GapObj(Dict{Symbol, Any}(:success => true, :result => String(take!(buffer))), recursive=true)
+               end
+             catch(e)
```
fingolfin, line 46:
```suggestion
             catch e
```

Issue #823, opened 2022-09-13
Manual (and/or docstrings) should list all built in conversions between GAP and Julia


fingolfin:
We have a table with an overview in the JuliaInterface manual in <https://oscar-system.github.io/GAP.jl/stable/assets/html/JuliaInterface/chap3_mj.html> but it is *not* in the GAP.jl manual.

There are probably more things discussed in one manual but not the other...

PR #824, opened 2022-09-14
extend the `julia_to_gap` documentation

ThomasBreuer:
by the explicit list of types for which conversions are provided in GAP.jl (the list which was already in the JuliaInterface manual)

This addresses issue #823.

### DIFF
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -24,6 +24,24 @@ julia> GAP.julia_to_gap([ [1, 2], [3, 4]], recursive = true)
 GAP: [ [ 1, 2 ], [ 3, 4 ] ]
 
 ```
+
+The following `julia_to_gap` conversions are supported by GAP.jl.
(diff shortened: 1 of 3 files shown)


src/julia_to_gap.jl
```diff
@@ -24,6 +24,24 @@ julia> GAP.julia_to_gap([ [1, 2], [3, 4]], recursive = true)
 GAP: [ [ 1, 2 ], [ 3, 4 ] ]
 
 ```
+
+The following `julia_to_gap` conversions are supported by GAP.jl.
```
fingolfin, line 28: What about the other direction? Are we actually "symmetric" in what we support? Might be good to state, too?

ThomasBreuer: Yes, of course the other direction makes sense. If you think the current proposal is more or less o.k. for `julia_to_gap` then I will update it to include also the other direction.

fingolfin (approved):

PR #825, opened 2022-09-29
Use Scratch.jl for GAP root

### LINKED ISSUE #728: use Scratch.jl and/or RelocatableFolders.jl for gaproot, other stuff
- https://github.com/JuliaPackaging/Scratch.jl
- https://github.com/JuliaPackaging/RelocatableFolders.jl

### LINKED ISSUE #822: Move `GAPROOT` out of package directory
Currently we create a symlink tree inside the GAP.jl package directory, inside `PKDIR/gaproot/v1.6/` and so on (depending on the Julia version).

This is problematic in shared Julia installation where the GAP.jl package directory might be read-only. In general, the Julia folks would like to move to a system where package directories (at least for release versions) are read only. And I think this makes a lot of sense, too.

Fixing this should essentially "just" require adjusting
```
const GAPROOT = abspath(@__DIR__, "..", "gaproot", "v$(VERSION.major).$(VERSION.minor)")
```
in `src/gap.jl` to a different path. Ideally it should probably be using [Scratch.jl](https://github.com/JuliaPackaging/Scratch.jl).

(A secondary objective would be to also adjust `build_JuliaInterface` to build outside of the package directory; but this code should normally only matter for dev builds where the C sources were modified, so it's not important for now).

One concern lingers: the report motivating this issue was performed on Kaiserslautern computer servers, with a shared Oscar install, with a custom Julia "depot". I should verify that `Scratch.jl` works correctly in that setup, i.e., it returns a path in the user's depot (in particular: it must be writable). 

CC @fieker @ThomasBreuer

fingolfin:
This also forces a full regeneration of our custom GAP root dir each time the package loads, instead of only during precompilation. This adds ~130 milliseconds startup overhead for me, of which 90% are spent in finding working C/C++ compilers... Not great, but with Scratch.jl any of the directories could move, so we kinda have to do it this way, as far as I can tell (Polymake.jl also does it this way). Perhaps I could at least cache the C/C++ compiler values somehow, but for now I just want to have this working.

Resolves #822
Resolves #728

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -15,12 +15,14 @@ Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 
 [compat]
 GAP_jll = "~400.1192.001"
 GAP_lib_jll = "~400.1192.002"
 GAP_pkg_juliainterface_jll = "=0.800.000"
 MacroTools = "0.5"
+Scratch = "1.1"
 julia = "1.6"
 
 [extras]
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -25,19 +25,18 @@ GAP_jll.is_available() ||
 
 include("setup.jl")
 
-# always force regeneration of GAPROOT when precompiling
-const GAPROOT = abspath(@__DIR__, "..", "gaproot", "v$(VERSION.major).$(VERSION.minor)")
-Setup.regenerate_gaproot(GAPROOT)
-
 import Base: length, finalize
 import Libdl
 import Markdown
 import Random
 
-include("types.jl")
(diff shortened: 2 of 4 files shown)


fingolfin:
This introduces a minor regression which I have not yet understood: upon `using GAP`, if GAP was already precompiled, all is fine; but if GAP gets precompiled then we don't show the GAP banner anymore. Huh?

ThomasBreuer:
@fingolfin I cannot reproduce the problem with the GAP banner.
I have checked out this branch and called `Pkg.resolve()`.
After starting Julia and calling `using GAP`, the GAP banner is shown in my Julia session, no matter whether the message `Info: Precompiling GAP` is shown or not.

ThomasBreuer (approved):
Looks good. Thanks.

fingolfin:
@ThomasBreuer which Julia version are you using? For me the problem of the missing banner only occurs in Julia 1.8 -- and indeed, the hack we use (I think perhaps invented by @thofma ?) checks for a function `_tryrequire_from_serialized` -- and starting with Julia 1.8 this function is called in more places. So I am guessing this code may need to be adapted to work for Julia 1.8.

thofma:
Yes, this was my doing. I can have a look.

thofma:
I would suggest you merge this here and I will fix it in a followup PR.

PR #844, opened 2022-11-24
Adjust ccall type arguments

fingolfin:
These are necessary for my experiments with turning GapObj into a
struct. But they seem to work in general.


fingolfin:
I'd like to perform some timing experiments to check whether this has any negative (or positive?) effects on timings

PR #845, opened 2022-11-25
Use `jl_reinit_foreign_type` if available

fingolfin:
This prepares us for an upcoming change to serialization in Julia 1.10, see issue #846.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## Version 0.9.2-DEV (released YYYY-MM-DD)
 
+- Silence scary warning about missing compiler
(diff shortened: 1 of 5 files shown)


fingolfin:
See also  https://github.com/gap-system/gap/pull/5224

ThomasBreuer (approved):
As far as I see, these changes will work as soon as the changes from gap-system/gap/pull/5224 are available.

fingolfin:
Note that the code in this PR is written in such a way that it *should* also work now, before those changes on the GAP and Julia side -- this is required so we stay compatible with Julia <= 1.9.

Issue #627, opened 2021-02-24
improve `Packages.install` and `Packages.load`


ThomasBreuer:
As soon as GAP's PackageManager supports a version number as an optional argument of `InstallPackage` (see gap-packages/PackageManager/issues/60), we should change `GAP.Packages.install` and `GAP.packages.load` accordingly.

(Today I ran into the "undefined symbol" problem with the ferret package (see oscar-system/Oscar.jl/issues/262):
Version 1.0.3 was already installed, and in this situation `GAP.packages.load` has currently no chance to force installing version at least 1.0.5. Note that `GAP.Packages.install` is called in a non-interactive way, and then `InstallPackage` does not try to upgrade.)

Once this functionalit is available, the `packages.load` call in `Oscar.jl/experimental/GaloisGrp/GaloisGrp.jl` can be changed to `GAP.Packages.load("ferret", "1.0.5", install = true)`.

PR #847, opened 2022-11-29
admit prescribing a package version ...

ThomasBreuer:
... in `GAP.Packages.install`.
This is meanwhile supported by GAP's PackageManager package.

The most interesting application is perhaps `GAP.Packages.load` with argument `install = true`, because now we can ask for loading a good enough version of a perhaps previously not installed package, and get `false` if this is not possible.

(Also fixed now: `GAP.Packages.load` had returned `fail` in some situations, now `false` is returned in these cases.)

Resolves issue #627.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 - Silence scary warning about missing compiler
 - Prepare for an upcoming change to serialization in Julia 1.10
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Looks good to me.

Could you also add an entry to the CHANGES file?

Issue #740, opened 2021-10-19
Conversions between GAP and Julia: What do we really want?


ThomasBreuer:
After some experiments that were inspired by issue #736, I think that the relation between Julia's constructors and `gap_to_julia` is more complicated than what we have at the moment.
Consider the following object `L`.
```
julia> using Oscar;  using GAP;  x = fmpz(10)^20;  L = GAP.evalstr("List([1..10], i -> Julia.Main.x + i)");
```
This is a GAP list containing Julia objects. Now we want to convert this to a Julia array of GAP objects. The following works.
```
julia> [GapObj(x) for x in L]
10-element Vector{GapObj}:
 GAP: 100000000000000000001
 GAP: 100000000000000000002
[...]
```
The corresponding constructor does not work.
```
julia> Vector{GapObj}(L)
ERROR: Don't know how to convert value of type fmpz to type GapObj
Stacktrace:
 [1] gap_to_julia(t::Type{GapObj}, x::fmpz)
[...]
```
The error message looks irritating: Oscar installs a method for converting `fmpz` to `GapObj`, see above.

The reason for the problem is that the constructor delegates to `gap_to_julia` with first argument `Vector{GapObj}` and with the wish to convert recursively; the method in question creates a `Vector{GapObj}` and wants to fill it with the results of `gap_to_julia(GapObj, x)` calls --but of course `julia_to_gap` would be the correct direction.
(And note that calling `julia_to_gap(L)` would not do what we want.)

My interpretation is that currently `gap_to_julia` promises a conversion in the direction from GAP to Julia, perhaps recursive (and then in this direction also for the subobjects), and if a Julia type is given as the first argument then this serves as a hint which conversion is intended but not as information about the direction of the conversion.
On the other hand, the constructor `Vector{GapObj}` promises a conversion to a `Vector` of `GapObj`s, which is impossible by delegation to the current `gap_to_julia`.

So what do we want:
change the recursions in `gap_to_julia` methods such that examples like the above one work,
separate the implementation of `gap_to_julia` from constructors where necessary, or
"reverse" the implementation such that `gap_to_julia` calls constructors, and not the other way round?

PR #849, opened 2022-11-30
admit `gap_to_julia` conversion to `Vector{GAP.Obj}`

ThomasBreuer:
This addresses issue #740, but I am still not sure what would be a good solution.

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -186,7 +186,11 @@ function gap_to_julia(
             end
             if recursive && !isbitstype(typeof(current_obj))
                 new_array[i] = get!(recursion_dict, current_obj) do
+                  if T == GapObj || T == Obj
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
The point is that `Vector{GapObj}(L)` for a GAP list `L` of Julia objects starts with `gap_to_julia`.
In the recursion, we want to convert the list entries to `GapObj`, which means that we have to switch to `julia_to_gap`.
A "natural place" for expressing this generically would be in `src/gap_to_julia`, where the conversion of `GAP.GapObj` and `GAP.Obj` for recursion situations is handled, but I do not see how to formulate in general what one wants.
(Perhaps the idea to specify a "direction" of conversion as in `gap_to_julia` and `julia_to_gap` is not a good idea at all.)

fingolfin:
I'd prefer to finish PR #777 which should also deal with this.

I'll try to prioritize that next week. We'll see.

PR #857, opened 2023-02-22
do not mention `fmpz` in the JuliaInterface manual

ThomasBreuer:


pkg/JuliaInterface/gap/JuliaInterface.gd
```diff
@@ -27,12 +27,11 @@
 #!  Also, the interface does not provide conversions to &GAP;
 #!  for &Julia; objects whose types are defined in &Julia; packages
 #!  (that is, not in the <Q>core &Julia;</Q>).
-#!  For example, the &Julia; package <Package>Nemo.jl</Package> defines
-#!  an integer type <C>fmpz</C>.
-#!  The conversion of integers of type <C>fmpz</C> between &Julia; and &GAP;
-#!  is handled in the context of the <Package>Oscar</Package> system,
-#!  which uses both &GAP; and <Package>Nemo.jl</Package>,
-#!  but <Package>JuliaInterface</Package> does not deal with it.
+#!  For example, the &Julia; package <Package>Oscar.jl</Package> defines
```
fingolfin, line 30:
```suggestion
#!  For example, the &Julia; package <Package>Oscar.jl</Package> provides
```
Then it also encompasses `fmpz`

pkg/JuliaInterface/gap/calls.gi
```diff
@@ -5,7 +5,7 @@ BindGlobal("_JL_Dict_Any", JuliaEvalString("Dict{Symbol,Any}"));
 
 ##
 ##  We want to use &GAP's function call syntax also for certain Julia objects
-##  that are <E>not</E> functions, for example for types such as <C>fmpz</C>.
+##  that are <E>not</E> functions, for example <C>ZZ</C>.
```
fingolfin, line 8: or we could just reference a base type?
```suggestion
##  that are <E>not</E> functions, for example for types such as <C>String</C>.

```

fingolfin (approved):
OK by me. I'd consider tweaking some details, we can talk about it tomorrow (or you just merge, either with or without my suggestions)

src/utils.jl
```diff
@@ -29,21 +29,26 @@ function get_symbols_in_module(m::Module)
...
 in the latter case, `val` is set to the string of the error message.
 
 # Examples
 ```jldoctest
-julia> GAP.call_with_catch( sqrt, 2 )
+julia> GAP.call_with_catch(sqrt, 2)
 (true, 1.4142135623730951)
 
-julia> GAP.call_with_catch( sqrt, -2 )
-(false, "DomainError(-2.0, \\"sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).\\")")
```
fingolfin, line 46: It seems there is a similar problem (changes in the error in Julia nightly) inside `pkg/JuliaInterface/tst/utils.tst`

PR #860, opened 2023-03-04
Remove BuildPackages.sh

fingolfin:
Not needed anymore in latest PackageManager


ThomasBreuer:
The error message `Compilation script not found ... It should be at ...` indicates that not the latest PackageManager version (1.4.1) is used.

PR #861, opened 2023-04-05
Allow iterating over Julia objects from within GAP

fingolfin:


pkg/JuliaInterface/gap/adapter.gi
```diff
@@ -84,6 +84,44 @@ InstallOtherMethod( \.\:\=,
...
+
+BindGlobal("IsDoneIterator_Julia", function(it)
+  return Julia.isnothing(it!.state);
+end);
+
+BindGlobal("ShallowCopy_Julia", function(it)
+  return rec(NextIterator := it!.NextIterator,
+    IsDoneIterator := it!.IsDoneIterator,
+    ShallowCopy := it!.ShallowCopy,
+    state := it!.state,
```
ThomasBreuer, line 106:
```suggestion
    state := Julia.Base.deepcopy(it!.state),
```

PR #866, opened 2023-05-08
Disable REPL hack in Julia 1.10

### LINKED ISSUE #864: Warning on julia master
I just got
```
┌ GAP [c863536a-3901-11e9-33e7-d5cd0df7b904]
│  WARNING: could not import REPLCompletions.get_value into GAP
```
on
```
julia> versioninfo()
Julia Version 1.10.0-DEV.1128
Commit bc2fa503a8* (2023-04-27 05:52 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores
```

fingolfin:
Resolves #864

@ThomasBreuer turns out this is much easier than we thought ;-)

### DIFF
--- a/src/globals.jl
+++ b/src/globals.jl
@@ -55,20 +55,25 @@ end
 
 propertynames(::GlobalsType) = Vector{Symbol}(Globals.NamesGVars())
 
+@static if VERSION < v"1.10-DEV"
+  # HACK to get tab completion to work for GAP globals accessed via GAP.Globals;
+  # e.g. if the REPL already shows `GAP.Globals.MTX.Is` and the user presses
+  # TAB, they should be shown a list of members of the GAP global variable `MTX`
+  # (which is a record) starting with `Is`. The easy part for supporting this is
+  # to implement `propertynames` for `GapObj` (at least those which are GAP
+  # records). Unfortunately that's not quite enough; we also have add methods
+  # for the `get_value` method below.
+  import REPL.REPLCompletions: get_value
+  function get_value(sym::Symbol, ::GAP.GlobalsType)
+      v = _ValueGlobalVariable(sym)
+      v === C_NULL && return (nothing, false)
+      return (_GAP_TO_JULIA(v), true)
+  end
+  get_value(sym::QuoteNode, fn::GAP.GlobalsType) = get_value(sym.value, fn)
 
-# HACK to get tab completion to work for GAP globals accessed via GAP.Globals;
-# e.g. if the REPL already shows `GAP.Globals.MTX.Is` and the user presses
-# TAB, they should be shown a list of members of the GAP global variable `MTX`
-# (which is a record) starting with `Is`. The easy part for supporting this is
-# to implement `propertynames` for `GapObj` (at least those which are GAP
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
I cannot reproduce this.

I have tried the nightly builds julia-1dcac5eb40 (10 days old) and julia-4e782bf91a (of today) with the proposed change.
In both versions, entering `GAP.Globals.MTX.Is` followed by `<TAB>` does not yield any output.
With Julia 1.8.5, this input yields the expected 10 matches which I get also in the GAP session.

Why do I get a different behaviour? (Is it because of the operating system (ubuntu 20.04) or because of the terminal (xfce), or what?)

fingolfin:
Argh. No, you are right, my mistake: I only tested `GAP.Globals.IsP<TAB>` as I forgot about the "nested" tab being the actual problem. Well, I *also* did run the test suite to verify it, knowing we had tests in it for the REPL completion... and the test suite passed!!!

... because we are not actually running the REPL completion tests 😠 so I just made PR #867 to fix *that*.

I'll resume looking for a proper solution. But in the meantime I will revamp this PR into simply disabling the REPL completion & tests for 1.10, to avoid the warnings when precompiling GAP.jl; we won't be worse off than we are this way.

ThomasBreuer:
Thanks.

> But in the meantime I will revamp this PR into simply disabling the REPL completion & tests for 1.10, to avoid the warnings when precompiling GAP.jl; we won't be worse off than we are this way.

The pull request does this already, except for the comment 
> In Julia >= 1.10, the hack is not necessary anymore.

fingolfin:
Updated

ThomasBreuer (approved):
Looks good. Thanks.

aviatesk:
You can add the following patch to get reasonable completions for cases like `GAP.Globals.MTX.I|`:
```diff
diff --git a/src/globals.jl b/src/globals.jl
index 8ee3805..0b7a3ac 100644
--- a/src/globals.jl
+++ b/src/globals.jl
@@ -37,7 +37,7 @@ Main
 """
 const Globals = GlobalsType()
 
-function getproperty(::GlobalsType, name::Symbol)
+Base.@assume_effects :effect_free :nothrow :terminates_globally function getproperty(::GlobalsType, name::Symbol)
     v = _ValueGlobalVariable(name)
     v === C_NULL && error("GAP variable $name not bound")
     return _GAP_TO_JULIA(v)
```

I am not sure if the `getproperty` method holds the declared effects though (please assert it yourself if you use this patch). And if that holds, the effects declaration should probably be in `_ValueGLobalVariable` and `_GAP_TO_JULIA` since it allows a broader optimization possibilities.

fingolfin:
OK, thanks for the hint! I am a bit wary of these annotations, after past experiments with `@pure` bit us. But these now new  seem at least better documented, and after reading those comments, it seems we should indeed be able to use them here. Alas, they require Julia >= 1.8, but of course we can deal with that, too

PR #867, opened 2023-05-08
Actually run REPL completion tests

fingolfin:


ThomasBreuer (approved):
Yes, these tests should be run. (The failure with Julia nightly will remain until we have a fix for Julia >= 1.10.)

PR #869, opened 2023-05-09
compat: allow GAP_pkg_juliainterface_jll 0.800.202

benlorenz:
Not sure if this is correct, or why the previous 0.800.202 was not allowed so far.
But something like this is needed to allow the new rebuild for julia nightly, maybe the restriction can be loosened up a bit?

The nightly build will probably still fail as the new releases are not in the julia registry yet.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 [compat]
 GAP_jll = "~400.1200.101"
 GAP_lib_jll = "~400.1201.100"
-GAP_pkg_juliainterface_jll = "=0.800.201"
+GAP_pkg_juliainterface_jll = "=0.800.201, =0.800.202"
 MacroTools = "0.5"
 Scratch = "1.1"
 julia = "1.6"


Project.toml
```diff
@@ -20,7 +20,7 @@ Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 [compat]
 GAP_jll = "~400.1200.101"
 GAP_lib_jll = "~400.1201.100"
-GAP_pkg_juliainterface_jll = "=0.800.201"
+GAP_pkg_juliainterface_jll = "=0.800.201, =0.800.202"
```
fingolfin, line 23: The reason I've been so strict here is that we want to make sure we really use the exact same C code in the JLL which matches the one used in GAP_jll. However, for quick fixes like this, that's a nasty impediment. So, we should loosen it and just be very careful about how we change versions for `GAP_pkg_juliainterface_jll`.

However, I'd like to do this carefully, so just to get a fix out ASAP, I'll merge this PR as-is and then follow up with a looser version (and maybe more) later. I'll open an issue to remind me about that.

PR #874, opened 2023-05-10
avoid warnings shown by `TerminalMenus`

### LINKED ISSUE #872: Get rid of annoying `Warning: TerminalMenus: Unable to enter raw mode` when running test suite
I think these are just cosmetic for now, but still annoying as it makes it look as if something went wrong. It would be good if we could eventually get rid of those

ThomasBreuer:
resolves #872

According to `?REPL.TerminalMenus.config`, the argument `supress_output` (sic!) is not supported anymore. The proposed change follows the advice to call `request` with a `suppress_output` keyword argument.

### DIFF
--- a/src/help.jl
+++ b/src/help.jl
@@ -3,7 +3,8 @@
 using REPL
 
 function gap_help_string(topic::String, onlyexact::Bool = false,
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Thanks!

PR #875, opened 2023-05-10
for debugging

ThomasBreuer:
added two modified `AutoDoc` functions to `makedoc.g`, which contain `Print` statements


fingolfin:
It hangs at this point:
```
AutoDoc: after AutoDocScanFiles
AutoDoc: before WriteDocumentation
Error: The operation was canceled.
```
I'll restart it again, to see if we can get another hang, and if it is at the same spot.

fingolfin:
I've disabled some of the unnecessary CI jobs

ThomasBreuer:
Current situation:
`CI / Julia 1.8 - macOS-latest (pull_request) Cancelled after 20m`
The output looks as follows.
```
WriteChunks called
AUTODOC_AbsolutePath called
AUTODOC_AbsolutePath: before Process, pwd is '/bin/pwd'
Error: The operation was canceled.
```
Now the process seems to hang inside `Process(Directory(dir), pwd, InputTextNone(), OutputTextString(result, true), []);`.

The test `CI / Julia nightly - macOS-latest (pull_request) ` succeeds, the relevant output lines are as follows.
```
WriteChunks called
AUTODOC_AbsolutePath called
AUTODOC_AbsolutePath: before Process, pwd is '/bin/pwd'
AUTODOC_AbsolutePath: after Process
AUTODOC_AbsolutePath: result is '/Users/runner/work/GAP.jl/GAP.jl/pkg/JuliaInt\
erface/doc
'
AUTODOC_AbsolutePath: result is '/Users/runner/work/GAP.jl/GAP.jl/pkg/JuliaInt\
erface/doc/_Chunks.xml'
leave AUTODOC_AbsolutePath
WriteChunks: after AUTODOC_AbsolutePath
[...]
```

fingolfin:
Aha, so apparently it hangs running `pwd` via `Process`. I guess that makes sense given the kind of issue we've been seeing. I'll investigate!

Of course we could just avoid that particular subprocess call in AutoDoc by using the `IO_getcwd` command available in current versions of the IO package, but that would just hide the deeper issue, not solve it

fingolfin:
Closing this for now as #906 should have resolved the issue for now. See also issue #908

PR #883, opened 2023-05-23
Add `Aqua.jl` for some static testing

lgoettgens:
See https://github.com/oscar-system/Oscar.jl/issues/2385.

### DIFF
--- a/test/Aqua.jl
+++ b/test/Aqua.jl
@@ -1,0 +1,15 @@
+using Aqua
+
+@testset "Aqua.jl" begin
+    Aqua.test_all(
+        GAP;
+        ambiguities=false,      # TODO: fix ambiguities
+        unbound_args=true,
+        undefined_exports=true,
+        project_extras=true,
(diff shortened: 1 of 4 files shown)


test/Aqua.jl
```diff
@@ -0,0 +1,15 @@
...
+    Aqua.test_all(
+        GAP;
+        ambiguities=false,      # TODO: fix ambiguities
+        unbound_args=true,
+        undefined_exports=true,
+        project_extras=true,
+        stale_deps=false,       # some weird error with GAP_lib_jll
+        deps_compat=true,
+        project_toml_formatting=true,
+        piracy=false            # TODO: fix
```
fingolfin, line 13: Note that I believe the reported issues are not actuality type piracy but rather a bug in Aqua; I already reported this there, too

lgoettgens: I haven't looked into everything that has been reported by Aqua and just disabled the failing checks for the time being. This should be tackled in future PRs.

Issue #884, opened 2023-05-29
Change `Artifacts` location


lgoettgens:
According to https://docs.julialang.org/en/v1/stdlib/Artifacts/#Artifacts, the `Artifacts` library moved from `Pkg.Artifacts` to being a base library. This should be changed in `GAP.jl`.

PR #885, opened 2023-05-29
Update `Artifacts` location

lgoettgens:
and use only specific parts of `Pkg`.

Resolves https://github.com/oscar-system/GAP.jl/issues/884.


fingolfin (approved):

fingolfin:
Actually I'll have to revert this: `Artifacts` is a subset of `Pkg.Artifacts` and does not provide all features. In particular it is missing `download_artifact`.

fingolfin:
Ahh, but `download_artifact` is only called from `etc/`, so it is fine after all. Phew

PR #889, opened 2023-06-05
Update `Aqua.jl` to support piracy tests

lgoettgens:
Cf. https://github.com/oscar-system/GAP.jl/pull/883#discussion_r1203134701

cc @fingolfin

### DIFF
--- a/test/Aqua.jl
+++ b/test/Aqua.jl
@@ -10,6 +10,6 @@ using Aqua
         stale_deps=true,
         deps_compat=true,
         project_toml_formatting=true,
-        piracy=true
+        piracy=(GAP.use_jl_reinit_foreign_type() ? true : (treat_as_own=[GapObj],))
     )
 end
--- a/test/Aqua.jl
+++ b/test/Aqua.jl
@@ -10,6 +10,6 @@ using Aqua
         stale_deps=true,
         deps_compat=true,
         project_toml_formatting=true,
-        piracy=false            # TODO: fix
+        piracy=true
     )
 end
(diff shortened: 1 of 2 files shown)


lgoettgens:
After researching a bit more on this, I found https://github.com/JuliaLang/julia/issues/44661#issuecomment-1072913912 and followed the hacks mentioned there.

`GapObj` has different `parentmodule`s depending on the julia version. In particular, if `GAP.use_jl_reinit_foreign_type()` is `true` `parentmodule(GapObj) === GAP`, and `parentmodule(GapObj) === GAP_jll` otherwise.

I would thus suggest to just exclude `GapObj` from piracy tests once https://github.com/JuliaTesting/Aqua.jl/pull/140 is available. If sometime in the future the distinction using `GAP.use_jl_reinit_foreign_type()` is dropped, we can remove it from the test exclusion again.

fingolfin:
`GAP.use_jl_reinit_foreign_type` is only available (and necessary) in Julia >= 1.9. Until we drop support for older Julia versions, it'll stay around.

Of course now Aqua.jl 0.6.4 is out, so we could just use its new `treat_as_own` feature?

lgoettgens:
> Of course now Aqua.jl 0.6.4 is out, so we could just use its new `treat_as_own` feature?

Unfortunately, it is broken due to a dump typo by me... Please see https://github.com/JuliaTesting/Aqua.jl/pull/153 for a fix. As soon as there is a release with this fix, this PR should succeed.

fingolfin (approved):

PR #891, opened 2023-06-07
Turn hash for GapObj into error

fingolfin:
... instead of providing a trivial (constant) hash. This may break
some trivial applications, but in return it helps us identify code
that relies on the useless default hash, i.e., it helps us find
performance bottlenecks.

This is a breaking change so also increase the package version.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # Changes in GAP.jl
 
+## Version 0.10.0-DEV (released YYYY-MM-DD)
+
+- Changed `hash(::GapObj)` to throw an error (no general non-trivial hashing is
+  possible for general GAP objects)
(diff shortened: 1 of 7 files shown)


ThomasBreuer (approved):
Do we have to document the problem in general, and give examples of forbidden objects such as `Set`s of `GapObj`s?

fingolfin:
It certainly would be good to discuss the hashing issue somewhere, if only so we can point people at it when they ask about it...

ThomasBreuer:
@fingolfin Shall we merge this pull request, and then I create a new pull request for improving the documentation?

fingolfin:
This is a breaking change, meaning that once this is merged, the next version should be 0.10.0. But there are already a bunch of other changes merged -- perhaps we can make one more 0.9.x release before merging this -- I've made PR #926 in preparation of this.

Issue #897, opened 2023-06-15
Change documentation to warn against using `GAP.evalstr`


fingolfin:
I think we should warn users that they ideally should avoid `GAP.evalstr`. I.e. say so in its docstring, and point out alternatives.

Related, ideally we could also reduce the places we use it in the documentation -- the prominent use there might unduly encourage people to use it.

lgoettgens:
Resolved by #898 and #899

Issue #894, opened 2023-06-15
Document `GAP.evalstr_ex`?


fingolfin:
Perhaps we should document `GAP.evalstr_ex` which is a more powerful variant of `GAP.evalstr`?

fingolfin:
Resolved by #898

PR #898, opened 2023-06-15
improve documentation of `evalstr_ex`, `evalstr`

ThomasBreuer:
- document `evalstr_ex` (addresses #894)
- extend the documentation of `evalstr` (adresses #897)

Changing `evalstr` such that GAP's screen output is shown in the Julia session seems to be not possible if the libgap function `GAP_EvalString` is used. The return value contains strings that describe the screen output *plus* the return value  of each command, and separating the screen output from the string that represents the return value is in general not possible. Note that the pointers to the return values are not useful here because the objects may have changed since they got created, as in the following example.
```
julia> GAP.evalstr_ex( "x:= []; Add(x, 1);" )
GAP: [ [ true, [ 1 ], false,, "[  ]" ], [ true,, false,, "" ] ]
```
In the case of functions that have a return value *and* cause some printing to the screen, we cannot decide which part of the 5th entry in each entry of the result of `evalstr_ex` belongs to the return value.
(Perhaps libgap's `GAP_EvalString` should be changed accordingly. The fourth position in the entries of the result is currently unused.)

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -43,8 +43,8 @@ end
 @doc raw"""
     evalstr_ex(cmd::String)
 
-Assume that `cmd` consists of $n$ GAP commands, each terminated by `;` or `;;`.
-Let GAP execute these commands and return the GAP list of length $n$ that
+Assume that `cmd` consists of $n$ GAP statements, each terminated by `;` or `;;`.
+Let GAP execute these statements and return a GAP list of length $n$ that
 describes their results.
 Each entry of the return value is a GAP list of length 5,
 with the following meaning.


ThomasBreuer:
Some tests got canceled because they were running out of time:
- Julia nightly - ubuntu-latest in "GAP tests",
- Julia 1.8 - macOS-latest in "GAP tests",
- Julia nightly - macOS-latest in "Run tests".

src/ccalls.jl
```diff
@@ -40,6 +40,52 @@ function _JULIA_TO_GAP(x::Int)
 end
 
 
+@doc raw"""
+    evalstr_ex(cmd::String)
+
+Assume that `cmd` consists of $n$ GAP commands, each terminated by `;` or `;;`.
+Let GAP execute these commands and return the GAP list of length $n$ that
```
fingolfin, line 47:
```suggestion
Assume that `cmd` consists of $n$ GAP statements, each terminated by `;` or `;;`.
Let GAP execute these statements and return a GAP list of length $n$ that
```

fingolfin (approved):

PR #899, opened 2023-06-15
reduce the occurrences of `evalstr` in docstrings

ThomasBreuer:
addresses #897


ThomasBreuer:
Some tests got canceled because they were running out of time (same pattern as for #898).

PR #900, opened 2023-06-16
Adjust docstring examples

fingolfin:
- avoid more evalstr uses
- add one more `Vector` conversion example, to type `Vector{Vector{Int64}}`
- prefer `GAP.Obj` over `GAPObj` as it works in more cases, i.e.,
  users are in general better of using `GAP.Obj`
- separate kwargs using `;` instead of `,` (this is optional, but
  IMHO enhances clarity)
- another typical Julia code convention is to format kwargs tightly, as
  in `recursive=true` not `recursive = true`

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -16,20 +16,20 @@ their conversion with `BigInt` is handled by Julia's methods.)
 
 # Examples
 ```jldoctest
-julia> val = GAP.evalstr("2^64")
-GAP: 18446744073709551616
+julia> val = GAP.Globals.Factorial(25)
+GAP: 15511210043330985984000000
 
 julia> BigInt(val)
-18446744073709551616
+15511210043330985984000000
 
-julia> val = GAP.evalstr("2^59")
-576460752303423488
+julia> val = GAP.Globals.Factorial(10)
+3628800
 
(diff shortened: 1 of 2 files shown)


ThomasBreuer (approved):
Concerning the replaced `evalstr` examples: Yes, calling some `GAP.Globals` function is a good alternative.
Concerning `GAP.Obj` vs. `GapObj`, we should decide (and document) what we prefer where; up to now, I had prefered `GapObj` where possible because it is more specific.
(`Obj` is not an exported name, but of course this is not a reason for not using `GAP.Obj` in doctests.)

src/constructors.jl
```diff
@@ -273,29 +275,34 @@ If `T` is `UInt8` then `obj` may be a
...
  Any[2]
 
-julia> Vector{Any}(val, recursive = false)
+julia> Vector{Any}(val; recursive=false)
 2-element Vector{Any}:
  GAP: [ 1 ]
  GAP: [ 2 ]
 
+julia> Vector{Vector{Int64}}(val)
+3-element Vector{Int64}:
```
ThomasBreuer, line 292:
```suggestion
2-element Vector{Int64}:
```

PR #909, opened 2023-06-29
Reduce implicit dependencies on GAP packages

fingolfin:
These changes avoid some direct or indirect access to functionality
provided by GAP packages (namely CTblLib, GAPDoc, SmallGrp).

The ultimate goal is to switch GAP.jl to start GAP with `-A` or even
`--bare` (or something equivalent) and then leave it to users to load
packages. We could also have a `GAP_full.jl` or so which loads `GAP.jl`
and all default packages etc., but that's all very hypothetical: These
changes are harmless either way, even if everything else stays as it is.

### DIFF
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -22,16 +22,18 @@
     @test occursin("Set the value", string(doc))
 
     # Do tester and setter refer to the right objects?
+    # (Choose a group `G` whose center is different from the center
+    # of its derived subgroup.)
     @gapattribute cendersub(G::GapObj) = GAP.Globals.Centre(dersub(G))
-    G = GAP.Globals.SmallGroup(72, 15)
+    G = GAP.Globals.SymmetricGroup(3)
     @test GAP.Globals.Size(GAP.Globals.Centre(G)) == 1
-    G = GAP.Globals.SmallGroup(72, 15)  # create the group anew
+    G = GAP.Globals.SymmetricGroup(3)  # create the group anew
     @test ! GAP.Globals.HasCentre(G)
     @test ! has_dersub(G)
     @test ! has_cendersub(G)
     @test has_dersub(G)  # the previous call has set the value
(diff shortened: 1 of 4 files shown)


mohamed-barakat:
Wonderful!

test/macros.jl
```diff
@@ -23,15 +23,15 @@
 
     # Do tester and setter refer to the right objects?
```
ThomasBreuer, line 24: This is the idea of the test, thus `G` and `DerivedSubgroup(G)` should have different centres.
I suggest the following.
```
    # Do tester and setter refer to the right objects?
    # (Choose a group `G` whose center is different from the center
    # of its derived subgroup.)
    @gapattribute cendersub(G::GapObj) = GAP.Globals.Centre(dersub(G))
    G = GAP.Globals.SymmetricGroup(3)
    @test GAP.Globals.Size(GAP.Globals.Centre(G)) == 1
    G = GAP.Globals.SymmetricGroup(3)  # create the group anew
    @test ! GAP.Globals.HasCentre(G)
    @test ! has_dersub(G)
    @test ! has_cendersub(G)
    @test has_dersub(G)  # the previous call has set the value
    ggens = GAP.Globals.GeneratorsOfGroup(G)
    set_cendersub(G, GAP.Globals.Subgroup(G, GAP.GapObj([ggens[1]])))
    @test has_cendersub(G)
    @test GAP.Globals.HasCentre(GAP.Globals.DerivedSubgroup(G))
    @test GAP.Globals.Size(GAP.Globals.Centre(G)) == 1
```

ThomasBreuer (approved):
Looks good, except for one changed test.

test/macros.jl
fingolfin: Thank you for catching this. I must admit I did not think about what the test is supposed to do. I've applied your suggestion!

PR #910, opened 2023-06-30
Allow GAP.Obj(x,true) for recursive conversion

fingolfin:
This makes it much more convenient to use from GAP: `Julia.GAP.Obj(x,true)`

TODO: document it somewhere, add a test

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -6,6 +6,9 @@ GapObj(x::GapObj) = x
 Obj(obj; recursive::Bool = false) = julia_to_gap(obj, IdDict(); recursive)::Obj
 GapObj(obj; recursive::Bool = false) = julia_to_gap(obj, IdDict(); recursive)::GapObj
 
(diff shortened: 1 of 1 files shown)


ThomasBreuer:
Concerning documentation, the obvious places are the docstrings for `GAP.Obj` and `GapObj`, which mention currently the keyword argument `recursive`.

However, this affects only the use from Julia. Since the remark mentions explicitly that the new syntax is intended for the use from GAP, the point is to mention it in the JuliaInterface manual. Currently we are advertising the GAP function (constructor) `JuliaToGAP`. If I understand the idea of this pull request right then it would be better to advertise `Julia.GAP.Obj`, for example because it covers also extensions from Oscar.jl, which `JuliaToGAP` does not know.

ThomasBreuer:
@fingolfin Shall we merge this pull request, and then I create a new pull request for improving the documentation?

fingolfin:
OK!

PR #912, opened 2023-07-06
Only load bare deps if env var GAP_BARE_DEPS is set

fingolfin:
This may or may not become default in a future (breaking) release. But for now it is simply useful for debugging the `GAP_pkg_*` work at https://github.com/oscar-system/GAP_pkg

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -286,6 +286,10 @@ function __init__()
         push!(cmdline_options, "-b")
     end
 
+    if haskey(ENV, "GAP_BARE_DEPS")
+        push!(cmdline_options, "-A")
+    end
+
     initialize(cmdline_options)
 
     if !show_banner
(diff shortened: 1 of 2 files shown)


src/GAP.jl
```diff
@@ -286,6 +286,9 @@ function __init__()
         push!(cmdline_options, "-b")
     end
 
+    if haskey(ENV, "GAP_BARE_DEPS")
+        push!(cmdline_options, "-A")
```
ThomasBreuer, line 290: I had expected that `push!(cmdline_options, "--bare")` would be enough, that is, the change in `gap/systemfile.g` would not be necessary. However, apparently then loading JuliaInterface causes problems. Thus the proposed changes seem to be a good solution for the moment.

ThomasBreuer (approved):

src/GAP.jl
fingolfin: Yeah I first tried `--bare`, too, but then it didn't load JuliaInterface either. Hence this solution now. (Besides, we still might want to always load e.g. GAPDoc... well, we definitely will want to load it and also smallgrp, transgrp, primgrap and a bunch more in Oscar.)

PR #914, opened 2023-07-10
Re-enable REPL completion for GAP records in Julia 1.10

fingolfin:
The added annotation might also benefit performance of some code.

Thanks to @aviatesk who helpfully suggested this in [this comment](https://github.com/oscar-system/GAP.jl/pull/866#issuecomment-1628181299).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.9.8-DEV"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 GAP_jll = "5cd7a574-2c56-5be2-91dc-c8bc375b9ddf"
 GAP_lib_jll = "de1ad85e-c930-5cd4-919d-ccd3fcafd1a3"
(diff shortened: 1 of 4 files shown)


fingolfin:
Weird. The CI on 1.10 keeps failing. But locally on my laptop with Julia 1.10.0-alpha1, the tests pass just fine.

fingolfin:
I figured out how to reproduce the errors locally: by enabling coverage, i.e., `using Pkg; Pkg.test(; coverage = true)`.

Indeed, I can also reproduce the problem interactively if I start Julia with the option `--code-coverage`, then do `using GAP` and try to tab complete e.g. `GAP.Globals.MTX.IsInd`.

So why does code coverage break this? Huh.

ThomasBreuer:
@fingolfin I can confirm your observation: When Julia 1.10 is started with `--code-coverage`, completion does not work in the session, but completion works when Julia 1.10 is started without `--code-coverage`.

Very strange.

ThomasBreuer:
I have run the two Julia sessions (with and without `--code-coverage`) in parallel, and followed the code that is used for the input `s = "GAP.Globals.MTX.IsI"; @which completions(s,lastindex(s))`.

I get a difference at `res = repl_eval_ex(ex, context_module)` where `ex = :(GAP.Globals.MTX)` and `context_module = Main`.
Specifically, calling `repl_eval_ex(:(GAP.Globals.MTX), Main)` yields `Any` in one Julia session and a complicated object of type `Core.Const` in the other session.

(Inside `REPLCompletions.repl_eval_ex`, the first difference occurs in the result of `REPLInterpreter(result)`, which looks rather cryptic.)

fingolfin:
I wonder if code coverage collection interferes in some way with `@assume_effects :effect_free :terminates_globally `

fingolfin:
OK this is a known thing, see https://github.com/JuliaLang/julia/issues/49978

fingolfin:
Based on that bug and the discussion on the Julia slack, I think the best solution for now is to just skip this test when code coverage tracking is on. It'll still work in REPL mode for 99.9% of users (I think people rarely turn on code coverage tracking for interactive use)

PR #925, opened 2023-07-24
document `Julia.GAP.Obj(x,true)`, add tests

ThomasBreuer:
This is the promised follow-up of #910.

### DIFF
--- a/pkg/JuliaInterface/gap/convert.gd
+++ b/pkg/JuliaInterface/gap/convert.gd
@@ -205,24 +205,24 @@
 #!    <C>UInt64</C>,
 #!    <C>UInt128</C>,
 #!    <C>BigInt</C>,
-#!    <C>Rational{T} where T &lt;:&nbsp;Integer</C>,
+#!    or <C>Rational{T} where T &lt;:&nbsp;Integer</C>,
 #!  </Item>
 #!  <Item>
 #!    <C>GapObj and IsInt</C> to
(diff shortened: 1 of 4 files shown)


ThomasBreuer:
The test `OscarCI / matching: [Oscar#master] - ubuntu-latest, julia ~1.6.0-0 (pull_request)` fails:
```
du Val Tester: Error During Test at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/test/AlgebraicGeometry/Schemes/duValSing.jl:12
  Test threw exception
  Expression: is_du_val_singularity(X, J3) == false
  Ideal does not describe a single K-point
```
This is almost the same error as in the failures of oscar-system/Oscar.jl/pull/2410 and oscar-system/Oscar.jl/pull/2528.

fingolfin:
This is https://github.com/oscar-system/Oscar.jl/issues/2531 -- it is a bug in Singular that needs to be fixed there, which should be in the works

ThomasBreuer:
Thanks @fingolfin. Then I regard this pull request as ready for review.
(Is it a reasonable idea to document `Julia.GAP.Obj` in the JuliaInterface manual as if it were a GAP function?)

fingolfin:
> (Is it a reasonable idea to document `Julia.GAP.Obj` in the JuliaInterface manual as if it were a GAP function?)

IMO it is1

fingolfin (approved):

Issue #928, opened 2023-08-21
Trouble installing GAP package `nq` (and hence `hap`) on macOS


fingolfin:
Getting this with latest GAP.jl master:
```
julia> GAP.Packages.install("nq")
#I  Getting PackageInfo URLs...
#I  Retrieving PackageInfo.g from https://gap-packages.github.io/nq/PackageInfo.g ...
#I  Downloading archive from URL https://github.com/gap-packages/nq/releases/download/v2.5.10/nq-2.5.10.tar.gz ...
#I  Saved archive to /var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T//gaptempdirrxth3u/nq-2.5.10.tar.gz.pkgman
#I  Extracting to /Users/mhorn/.julia/gaproot/v4.12/pkg/nq-2.5.10 ...
#I  Checking dependencies for nq...
#I    polycyclic 2.11: true
#I  Package ``nq'': The executable program is not available
#I  Running compilation script on /Users/mhorn/.julia/gaproot/v4.12/pkg/nq-2.5.10 ...
#I  Possible error detected: see log at /var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T//gaptempdirKXQUeD/exec-log.txt
#I  Compilation failed for package 'nq' (package may still be usable)
#I  Package ``nq'': The executable program is not available
#I  Checking dependencies for nq...
#I    polycyclic 2.11: true
#I  Package ``nq'': The executable program is not available
#I  Package availability test failed
#I  (for nq 2.5.10)
#I  Removed directory /Users/mhorn/.julia/gaproot/v4.12/pkg/nq-2.5.10
false
```

ThomasBreuer:
Does the mentioned log file give any idea what went wrong?

(I cannot check the macOS version. When I try the same under Linux, I get a few warnings from the compilation step, but the return value is `true`.)

fingolfin:
Sorry, I should have mentioned that it works under Linux.

I cannot check the logs because PackageManager deletes them; and it also deletes the failed directory, so I can't debug the output of *that* either.

I've logged this as a debugging issue two years ago at https://github.com/gap-packages/PackageManager/issues/66 ...

I think it would be great if we had a kwarg `debug=true/false` for `GAP.Packages.install`; if set to `true` we'd instruct `PackageManager` to not delete the log and package directory. Of course that then also requires a corresponding feature in `PackageManager`

[...]

PR #931, opened 2023-08-21
Add options to debug package manager

fingolfin:
Motivated by #928.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -56,6 +56,14 @@ function init_packagemanager()
       # put the new method in the first position
       meths = Globals.Download_Methods
       Wrappers.Add(meths, GapObj(r, recursive=true), 1)
+
+      # monkey patch PackageManager so that we can disable removal of
+      # package directories for debugging purposes
+      orig_PKGMAN_RemoveDir = Globals.PKGMAN_RemoveDir
+      replace_global!(:PKGMAN_RemoveDir, function(dir)
+        Globals.ValueOption(GapObj("debug")) == true && return
+        orig_PKGMAN_RemoveDir(dir)
+      end)
     end
 end
 
@@ -186,6 +194,7 @@ end
 """
     install(spec::String, version::String = "";
                           interactive::Bool = true, quiet::Bool = false,
+                          debug::Bool = false,
                           pkgdir::AbstractString = GAP.Packages.DEFAULT_PKGDIR[])
 
 Download and install the GAP package given by `spec` into the `pkgdir`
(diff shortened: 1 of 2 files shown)


src/packages.jl
```diff
@@ -188,6 +194,7 @@ end
 """
     install(spec::String, version::String = "";
                           interactive::Bool = true, quiet::Bool = false,
+                          debug::Bool = false,
```
fingolfin, line 197: Of course this should also be documented; I'll do that once we settled on what final form this should have.

src/packages.jl
```diff
@@ -215,19 +222,20 @@ For details, please refer to its documentation.
...
                                pkgdir::AbstractString = DEFAULT_PKGDIR[])
     # point PackageManager to the given pkg dir
     Globals.PKGMAN_CustomPackageDir = GapObj(pkgdir)
     mkpath(pkgdir)
 
-    if quiet
+    if quiet || debug
       oldlevel = Wrappers.InfoLevel(Globals.InfoPackageManager)
-      Wrappers.SetInfoLevel(Globals.InfoPackageManager, 0)
+      Wrappers.SetInfoLevel(Globals.InfoPackageManager, quiet ? 0 : 3)
```
fingolfin, line 233: I wonder if it might not be better to separate `debug` and this "verbose" mode: leaving around partial packages could be a stumbling block for non-experts. Yet we may end up asking these very people to use `debug` if they report issues installing packages. Of course we can ask them to `GAP.Packages.remove("blah")` afterwards, but still...

So one idea would be to also add `verbose::Bool = false` as another kwarg. Of course then one can wonder what happens if both `quiet` and `verbose` are set (an error?). Seeing how `quiet` is off by default, perhaps a better alternative would be to add a "verbosity level". So:
- `verbose::Int = 0` is a new kwarg and directly corresponds to the `InfoLevel`
  - alternative names: `verbosity`, `verbosity_level`, `infolevel`, `info_level`, ... ?
  - default value is `0` to match the current default of `quiet=true`
- we either drop `quiet`; or keep it as "deprecated" ("please instead use ...")
  - if set to `true` that sets `verbose` to 0, if set to `false` it sets it to whatever is the default value for `InfoPackageManager` (or whatever value we prefer)
  - implementation:  `function foo(quiet::Bool = true, verbose::Int = (quiet ? 0 : 2))
- thus if both are given, `verbose` wins (we could of course also use `nothing` to detect the case and report an error, but I feel this is overthinking it... on the long run I'd just remove `quiet`)
- we remove the option of "not changing the infolevel (it is redundant if the user has full control over the info level anyway)

src/packages.jl
```diff
@@ -58,6 +56,14 @@ function init_packagemanager()
       # put the new method in the first position
       meths = Globals.Download_Methods
       Wrappers.Add(meths, GapObj(r, recursive=true), 1)
+
+      # monkey patch PackageManager so that we can disable removal of
+      # package directories for debugging purposes
+      orig_PKGMAN_RemoveDir = Globals.PKGMAN_RemoveDir
+      replace_global!(:PKGMAN_RemoveDir, function(dir)
+        Globals.ValueOption(GapObj("debug")) == true && return
+        orig_PKGMAN_RemoveDir(dir)
```
fingolfin, line 65: Of course we can remove this hack once we are using a PackageManager release with https://github.com/gap-packages/PackageManager/pull/119 or something equivalent in it

src/globals.jl
```diff
@@ -55,6 +55,13 @@ end
 
 propertynames(::GlobalsType, private::Bool=false) = Vector{Symbol}(Globals.NamesGVars())
 
+function replace_global!(name::Symbol, val::Any)
+    n = GapObj(name)
+    Wrappers.MakeReadWriteGlobal(n)
+    setproperty!(Globals, name, val)
+    Wrappers.MakeReadOnlyGlobal(n)
+end
```
ThomasBreuer, line 63: The analogous function `ReplaceGapFunc` in Oscar.jl/gap/OscarInterface/gap/OscarInterface.gi` creates a global variable `Concatenation("_ORIG_", name)` if it does not yet exist. Perhaps do this also here?

ThomasBreuer (approved):
I have made just two comments, one concrete about the code and one about the conceptual questions.
Yes, if setting the info level per function call is supported then an additional `quiet` option is not reasonable.
A function call that requests diagnostic output but without using `debug` mode would then have a keyword argument `info_level = 4`  or so.

The current default makes sense:
Show no messages if `GAP.Packages.load` is called, but use the current info level of PackageManager for the other functions of `GAP.Packages`.

fingolfin:
I fixed a few issues with this.

Overall I agree with your analysis @ThomasBreuer and since you've approved it, I assume we can merge it after CI runs (but since we'll meet in half an hour anyway, we can also talk about it then)

PR #933, opened 2023-09-03
Bump Aqua compat

lgoettgens:


fingolfin (approved):

fingolfin:
complains about possible type piracy due to (I think) GapObj (again)

lgoettgens:
I'll look into this next week. Could possibly be a bug in the new Aqua stuff as well, there are no tests with kwargs if I remember correctly.

lgoettgens:
> complains about possible type piracy due to (I think) GapObj (again)

Indeed a bug in Aqua (see https://github.com/JuliaTesting/Aqua.jl/pull/171). Your original fix for Unions like `GapObj` does not work for functions with a kwarg, like the one here. Once there is a Aqua release with https://github.com/JuliaTesting/Aqua.jl/pull/171, we can proceed here.

PR #936, opened 2023-09-11
Enable `Aqua.jl`'s ambiguity check

lgoettgens:
Parts of this have already been merged as https://github.com/oscar-system/GAP.jl/pull/937.

- 470bde701eb1ea095079c470ff62fb4af0604847 removes `(::Type{<:AbstractString})(::GapObj)`. This led to ambiguities with `Test.GenericString`, `LazyString`, `Core.Compiler.LazyString`. The only real removals here are for `SubString` and `SubstitutionString` as the other subtypes of `AbstractString` either have their specialized function or didn't work anyway due to ambiguities.
  It furthermore removes the unclear `gap_to_julia(AbstractString, ...)` (as it didn't specify what exactly to return). The implementation chose to return a simple `String`. I made that a bit more clear. Even the documentation says that one should put `String` as the first argument (cf. https://github.com/oscar-system/GAP.jl/blob/01b5c135392374124246e9a99d526f281a5e6e59/src/gap_to_julia.jl#L78). 
- 1291ab09b307d0aebc98d0fb837b191f12a7cdb6 fixes all calls of it and the corresponding tests.

Due to the AbstractString stuff above, I would technically consider this breaking. (E.g. `SubString(GAP.evalstr("\"foo\""))` used to work, but now needs to be `SubString(String(GAP.evalstr("\"foo\"")))`.)

Once this is merged, `GAP.jl` fulfills all of `Aqua.jl`'s tests! :tada:

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -141,7 +141,6 @@ gap_to_julia(::Type{Cuchar}, obj::GapObj) = Cuchar(obj)
 
 ## Strings
 gap_to_julia(::Type{String}, obj::GapObj) = String(obj)
-gap_to_julia(::Type{T}, obj::GapObj) where {T<:AbstractString} = T(obj)
 
 ## Symbols
 gap_to_julia(::Type{Symbol}, obj::GapObj) = Symbol(obj)
@@ -333,7 +332,7 @@ function gap_to_julia(x::GapObj; recursive::Bool = true)
     GAP_IS_MACFLOAT(x) && return gap_to_julia(Float64, x)
     GAP_IS_CHAR(x) && return gap_to_julia(Cuchar, x)
     # Do not choose this conversion for other lists in 'IsString'.
-    Wrappers.IsStringRep(x) && return gap_to_julia(AbstractString, x)
+    Wrappers.IsStringRep(x) && return gap_to_julia(String, x)
     # Do not choose this conversion for other lists in 'IsRange'.
     Wrappers.IsRangeRep(x) && return gap_to_julia(StepRange{Int64,Int64}, x)
     # Do not choose this conversion for other lists in 'IsBlist'.
--- a/pkg/JuliaInterface/tst/convert.tst
+++ b/pkg/JuliaInterface/tst/convert.tst
@@ -169,7 +169,7 @@ gap> ForAll([0..64], n -> Julia.GAP.Obj( -big2^n ) = -2^n);
 true
 
 #
-gap> string := GAPToJulia( Julia.Base.AbstractString, "bla" );
+gap> string := GAPToJulia( Julia.Base.String, "bla" );
 <Julia: "bla">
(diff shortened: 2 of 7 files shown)


lgoettgens:
There seems to be more behind this as initially expectes. I continue later.

src/gap_to_julia.jl
```diff
@@ -133,15 +133,15 @@ gap_to_julia(::Type{T}, x::GapInt) where {T<:Integer} = T(x)
 gap_to_julia(::Type{Rational{T}}, x::GapInt) where {T<:Integer} = Rational{T}(x)
 
 ## Floats
-gap_to_julia(::Type{T}, obj::GapObj) where {T<:AbstractFloat} = T(obj)
+gap_to_julia(::Type{T}, obj::GapObj) where {T<:AbstractFloat} = applicable(T, obj) ? T(obj) : T(Float64(obj))
```
fingolfin, line 136: I am really uncomfortable with using `applicable` here, IMHO this results in difficult to understand and handle code.

I think a more appropriate fix is to just do
```suggestion
gap_to_julia(::Type{Float64}, obj::GapObj) = Float64(obj)
```
and anyone who wants to convert to stuff to `Float32` or `BigFloat` has to deal with it by themselves (e.g. for a GAP big int it's a bad idea to first convert to Float64 in that case).

src/gap_to_julia.jl
```diff
@@ -133,15 +133,15 @@ gap_to_julia(::Type{T}, x::GapInt) where {T<:Integer} = T(x)
...
 
 ## Chars
 gap_to_julia(::Type{Char}, obj::GapObj) = Char(obj)
 gap_to_julia(::Type{Cuchar}, obj::GapObj) = Cuchar(obj)
 
 ## Strings
 gap_to_julia(::Type{String}, obj::GapObj) = String(obj)
-gap_to_julia(::Type{T}, obj::GapObj) where {T<:AbstractString} = T(obj)
+gap_to_julia(::Type{T}, obj::GapObj) where {T<:AbstractString} = applicable(T, obj) ? T(obj) : T(String(obj))
```
fingolfin, line 144:
```suggestion
gap_to_julia(::Type{String}, obj::GapObj) where {T<:AbstractString} = String(obj)
```

lgoettgens: I reworked most of this PR and included this change.

src/gap_to_julia.jl
lgoettgens: After looking again through the Float stuff, it seems that it caused no ambiguities, and it is fundamentally different to the removed String stuff, as it does not use a magic `convert`. So this doesn't need to be changed at all.

fingolfin: OK, great! (Sorry, didn't see this comment before)

fingolfin (approved):

PR #938, opened 2023-09-13
Simplify JuliaToGAP methods

fingolfin:
Avoid using CallJuliaFunctionWithKeywordArguments


ThomasBreuer (approved):
Nice.
(First I had thought that `CallJuliaFunctionWithKeywordArguments` could disappear, but no, just the better syntax for `julia_to_gap` allows us to avoid it here.)

PR #939, opened 2023-09-15
Teach InteractiveUtils.edit about GAP function

fingolfin:
Allow `InteractiveUtils.edit(GAP.Globals.Group)` and similar for any global function. Not (yet?) useful for GAP operations, but better than nothing.


test/convenience.jl
```diff
@@ -119,7 +119,12 @@ end
     @test sym5.:1 === sym5."1"
 end
 
-## Need to be created outside of test
-module test3
-SymmetricGroup = 5
+@testset "functionloc" begin
+
+    file, line = Base.functionloc(GAP.Globals.DirectoriesPackagePrograms)
+    @test samefile(file, "../gap/pkg.g")
```
ThomasBreuer, line 125: `Base.samefile` seems to be necessary.
And the result is `true` only if the tests are run from the `test` directory of the package; shall we assume this?

ThomasBreuer (approved):
Nice.
There is one problem with `samefile`, see my comment.

test/convenience.jl
fingolfin: It works in 1.9 (which is why the tests passed for me and I didn't notice the problem). But apparently it was undocumented in 1.6 and not exported, so yeah, I've change it

And it now uses `@__DIR__` to make the tests agnostic to where they are run from

ThomasBreuer (approved):

PR #941, opened 2023-09-18
Remove `(::<AbstractString)(::GapObj)`

fingolfin:
This is part of PR #936 by @lgoettgens and already mergable by itself.

It is also technically a breaking change, I guess, although I don't know of any code relying on this... but who knows....


lgoettgens:
I do not see any difference between this and #936 (except for toggling the Aqua test switch)

fingolfin:
The other PR was marked as draft and I thought there is more work to be done there, so I thought we could merge this part of it already now -- but I now see you actually "completed" PR #936 now, in which case I am of course also happy to just merge that one directly :-)

PR #942, opened 2023-09-18
Change 'nonnegative' to 'non-negative' for consistency

fingolfin:


ThomasBreuer (approved):
O.k.
Concerning consistency, I think "nonnegative" occured more often in GAP.jl than "non-negative". Oscar.jl has a clear majority for "non-negative", and gap-system has a small majority for "non-negative". So let us switch to "non-negative".

PR #944, opened 2023-09-26
added `hasbangindex`, `hasbangproperty`

ThomasBreuer:


pkg/JuliaInterface/gap/utils.gi
```diff
@@ -7,13 +7,19 @@
 ## Access data from component and positional objects
 BindGlobal( "BangComponent", { obj, nam } -> obj!.( nam ) );
 
+BindGlobal( "HasBangComponent",
+  { obj, nam } -> nam in REC_NAMES_COMOBJ( obj ) );
```
fingolfin, line 11: How about this instead?
```suggestion
  { obj, nam } -> IsBound( obj!.( nam ) );
```

pkg/JuliaInterface/gap/utils.gi
```diff
@@ -7,13 +7,19 @@
...
+
 BindGlobal( "SetBangComponent",
   function( obj, nam, val )
     obj!.( nam ):= val;
   end );
 
 BindGlobal( "BangPosition", { obj, pos } -> obj![ pos ] );
 
+BindGlobal( "HasBangPosition",
+  { obj, pos } -> pos <= LEN_POSOBJ( obj ) );
```
fingolfin, line 21: That could still return `true` for entries that are not bound, e.g. if entry 2 is bound but entry 1 is not... This sometimes happens. 

Why not just
```suggestion
  { obj, pos } -> IsBound( obj![ pos ] ) );
```

pkg/JuliaInterface/gap/utils.gi
```diff
@@ -7,13 +7,19 @@
 ## Access data from component and positional objects
 BindGlobal( "BangComponent", { obj, nam } -> obj!.( nam ) );
 
+BindGlobal( "HasBangComponent",
+  { obj, nam } -> IsBound( obj!.( nam ) );
```
fingolfin, line 11:
```suggestion
  { obj, nam } -> IsBound( obj!.( nam ) ) );
```

fingolfin (approved):

PR #945, opened 2023-09-26
adjust `docs/make.jl` to changes in Documenter.jl

ThomasBreuer:
Now we can *almost* process the documentation with Documenter.jl 1.0.

- The creation of hyperlinks to the GAP documentation works again.
- The errors because of docstrings that are not referenced in `docs/src/*.md` are gone because the docstrings are now referenced. (I do not like this at all. Why is it an error not to include all available docstrings?)
- There are still errors because three explicit cross-references to the JuliaInterface manual (in the directory `assets/html/JuliaInterface`) are not accepted because they are local links outside the build directory.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -18,43 +18,96 @@ using GAP.Markdown
 ## of the function `Documenter.Selectors.order` for these subtypes,
 ## and executing the step for the subtype `T` means to call
 ## `Documenter.Selectors.runner` with `T` and the document object.
-##
-## The code was adapted from the package
-## https://github.com/ali-ramadhan/DocumenterBibliographyTest.jl,
-## the hint to use this approach came up in the discussion of
-## https://github.com/JuliaDocs/Documenter.jl/issues/1343.
 
 ## Declare a new subtype for our step,
 ## the evaluation of external references.
 abstract type ExternalReference <: Documenter.Builder.DocumentPipeline end
 
-## Execute our step after the computation of cross-references (3.0)
-## and before the check of the document (4.0).
-Documenter.Selectors.order(::Type{ExternalReference}) = 3.1
+## Distinguish different versions of Documenter.jl.
+if isdefined(Documenter, :Document)
+  # version at least 1.0
+  # The code was adapted from `expand_citations`
+  # in DocumenterCitations.jl/src/citations.jl.
+  Document = Documenter.Document
 
-## When executing our step, print an info line
-## and call the function that does the work.
-function Documenter.Selectors.runner(::Type{ExternalReference}, doc::Documenter.Documents.Document)
-    @info "ExternalReference: building external references."
-    compute_external_references(doc)
-end
+  # Execute our step before the computation of cross-references (3.0)
+  # because otherwise the syntax of external references leads to errors.
+  Documenter.Selectors.order(::Type{ExternalReference}) = 2.99
 
-function compute_external_references(doc::Documenter.Documents.Document)
+  function compute_external_references(doc::Document)
+    for (src, page) in doc.blueprint.pages
+      @debug "ExternalReference: compute for entries in $(src)"
+      empty!(page.globals.meta)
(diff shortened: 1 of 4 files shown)


fingolfin:
You can also add `checkdocs = :none` to the arguments of `makedocs` to stop it from throwing an error when a docstring does not appear in the manual.

Alternatively one can set `warnonly = true` -- this is actually what we had before (`warnonly` replaces the former option `strict` which defaulted to `false` -- so in a sense they just toggled the default value...).  The drawback is that `warnonly = true` will also treat other things as only warnings, were we really might want to get an error, so we notice it and can fix it....

docs/src/other.md
```diff
@@ -171,3 +171,12 @@ true
 ```@docs
 show_gap_help
 ```
+
+## Objects that must be documented, otherwise Documenter.jl throws an error.
```
fingolfin, line 175: If we include these, then I think we should follow the example Documenter itself: it has a section "Developers" in its manual, and under it "Contributing" and "Internals".

Listing these under "Developers >> Internal" seems OK to me...

docs/make.jl
```diff
@@ -18,43 +18,96 @@ using GAP.Markdown
...
+    @info "Computing external reference: $(url)."
+    reference = match(r"GAP_ref\((.*)\)", url)
+    (! isnothing(reference) && length(reference.captures) == 1) || return true
+    reference = string(reference.captures[1])
+    key = split(reference, ":")
+    length(key) == 2 || return true
+    key = GapObj(string(key[2]))
+    # Find all matches of `"<bookname>:<label>"`.
+    urls = GAP.Globals.MatchURLs(GapObj(reference),
+               GapObj("https://www.gap-system.org/Manuals/"))
```
fingolfin, line 74: Actually, GAP manuals these days are under https://docs.gap-system.org/ e.g. https://docs.gap-system.org/doc/ref/chap0_mj.html

docs/make.jl
```diff
@@ -18,43 +18,96 @@ using GAP.Markdown
...
+  end
 
-function compute_external_references(doc::Documenter.Documents.Document)
+  function compute_external_reference(node::Documenter.MarkdownAST.Node, meta, page, doc)
+    # Do something only if the current element is a `MarkdownAST.Link`.
+    link = node.element
+    isa(link, Documenter.MarkdownAST.Link) || return true
+    url = link.destination
+    startswith(url, "GAP_ref") || return true
+    @info "Computing external reference: $(url)."
```
fingolfin, line 65: This function seems to be mostly identical to the old `compute_external_reference(link, meta, page, doc)` method below. Only instead of nested `if` blocks this one here aborts early via `|| return true` ; and of course `links.url` vs. `links.destination`.

Can't we extract the main body of this into a helper taking a single argument, `url`, and which returns either `nothing` or else the modified `url` value.

Then the two methods can both call this helper, and if it returns a value `!== nothing` they assign it to `link.destination` resp. `link.url`.

docs/src/other.md
ThomasBreuer: O.k.
(The hierarchy of the Documenter.jl manual is deeper than that of the GAP.jl manual, thus the result looks different.)

docs/make.jl
ThomasBreuer: O.k.

docs/make.jl
ThomasBreuer: O.k.
I think the point is that having such a helper function is anyhow a good idea, not just that it avoids a (temporary) code duplication.

fingolfin:
Oops, I kinda forgot about this PR.

Everything looks good, tests pass -- is there any reason not to merge it?

fingolfin (approved):

fingolfin:
Oh wait, now I understand: the three errors you are referring to are for when one switches to actually using Documenter.jl 1.x -- which this PR does not, it just prepares for this.

PR #950, opened 2023-10-27
Resolve a bunch of issues reported by JET

fingolfin:
Specifically I used

  using JET, GAP ; report_package(GAP)

There are some remaining warnings; however, those related to
`Downloads.download` and to `REPL.REPLCompletions.completions`
seem to be issues in those stdlibs, not in our code.

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -254,7 +254,7 @@ function BitVector(obj::GapObj)
     !Wrappers.IsBlist(obj) && throw(ConversionError(obj, BitVector))
     # TODO: a much better conversion would be possible, at least if `obj` is
     # in IsBlistRep, then we could essentially memcpy data
-    len = Wrappers.Length(obj)
+    len = Wrappers.Length(obj)::Int
     result = BitVector(undef, len)
     for i = 1:len
         result[i] = obj[i]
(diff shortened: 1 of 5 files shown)


src/adapter.jl
```diff
@@ -14,8 +14,9 @@ function show_string(io::IO, obj::Union{GapObj,FFE})
       # For objects that do not fit on the screen,
       # show only the first and the last lines.
       upper = div(rows, 2)
-      stri = join(lines[1:upper], "\n") * "\n  ⋮\n" *
-             join(lines[(end-rows+upper+2):end], "\n")
+      stri = join(lines[1:upper], "\n")::String
+      stri *= "\n  ⋮\n"
+      stri *= join(lines[(end-rows+upper+2):end], "\n")::String
```
ThomasBreuer, line 19: I do not understand why the old code was problematic:
Doesn't the documentation of `join` say that a `String` is returned?

src/constructors.jl
```diff
@@ -254,7 +254,7 @@ function BitVector(obj::GapObj)
     !Wrappers.IsBlist(obj) && throw(ConversionError(obj, BitVector))
     # TODO: a much better conversion would be possible, at least if `obj` is
     # in IsBlistRep, then we could essentially memcpy data
-    len = Wrappers.Length(obj)
+    len = Wrappers.Length(obj)::Int
```
ThomasBreuer, line 257: Ah, I have learned something:
`Wrappers.Length` defines its return type to be just `GapInt` (for example because there are infinite lists), but here we know more.
Thus it is worth to think about type annotations also for functions defined in `Wrappers`.

src/adapter.jl
fingolfin: I've dropped this for now until I understand it properly.

PR #951, opened 2023-11-01
Fix more JET issues

fingolfin:
Some found via @report_opt

### DIFF
--- a/src/ccalls.jl
+++ b/src/ccalls.jl
@@ -136,17 +136,17 @@ whereas `x` in the latter example lives in the Julia session.
 """
 function evalstr(cmd::String)
     res = evalstr_ex(cmd * ";")
-    if any(x->x[1] == false, res)
+    if any(x::GapObj->x[1] === false, res)
       # error
       global last_error
       # HACK HACK HACK: if there is an error string on the GAP side, call
       # error_handler to copy it into `last_error`
-      if !Wrappers.IsEmpty(Globals._JULIAINTERFACE_ERROR_BUFFER)
+      if !Wrappers.IsEmpty(Globals._JULIAINTERFACE_ERROR_BUFFER::GapObj)
         error_handler()
       end
       error("Error thrown by GAP: $(last_error[])")
     end
-    res = res[end]
+    res = res[end]::GapObj
     if Wrappers.ISB_LIST(res, 2)
       return res[2]
     else
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -219,7 +219,8 @@ function julia_to_gap(
     elseif Wrappers.IsRecord(obj)
         ret_val = NewPrecord(0)
         recursion_dict[obj] = ret_val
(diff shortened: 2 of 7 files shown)


src/julia_to_gap.jl
```diff
@@ -219,7 +219,7 @@ function julia_to_gap(
     elseif Wrappers.IsRecord(obj)
         ret_val = NewPrecord(0)
         recursion_dict[obj] = ret_val
-        for x in Vector{String}(Wrappers.RecNames(obj))
+        for x in Wrappers.RecNames(obj)::GapObj
```
fingolfin, line 222: This was a genuine bug here -- we are missing code coverage here.

src/julia_to_gap.jl
```diff
@@ -219,7 +219,7 @@ function julia_to_gap(
...
             Wrappers.ASS_REC(ret_val, x, julia_to_gap(Wrappers.ELM_REC(obj, x), recursion_dict; recursive = true))
```
fingolfin, line 223: (and this code is still wrong, I'll look into fixing it

src/julia_to_gap.jl
ThomasBreuer: Yes, this was wrong.
Looking at the function again, I am also not sure whether the statement
> We have to do something only if recursive conversion is required

is correct: Don't we have to make at least a shallow copy?

src/julia_to_gap.jl
ThomasBreuer: What about
```
xx = Wrappers.RNamObj(x)
Wrappers.ASS_REC(ret_val, xx, julia_to_gap(Wrappers.ELM_REC(obj, xx), recursion_dict; recursive = true))
```

src/ccalls.jl
```diff
@@ -136,17 +136,17 @@ whereas `x` in the latter example lives in the Julia session.
 """
 function evalstr(cmd::String)
     res = evalstr_ex(cmd * ";")
-    if any(x->x[1] == false, res)
+    if any(x::GapObj->x[1] === false, res)
```
ThomasBreuer, line 139: I read this as the recommendation to use always `===` or `!==` when comparing something with `true`, `false`, `GAP.Globals.fail`.

ThomasBreuer (approved):
Thanks.
This looks good, modulo the final fix in `julia_to_gap.jl`.

src/julia_to_gap.jl
fingolfin: Sounds good. I just updated the PR accordingly and added a test.

src/ccalls.jl
fingolfin: Yeah. It at least ensures that the compiler can always turn this into the most optimal code; here, it can't predict the type of `x[1]` so it should check at runtime whether there is a `==` method for whatever `x[1]` is plus a `Bool`... But we have extra information about the intent of the code, so that `===` is what we want, and can totally away runtime dispatch here.

PR #953, opened 2023-11-02
force version 1.0 of Documenter.jl

ThomasBreuer:
(as in AbstractAlgebra.jl and Nemo.jl)

For that, I have changed the three cross-references to the JuliaInterface manual from relative
(pointing to `../assets/html/JuliaInterface/chap0_mj.html`) to absolute
(pointing to `https://oscar-system.github.io/GAP.jl/stable/assets/html/JuliaInterface/chap0_mj.html`).

This is not really nice, but I did not find another solution. When Documenter.jl (in version 1.0 or higher) checks the document's structure, it does not accept relative links outside the build directory, and apparently there is no option to suppress this check.

(One could think of encoding the links in question as `https://..` URLs in the document, and inserting a new step into the build process, after the checks of course, which just replaces these URLs by the intended relative ones.  I think this would be absurd.
If it would be possible to insert inline HTML code then the solution would be easy, but apparently this is not intended.)

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,4 +2,4 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
-Documenter = "^0.27.0"
+Documenter = "1.0"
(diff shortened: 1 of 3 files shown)


ThomasBreuer:
I do not understand the test outputs:
Some tests pass, but when I look at the details then I find that "Build package" was the last step that was started, and this step got canceled or reported an error.
Some tests fail, and this happens also in one of the early steps, before the intended tests start.

fingolfin:
As to the errors, they may be temporary? Let's close and re-open to see

fingolfin (approved):
works now just fine

PR #958, opened 2024-01-11
improve package handling

### LINKED ISSUE #934: Add a helper to figure out where a loaded package is installed
Useful for e.g. debugging.
Basically something like
```
function locate_package(name::String)
  String(GAP.Globals.PackageInfo(GAP.Obj(name))[1].InstallationPath::GapObj)
end
```
though ideally also with a check that the package name is valid and the package loaded.

### LINKED ISSUE #954: Feature request: Support `GAP.Packages.load("BLA", install=URL)`
Recently we had need to load & install a GAP package that is not bundled with GAP.

We ended up doing
```julia
GAP.Packages.install(URL)
GAP.Packages.load("BLA")
```
But I wonder: we already have
```
GAP.Packages.load("BLA"; install=true)
```
which is quite convenient sometimes. But it can only work with packages where GAP already knows the URL. 

So perhaps we should also allow
```
GAP.Packages.load("BLA"; install=URL)
```
as a convenient shortcut for the first example?

ThomasBreuer:
- add `GAP.Packages.locate_package(name)`, resolves #934
- admit `load(pkgname, install = url)`, resolves #954

### DIFF
--- a/test/packages.jl
+++ b/test/packages.jl
@@ -6,6 +6,8 @@
     @test GAP.Packages.locate_package("no such package") == ""
 
     @test GAP.Packages.install("fga", interactive = false)
+    @test ! isempty(GAP.Packages.locate_package("fga"))
+    @test ! isempty(GAP.Packages.locate_package("FGA"))
     @test GAP.Packages.remove("fga", interactive = false)
 
 #    pkgdir = mktempdir()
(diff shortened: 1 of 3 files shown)


test/packages.jl
```diff
@@ -3,6 +3,7 @@
     @test GAP.Packages.load("PackageManager")
     @test ! GAP.Packages.load("no such package")
     @test ! GAP.Packages.load("no such package", install = true)
+    @test GAP.Packages.locate_package("no such package") == ""
 
     @test GAP.Packages.install("fga", interactive = false)
```
fingolfin, line 8:
```suggestion
    @test GAP.Packages.install("fga", interactive = false)
    @test !isempty(GAP.Packages.locate_package("fga"))
    @test !isempty(GAP.Packages.locate_package("FGA"))
```

PR #965, opened 2024-02-09
CI: update to codecov/codecov-action@v4

lgoettgens:
Same as https://github.com/oscar-system/Oscar.jl/pull/3334

PR #966, opened 2024-02-13
adjust code and tests to changes in Nemo.jl

ThomasBreuer:
addresses the failures that happen in #965

### DIFF
--- a/pkg/JuliaInterface/tst/wrapper.tst
+++ b/pkg/JuliaInterface/tst/wrapper.tst
@@ -14,6 +14,7 @@ gap> Julia.Base.typeof(N);
 <Julia: BigInt>
 gap> N(1);
 Error, MethodError: objects of type BigInt are not callable
+Maybe you forgot to use an operator such as *, ^, %, / etc. ?
 
 ## wrap a Julia function
 gap> f := Objectify(type, rec());;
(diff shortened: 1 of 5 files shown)


ThomasBreuer:
After my second commit, things are getting even more funny:
The changes from this commit affect just the test file `pkg/JuliaInterface/tst/wrapper.tst`.
Now (among others) the test `Julia 1.6 - ubuntu-latest` fails, which had passed before the second commit, and reports problems in the test file `pkg/JuliaExperimental/tst/realcyc.tst`.
What is going on here?

pkg/JuliaInterface/tst/wrapper.tst
```diff
@@ -14,6 +14,7 @@ gap> Julia.Base.typeof(N);
 <Julia: BigInt>
 gap> N(1);
 Error, MethodError: objects of type BigInt are not callable
+Maybe you forgot to use an operator such as *, ^, %, / etc. ?
```
fingolfin, line 17: I don't see what the purpose of this test is supposed to be. It checks that a Julia `BigInt` is not callable, but how is that useful to use?
```suggestion
```

fingolfin (approved):

PR #967, opened 2024-02-15
Implement iteration over GAP iterators

fingolfin:


ThomasBreuer (approved):
Logically, this extension is fine:
Up to now, a new GAP iterator gets created when Julia iterates over a domain. Now a new iterator (a shallow copy) gets created when Julia iterates over a GAP iterator.

Perhaps the differences between iterating in GAP and in Julia can be documented (where is a good place for that?), in order to avoid confusion:
- Iterating in GAP over an iterator changes the iterator, iterating in Julia doesn't.
- Iterating in GAP over a list with holes skips the holes, iterating in Julia over a GAP list with holes also iterates over the holes (well, the holes before the end of the list) by returning `nothing` for each hole.

fingolfin:
> Perhaps the differences between iterating in GAP and in Julia can be documented (where is a good place for that?), in order to avoid confusion:
> 
> * Iterating in GAP over an iterator changes the iterator, iterating in Julia doesn't.

I could of course change it so that Julia iteration also "eats up" the GAP iterator. It just seemed less useful and would make it behave less like a Julia iterator.

That said, how would this cause trouble for anyone coming from GAP? Maybe people would copy iterators too often thinking they need to, but that's note a major issue.

> * Iterating in GAP over a list with holes skips the holes, iterating in Julia over a GAP list with holes also iterates over the holes (well, the holes before the end of the list) by returning `nothing` for each hole.

Ah, good point. Perhaps we should skip holes? I don't think it was a very intentional decision to not skip them.

ThomasBreuer:
@fingolfin The message `#I  The Julia package 'Downloads' cannot be loaded.` appears in the test logs,
for [a failed test](https://github.com/oscar-system/GAP.jl/actions/runs/10419258626/job/28856937895?pr=967) as well as [for a successful test](https://github.com/oscar-system/GAP.jl/actions/runs/10419258620/job/28856937497?pr=967).
I do not understand why this message appears: First, the package should be loadable since it is listed as a dependency. Second, the only function that shows this message is the GAP function `JuliaImportPackage` from the JuliaInterface package. But who calls this function with the argument `"Downloads"`?

Issue #680, opened 2021-08-18
Add `@gapproperty`


fingolfin:
This would be identical to `@gapattribute` except it would add a type assertion to the return value of the getter function, so that Julia code calling the wrapper could be type stable.

In other words, this:
```julia
@gapproperty isperfect(G::GAPGroup) = GAP.Globals.IsPerfectGroup(G.X)
```
would be equivalent to this:
```julia
@gapattribute isperfect(G::GAPGroup) = GAP.Globals.IsPerfectGroup(G.X)::Bool
```

PR #974, opened 2024-03-13
Add `@gapproperty`

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/680.

### DIFF
--- a/test/macros.jl
+++ b/test/macros.jl
@@ -40,6 +40,28 @@
 
 end
 
+@testset "@gapproperty" begin
+
+    """
+        isoddint_GAP(x)
+    
+    Return `true` if the input integer is odd.
+    """
+    @gapproperty isoddint_GAP(x::Int) = GAP.Globals.IsOddInt(x)
+
+    doc = string(@doc isoddint_GAP)
+    @test !occursin("No documentation found", string(doc))
+    @test occursin("if the input integer is odd", string(doc))
+
+    doc = string(@doc has_isoddint_GAP)
+    @test !occursin("No documentation found", string(doc))
+    @test occursin("has already been computed", string(doc))
+
+    doc = string(@doc set_isoddint_GAP)
+    @test !occursin("No documentation found", string(doc))
+    @test occursin("Set the value", string(doc))
+end
+
 @testset "compat" begin
 
     x = @gap (1, 2, 3)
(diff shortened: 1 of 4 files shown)


test/macros.jl
```diff
@@ -40,6 +40,28 @@
...
 end
 
+@testset "@gapproperty" begin
+
+    """
+        isoddint_GAP(x)
+    
+    Return `true` if the input integer is odd.
+    """
+    @gapproperty isoddint_GAP(x::Int) = GAP.Globals.IsOddInt(x)
```
fingolfin, line 50: Perhaps there could be a test using `@inferred` to verify it really is type stable and returning a `Bool`?

fingolfin (commented):
First off, thank you for your contribution!

However, I recommend that next time we have a quick chat before you tackle one of these -- because in this case, re-reading issue #680, today my thoughts about it are "let's just close it, adding `::Bool` is easy, and the only problem with it is that one can forget to do it; but one can also forget to use `@gapproperty` instead of `@gapattribute`, so we really gain nothing but have to invest a lot of work."

But since you already invested the work now, I am reluctant to let it go to waste... OTOH if we add it, we need to maintain it, and it ever so slightly increases complexity... 

Perhaps @ThomasBreuer has a thought on this as well?

lgoettgens:
> First off, thank you for your contribution!
> 
> However, I recommend that next time we have a quick chat before you tackle one of these -- because in this case, re-reading issue #680, today my thoughts about it are "let's just close it, adding `::Bool` is easy, and the only problem with it is that one can forget to do it; but one can also forget to use `@gapproperty` instead of `@gapattribute`, so we really gain nothing but have to invest a lot of work."
> 
> But since you already invested the work now, I am reluctant to let it go to waste... OTOH if we add it, we need to maintain it, and it ever so slightly increases complexity...

Yeah, I totally understand and am for than fine with just closing this PR. If I remember correctly, I was a bit frustrated from some other task and just wanted something "easy to tackle" and thus just took something from the issue tracker to do. Next time, I'll talk to you beforehand

fingolfin:
Just to avoid misunderstanding: of course it is fine for you to work on something w/o consulting with me first, I just want to avoid *causing* frustration on your part when I then later come and say "oh no we don't want that" and discard your hard work -- I hate doing that :-/. Hence my request (I think you already got that, but internet communication is difficult).

And I totally get the bit about working on something "attainable" when frustrated on something else, I also do that ;-)

fingolfin:
We talked about it, and decided to not move forward. Thank again Lars for your efforts, and sorry!

PR #975, opened 2024-03-19
Switch to GAP 4.13.0

fingolfin:


Project.toml
```diff
@@ -25,9 +25,9 @@ Scratch = "6c6a2e73-6563-6170-7368-637461726353"
 Artifacts = "1.6"
 Compat = "4.4.0"
 Downloads = "1.4.3"
-GAP_jll = "~400.1200.101"
-GAP_lib_jll = "~400.1201.100"
-GAP_pkg_juliainterface_jll = "=0.800.300"
+GAP_jll = "~400.1300.000"
+GAP_lib_jll = "~400.1300.000"
+#GAP_pkg_juliainterface_jll = "=0.800.300"
```
fingolfin, line 30: TODO: reactivate this once GAP_pkg_juliainterface_jll is updated

src/setup.jl
```diff
@@ -112,24 +111,30 @@ function regenerate_gaproot()
...
     gap_lib = joinpath(gap_prefix, "lib")
     sysinfo["GAP_LDFLAGS"] = "-L$(gap_lib) -lgap"
 
-    # adjust linker flags for GAP kernel extensions on macOS
+    # set gac compiler flags (get appended to GAP_CFLAGS and GAP_CXXFLAGS)
+    sysinfo["GAC_CFLAGS"] = "-fPIC"
     if Sys.isapple()
-        sysinfo["GAC_LDFLAGS"] = "-bundle -L$(gap_lib) -lgap"
+        sysinfo["GAC_CFLAGS"] = "-fno-common"
     end
```
ThomasBreuer, line 133: perhaps
`sysinfo["GAC_CFLAGS"] = Sys.isapple() ? "-fno-common" : "--fPIC"`?

ThomasBreuer (approved):
Looks good, as far as I can judge.
Just one possible simplification (see remark).
(And I assume that adjusting `sysinfo["GAP_CFLAGS"]` and `sysinfo["GAP_CXXFLAGS"]` had first been changed but now this is not needed anymore.)

src/setup.jl
fingolfin: In fact we can omit it now as the file provided by `GAP_jll` already got this right :-). I've pushed a commit cleaning this and a few other flags up, please have a look.

One thing confuses me, though: 
```
julia> gap_prefix = GAP.GAP_jll.find_artifact_dir();

julia> sysinfo = GAP.Setup.read_sysinfo_gap(joinpath(gap_prefix, "lib", "gap"));

julia> haskey(sysinfo,"GAP_CPPFLAGS")
false

julia> haskey(GAP.sysinfo,"GAP_CPPFLAGS")
true
```

So the file we get from `GAP_jll` does not define `GAP_CPPFLAGS`. But I am pretty sure GAP 4.13.0 sets `GAP_CPPFLAGS` in there... Huh. Anyway, we (still) need to override it, I am just confused by this discrepency. Ah well.

ThomasBreuer: The line in question in the file from GAP 4.13.0 is as follows.
```
GAP_CPPFLAGS="-I/workspace/destdir/include/gap -I/workspace/destdir/include -DUSE_JULIA_GC=1"
```
It contains two `=` characters.
Thus the condition from `read_sysinfo_gap` is violated that only lines are considered which contain exactly one `=`.

fingolfin: great catch, thanks Thomas

PR #981, opened 2024-03-19
fix GAP help access for Julia nightly

### LINKED ISSUE #978: CI tests fail in Julia nightly: `terminal` not defined in `REPL.TerminalMenus`
This is the error:
```
help: Error During Test at /home/runner/work/GAP.jl/GAP.jl/test/help.jl:1
  Got exception outside of a @test
  UndefVarError: `terminal` not defined in `REPL.TerminalMenus`
  Suggestion: check for spelling errors or missing imports.
  Stacktrace:
    [1] getproperty(x::Module, f::Symbol)
      @ Base ./Base.jl:42
    [2] top-level scope
      @ ~/work/GAP.jl/GAP.jl/test/help.jl:2
    [3] macro expansion
      @ /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.12/Test/src/Test.jl:164 [inlined]
    [4] macro expansion
      @ ~/work/GAP.jl/GAP.jl/test/help.jl:3 [inlined]
    [5] include(fname::String)
      @ Main ./sysimg.jl:38
    [6] top-level scope
      @ ~/work/GAP.jl/GAP.jl/test/runtests.jl:14
```

@ThomasBreuer maybe you can have a look?

ThomasBreuer:
In Julia 1.12, one can use `default_terminal()`
instead of `REPL.TerminalMenus.terminal`.

resolves #978

### DIFF
--- a/src/help.jl
+++ b/src/help.jl
@@ -5,7 +5,6 @@ import REPL
 
 if isdefined(REPL.TerminalMenus, :default_terminal)
   import REPL.TerminalMenus: default_terminal
-# default_terminal = REPL.TerminalMenus.default_terminal
 else
   # before Julia 1.12
   function default_terminal()
--- a/src/help.jl
+++ b/src/help.jl
@@ -3,7 +3,7 @@
 import Markdown
 import REPL
 
-if isdefined( REPL.TerminalMenus, :default_terminal)
+if isdefined(REPL.TerminalMenus, :default_terminal)
   import REPL.TerminalMenus: default_terminal
 # default_terminal = REPL.TerminalMenus.default_terminal
 else
--- a/src/help.jl
+++ b/src/help.jl
@@ -3,8 +3,17 @@
 import Markdown
 import REPL
 
+if isdefined( REPL.TerminalMenus, :default_terminal)
+  default_terminal = REPL.TerminalMenus.default_terminal
+else
+  # before Julia 1.12
+  function default_terminal()
+    return REPL.TerminalMenus.terminal
+  end
+end
+
 function gap_help_string(topic::String, onlyexact::Bool = false,
-    term::REPL.Terminals.TTYTerminal = REPL.TerminalMenus.terminal;
+    term::REPL.Terminals.TTYTerminal = default_terminal();
     suppress_output::Bool = false)
     # Let GAP collect the information.
     info = Globals.HELP_Info(GapObj(topic), onlyexact)
--- a/src/help.jl
+++ b/src/help.jl
@@ -4,7 +4,8 @@ import Markdown
 import REPL
 
 if isdefined( REPL.TerminalMenus, :default_terminal)
-  default_terminal = REPL.TerminalMenus.default_terminal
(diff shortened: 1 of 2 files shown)


src/help.jl
```diff
@@ -3,8 +3,17 @@
...
+else
+  # before Julia 1.12
+  function default_terminal()
+    return REPL.TerminalMenus.terminal
+  end
+end
+
 function gap_help_string(topic::String, onlyexact::Bool = false,
-    term::REPL.Terminals.TTYTerminal = REPL.TerminalMenus.terminal;
+    term::REPL.Terminals.TTYTerminal = default_terminal();
```
fingolfin, line 16:
```suggestion
    term::REPL.Terminals.TTYTerminal = default_terminal()
```
As far as I can tell, `term` is only used in one spot, in the call to `REPL.TerminalMenus.request` below. So perhaps we should just replace `term` there by `default_terminal();`

src/help.jl
```diff
@@ -3,8 +3,17 @@
 import Markdown
 import REPL
 
+if isdefined( REPL.TerminalMenus, :default_terminal)
+  default_terminal = REPL.TerminalMenus.default_terminal
```
fingolfin, line 7: Can we also do this? (Might have to use `@static if ...` above)
```suggestion
  import REPL.TerminalMenus: default_terminal
```

fingolfin (approved):

src/help.jl
ThomasBreuer: Here `term` is an argument for the function `gap_help_string`.
The idea behind the complicated syntax is to make this function testable, by providing input from a terminal, see `test/help.jl`.
(I think the fact that we found the problem with Julia nightly shows that this test is useful.)

fingolfin: Ahhhhh of course you are 100% right; I completely misread the code and missed that term is an argument

src/help.jl
ThomasBreuer: Yes, this works.

src/help.jl
```diff
@@ -3,8 +3,18 @@
 import Markdown
 import REPL
 
+if isdefined( REPL.TerminalMenus, :default_terminal)
+  import REPL.TerminalMenus: default_terminal
+# default_terminal = REPL.TerminalMenus.default_terminal
```
fingolfin, line 8: Can this go?
```suggestion
```

src/help.jl
```diff
@@ -3,8 +3,18 @@
 import Markdown
 import REPL
 
+if isdefined( REPL.TerminalMenus, :default_terminal)
```
fingolfin, line 6:
```suggestion
if isdefined(REPL.TerminalMenus, :default_terminal)
```

src/help.jl
ThomasBreuer: Yes.

src/help.jl
ThomasBreuer: Yes.

fingolfin (approved):

PR #982, opened 2024-03-21
read_sysinfo_gap: deal with multiple = on one line

fingolfin:


ThomasBreuer (approved):
This solves the problem from the discussion of #975.
(The change assumes that any line that is not a comment line and that contains at least one `=` character means an assignment to a variable, where the variable name is on the left of the first `=` and the value is on the right. If this is correct then the code is o.k.)

PR #991, opened 2024-05-14
Add coercion of arguments to `@wrap`

lgoettgens:
Resolves the first part of https://github.com/oscar-system/Oscar.jl/issues/3720.

Due to the way this is now implemented, it is very easy to adapt it for similar type annotations like `GapInt`.

Currently also contains https://github.com/oscar-system/GAP.jl/pull/992 to make the testsuite run on nightly. This PR can be rebased once https://github.com/oscar-system/GAP.jl/pull/992 is merged.


fingolfin (approved):
This looks exactly like what I was hoping for, great :-)

fingolfin:
@ThomasBreuer any concerns, comments, questions about this

ThomasBreuer:
I have no objections.

If I understand the idea right then we can (should?) use the new feature in Oscar as follows.
For an Oscar group `G`, we can call `Oscar.GAPWrap.ConjugacyClasses(G)` instead of `Oscar.GAPWrap.ConjugacyClasses(GapObj(G))`.
We cannot call `Oscar.GAPWrap.Size(G)` instead of `Oscar.GAPWrap.Size(GapObj(G))` because `Size` does not get wrapped with the requirement that its argument is a `GapObj`.
Thus the Oscar code will look a bit funny if we omit the unnecessary `GapObj` calls for arguments of `Oscar.GapWrap.Something` functions.

fingolfin:
But we could change the wrapper for Size to require a GAPObj. In fact there is a ton of wrappers we could do that. I was simply lazy when setting up most of them.

ThomasBreuer:
Looking at the list in `Oscar/src/GAP/wrappers.jl`, my understanding is that most (all?) of the `::Any` arguments can be safely changed to `::GAP.Obj`, assuming that there are no methods for the GAP functions in question that take other Julia objects.

This change would not help us to get an automatic indirection from Oscar objects to underlying GAP objects.

Writing `::GapObj` instead would probably fit to the way the functions are intended for Oscar, although the GAP functions in question really admit any input --the `IsSomething` filters are of that kind. We could go this way.

fingolfin:
We may be looking at different subsets of the wrappers? Because my view is that the majority would be safe to use `GapObj`, as no Int, Bool or FFE makes sense as argument for e.g. Size, IsFreeGroup, Image (for the first argument at least, the second is different), and a ton more.


That said, as I already suggested in the feature request leading to this: we could also add code to deal with GAP.Obj and GAP.GapInt. Indeed, we just have to add a few more `elseif` to the code in this PR to catch those.

src/macros.jl
```diff
@@ -359,23 +368,42 @@ macro wrap(ex)
...
+    # - if the type annotation is `GapObj`, the rhs has to be wrapped in `GAP.GapObj(...)` and the lhs has no type annotation
+    tempargs = [
+        begin
+            if x isa Symbol
+                (x, x)
+            elseif x.head == :(::) && length(x.args) == 2
+                var = x.args[1]
+                typeannot = x.args[2]
+                if typeannot in [:GapObj, :(GAP.GapObj)]
+                    (var, :(GAP.GapObj($var)))
```
fingolfin, line 383:
```suggestion
                if typeannot in [:GapObj, :(GAP.GapObj)]
                    (var, :(GAP.GapObj($var)))
                elseif typeannot == :(GAP.Obj)
                    (var, :(GAP.Obj($var)))
                elseif typeannot in [:GapInt, :(GAP.GapInt)]
                    (var, :(GAP.GapInt($var)))
```

fingolfin: Like this, @ThomasBreuer @lgoettgens ?

lgoettgens: one a purely syntactical level this looks right. But I don't really know about the capabilities of the `GAP.Obj(...)` and `GAP.GapInt(...)` constructors to judge this

lgoettgens, line 383: Do all of the constructors have a (inferrable) contract to return objects of the specific type?
One could adapt this here to 
```suggestion
                    (var, :(GAP.GapObj($var)::GAP.GapObj))
```
to make sure

ThomasBreuer: Concerning the possible "restriction" of functions to `::GapObj` arguments, this is what I wanted to express:
On the one hand, GAP's `IsFreeGroup` is a filter that accepts any `GAP.Obj`, hence `::(GAP.Obj)` would be correct.
On the other hand, we want to call `IsFreeGroup` from Oscar only with a `::GapObj` argument, hence it is o.k. to define it like this in the `@wrap` call.

Concerning automatic indirection from Oscar objects to GAP objects via `GAP.Obj`:
I see that this works already for Oscar groups because Oscar defines `GAP.julia_to_gap(obj::GAPGroup) = obj.X` and not just `GapObj(obj::GAPGroup) = obj.X`.
(Thus I did this the wrong way for `GAPGroupCharacterTable` ...)

I think this new feature looks elegant, but understanding the Oscar code becomes more subtle:
We have to distinguish carefully between the use of `GapObj` for the *conversion* of objects (which creates a new object in each call and should be avoided if possible) and its use for *indirection* to objects which already exist (which is fine).

fingolfin: I am not sure what you mean with "inferrable contract", I don't think such a thing exists in Julia? In the end it's the same as with every such thing in Julia: Even `String(x)` may not have an inferable type if some package (loaded outside of your control) decides to add a crazy method to it.

lgoettgens: That's exactly what I was talking about. Someone (maybe even GAP.jl itself) could add a method to GapObj that does not return a GapObj. With a type annotation as suggested above, we could make sure that we indeed get a GapObj in the coercion

fingolfin: Fair enough, so yeah add that :-)

fingolfin: @lgoettgens can you also adapt this suggestion?

src/macros.jl
```diff
@@ -359,23 +368,42 @@ macro wrap(ex)
...
+    # - if the type annotation is `GapObj`, the rhs has to be wrapped in `GAP.GapObj(...)` and the lhs has no type annotation
+    tempargs = [
+        begin
+            if x isa Symbol
+                (x, x)
+            elseif x.head == :(::) && length(x.args) == 2
+                var = x.args[1]
+                typeannot = x.args[2]
+                if typeannot in [:GapObj, :(GAP.GapObj)]
+                    (var, :(GAP.GapObj($var)::GAP.GapObj))
```
lgoettgens, line 383:
```suggestion
                if typeannot in [:GapObj, :(GAP.GapObj)]
                    (var, :(GAP.GapObj($var)::GAP.GapObj))
                elseif typeannot == :(GAP.Obj)
                    (var, :(GAP.Obj($var)::GAP.Obj))
                elseif typeannot in [:GapInt, :(GAP.GapInt)]
                    (var, :(GAP.GapInt($var)::GAP.GapInt))
```
as suggested in https://github.com/oscar-system/GAP.jl/pull/991#discussion_r1603432546

src/macros.jl
lgoettgens: Done

src/macros.jl
```diff
@@ -359,23 +368,46 @@ macro wrap(ex)
 
     fullargs = ex.args[2:length(ex.args)]
 
-    # strip type annotation from arguments for use in the call to GAP
-    args = [x isa Symbol ? x : x.args[1] for x in fullargs]
-
+    # splits the arguments with type annotations into expressions for the lhs and rhs
+    # of the call of the form `func(args) = GAP.Globals.func(args)`:
+    # - type annotations have to be removed for the rhs
+    # - if the type annotation is `GapObj`, the rhs has to be wrapped in `GAP.GapObj(...)` and the lhs has no type annotation
```
fingolfin, line 374: Should we also extend the comment to match the code?
And/or add relevant comments next to the code.```suggestion
    # - if the type annotation is `GapObj`, `GapInt`, or a few other specific cases (see the code for details), the rhs is wrapped in `GAP.GapObj(...)` and we drop the type annotation for the lhs
```
(and perhaps line-wrap this)

File: src/macros.jl [hunk 4]
```diff
@@ -359,23 +368,46 @@ macro wrap(ex)
...
+    tempargs = [
+        begin
+            if x isa Symbol
```
fingolfin, line 377:
```suggestion
            if x isa Symbol
                # no type annotations, we can just use x on lhs and rhs
```

src/macros.jl
```diff
@@ -359,23 +368,46 @@ macro wrap(ex)
...
+                (x, x)
+            elseif x.head == :(::) && length(x.args) == 2
+                var = x.args[1]
```
fingolfin, line 380:
```suggestion
                # type annotations, split it and decide what to do
                var = x.args[1]
```

fingolfin (approved):
Looks good to me. Perhaps a test could be added?

@ThomasBreuer I agree that the drawback of this code is that things become ever so slightly more "magical" and harder to understand, but I feel it is worth it here. What do you think?

lgoettgens:
I added something similar as your suggestions as comments in the different cases.

> Perhaps a test could be added?

There already is a doctest coercing a julia string to a GAP string. For tests for `GAP.Obj` and `GAP.GapInt` I am afraid that I don't know enough about the types to find suitable GAP functions to test this properly.

src/macros.jl
```diff
@@ -343,6 +346,12 @@ Jacobi (generic function with 1 method)
 
 julia> Jacobi(11,35)
 1
```
fingolfin, line 348: To also test the `GapInt` auto-conversion could try this:
```suggestion
julia> Jacobi(11,35)
1

julia> Jacobi(big(35)^100+11, 35)
1
```
But not urgent.

Issue #994, opened 2024-05-22
Add CI tests that build against and test using latest GAP master


fingolfin:
... so that we can find regressions sooner.

Ideally we'd also have a mirror CI tests on the GAP repository so that PRs there can be pre-checked.

PR #995, opened 2024-05-22
Add CI job against GAP master

lgoettgens:
Resolves the GAP.jl part of https://github.com/oscar-system/GAP.jl/issues/994

This now seems to work.

~~This currently contains https://github.com/oscar-system/GAP.jl/pull/999, and should be rebased once merged.~~

### DIFF
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -1,4 +1,4 @@
-name: CI with gap master
+name: CI with gap
 
 on:
   push:
@@ -24,10 +24,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        gap-version:
+          # - 'master'
+          # - 'stable-4.13'
+          - 'v4.13.0'       # latest release
         julia-version:
-          #- '1.6'
+          # - '1.6'
           - '1.10'
-          #- 'nightly'
+          # - 'nightly'
         julia-arch:
           - x64
         os:
@@ -43,10 +47,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-        with:
-          # For Codecov, we must also fetch the parent of the HEAD commit to
-          # be able to properly deal with PRs / merges
-          fetch-depth: 2
       - name: "Set up Julia"
         uses: julia-actions/setup-julia@v2
         with:
@@ -56,14 +56,14 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: 'gap-system/gap'
-          ref: 'v4.13.0'
+          ref: ${{ matrix.gap-version }}
           path: 'GAPROOT'
       - name: "Build GAP"
         run: |
           cd GAPROOT
           ./autogen.sh
           ./configure
-          make
+          make -j`nproc`
       - name: "Override bundled GAP"
         run: |
           julia --proj=override etc/setup_override_dir.jl GAPROOT /tmp/gap_jll_override
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -1,0 +1,72 @@
+name: CI with gap master
+
+on:
+  push:
+    branches:
+      - 'master'
+      - 'release-*'
+    tags: '*'
+  pull_request:
+
+concurrency:
+  # group by workflow and ref; the last slightly strange component ensures that for pull
+  # requests, we limit to 1 concurrent job, but for the default repository branch we don't
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref_name != github.event.repository.default_branch || github.run_number }}
+  # Cancel intermediate builds, but only if it is a pull request build.
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  test:
+    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
+    continue-on-error: ${{ matrix.julia-version == 'nightly' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version:
+          #- '1.6'
+          - '1.10'
+          #- 'nightly'
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+        # include:
+          # Add a few macOS jobs (not too many, the number we can run in parallel is limited)
+          # - julia-version: '1.10'
+          #   julia-arch: x64
+          #   os: macOS-latest
+          # - julia-version: 'nightly'
+          #   julia-arch: x64
+          #   os: macOS-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # For Codecov, we must also fetch the parent of the HEAD commit to
+          # be able to properly deal with PRs / merges
+          fetch-depth: 2
+      - name: "Set up Julia"
+        uses: julia-actions/setup-julia@v2
+        with:
+          version: ${{ matrix.julia-version }}
+          arch: ${{ matrix.julia-arch }}
+      - name: Checkout GAP
+        uses: actions/checkout@v4
+        with:
+          repository: 'gap-system/gap'
+          ref: 'master'
+          path: 'GAPROOT'
+      - name: "Build GAP"
+        run: |
+          cd GAPROOT
+          ./autogen.sh
+          make
+      - name: "Override bundled GAP"
+        run: |
+          julia --proj=override etc/setup_override_dir.jl GAPROOT /tmp/gap_jll_override
+      - name: "Run tests"
+        run: |
+           julia --proj=override etc/run_with_override.jl /tmp/gap_jll_override --depwarn=error -e "using Pkg; Pkg.test()"
+
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -69,5 +69,5 @@ jobs:
           julia --proj=override etc/setup_override_dir.jl GAPROOT /tmp/gap_jll_override
       - name: "Run tests"
         run: |
-           julia --proj=override etc/run_with_override.jl /tmp/gap_jll_override --depwarn=error -e "using Pkg; Pkg.test()"
+           julia --proj=override etc/run_with_override.jl /tmp/gap_jll_override --depwarn=error -e "using Pkg; Pkg.test(\"GAP\")"
 
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   test:
-    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }} - GAP ${{ matrix.gap-version }}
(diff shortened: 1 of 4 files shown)


.github/workflows/gap.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: "Set up Julia"
+        uses: julia-actions/setup-julia@v2
+        with:
+          version: ${{ matrix.julia-version }}
+          arch: ${{ matrix.julia-arch }}
+      - name: Checkout GAP
+        uses: actions/checkout@v4
+        with:
+          repository: 'gap-system/gap'
+          ref: 'v4.13.0'
```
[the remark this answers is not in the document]
lgoettgens: This is more or less what I planned to do anyway. I just put 4.13.0 here for the time being since I know that the tests should succeed with this gap version as it is exactly the same as what Gap_jll gives us anyway. Once this works, I'll add this to the matrix

etc/run_with_override.jl
```diff
@@ -12,21 +12,23 @@ gapoverride = abspath(gapoverride)
 #
 @info "Install needed packages"
 using Pkg
-Pkg.add(["GAP_jll"])
-Pkg.add(["GAP_lib_jll"])
 Pkg.develop(path=dirname(dirname(@__FILE__)))
+Pkg.add(["GAP_jll", "GAP_lib_jll"])
```
lgoettgens, line 16: This changed order of additions makes a little bit faster to precompile

etc/run_with_override.jl
```diff
@@ -12,21 +12,23 @@ gapoverride = abspath(gapoverride)
...
+
 #
 #
 #
-function add_jll_override(depot, pkgname, newdir)
+function add_jll_override(depot, pkgname, pkguuid, newdir)
     pkgid = Base.identify_package("$(pkgname)_jll")
-    uuid = string(pkgid.uuid)
+    # does not work with julia 1.12-dev
+    # pkguuid = string(pkgid.uuid)
```
lgoettgens, line 27: This threw some (seemingly random) segfaults with nightly. But only if running from the console, not if copying everything to the reply.
Thus I took the easy solution and added the UUID to the arguments instead

etc/run_with_override.jl
```diff
@@ -40,8 +42,14 @@ tmpdepot = mktempdir(; cleanup=true)
 @info "Created temporary depot at $(tmpdepot)"
 
 # create override file for GAP_jll
-add_jll_override(tmpdepot, "GAP", gapoverride)
-add_jll_override(tmpdepot, "GAP_lib", gapoverride)
+add_jll_override(tmpdepot, "GAP", "5cd7a574-2c56-5be2-91dc-c8bc375b9ddf", gapoverride)
+add_jll_override(tmpdepot, "GAP_lib", "de1ad85e-c930-5cd4-919d-ccd3fcafd1a3", gapoverride)
+
+# trivially change JuliaInterface for rebuild
+run(`touch $(abspath(dirname(dirname(@__FILE__)), "pkg", "JuliaInterface", "src", "JuliaInterface.c"))`)
```
lgoettgens, line 49: that should be enough, shouldn't it?

etc/run_with_override.jl
```diff
@@ -40,8 +42,14 @@ tmpdepot = mktempdir(; cleanup=true)
...
+
+# HACK: use the documentation from GAP_lib_jll instead of rebuilding it
+run(`ln -sf $(abspath(GAP_lib_jll.find_artifact_dir(), "share", "gap", "doc")) $(abspath(gapoverride, "share", "gap", "doc"))`)
```
lgoettgens, line 52: this indeed works for the help system tests

lgoettgens (commented):
it now works more or less for ubuntu. for macos there is some issue with JuliaInterface that I don't know how to handle

etc/run_with_override.jl
fingolfin: no, it computes a tree hash, you need to actually change the file content, or use something else as suggested in issue #997

lgoettgens: I like the idea of #997 and am already working on it.

fingolfin:
For the macOS failures you basically need to insert `sysinfo["GAC_LDFLAGS"] = "-bundle"` into `regenerate_gaproot()`. This works around a bug in the GAP_jll build recipe and will be fixed in its next version.

src/setup.jl
```diff
@@ -118,6 +118,7 @@ function regenerate_gaproot()
     # add the necessary flags to link against libgap
     gap_lib = joinpath(gap_prefix, "lib")
     sysinfo["GAP_LDFLAGS"] = "-L$(gap_lib) -lgap"
+    sysinfo["GAC_LDFLAGS"] *= " -bundle" # Remove this line once the GAP_jll build recipe is fixed
```
fingolfin, line 121:
```suggestion
    Sys.isapple() && sysinfo["GAC_LDFLAGS"] = " -bundle" # Remove this line once the GAP_jll build recipe is fixed
```

src/setup.jl
```diff
@@ -118,6 +118,7 @@ function regenerate_gaproot()
     # add the necessary flags to link against libgap
     gap_lib = joinpath(gap_prefix, "lib")
     sysinfo["GAP_LDFLAGS"] = "-L$(gap_lib) -lgap"
+    Sys.isapple() && sysinfo["GAC_LDFLAGS"] = " -bundle" # Remove this line once the GAP_jll build recipe is fixed
```
lgoettgens, line 121:
```suggestion
    Sys.isapple() && (sysinfo["GAC_LDFLAGS"] = " -bundle") # Remove this line once the GAP_jll build recipe is fixed
```

.github/workflows/gap.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Checkout GAP
+        uses: actions/checkout@v4
+        with:
+          repository: 'gap-system/gap'
+          ref: ${{ matrix.gap-version }}
+          path: 'GAPROOT'
+      - name: "Install dependencies (for macOS)"
+        if: runner.os == 'macOS'
+        run: |
+          brew install autoconf gmp zlib pkg-config
```
lgoettgens, line 61:
```suggestion
          brew install autoconf zlib # gmp pkg-config
```
These are already installed on the runner

.github/workflows/gap.yml
```diff
@@ -0,0 +1,75 @@
...
+          - '1.6'
+          - '1.10'
+          - 'nightly'
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+          - macOS-latest
+        exclude:
+          # Remove a few macOS jobs (the number we can run in parallel is limited)
```
fingolfin, line 41: This still add 6 macOS jobs, that's too much

.github/workflows/gap.yml
```diff
@@ -0,0 +1,75 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        gap-version:
+          - 'master'
+          - 'stable-4.13'
+          - 'v4.13.0'       # latest release
+        julia-version:
+          - '1.6'
+          - '1.10'
```
fingolfin, line 33: Maybe test against the latest stable Julia release instead, i.e.
```suggestion
          - '1'
```

.github/workflows/gap.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
+    continue-on-error: ${{ matrix.julia-version == 'nightly' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        gap-version:
+          - 'master'
+          - 'stable-4.13'
+          - 'v4.13.0'       # latest release
```
fingolfin, line 30: I am not sure that's all too helpful. Point is: we may evolve some internals of the GC integration requiring changes both in GAP and in GAP.jl, and it may be impractical to also keep supporting the last stable release. I'd just drop this, to reduce the overall number of jobs.

Or at the very least, drop the macOS tests against that, we really only need to test macOS against Julia 1.10 (resp latest 1.x) with GAP master and stable-4.13

lgoettgens: Thinking again about it: the latest release is already tested via the latest GAP_jll release (more or less at least), so the version bump in GAP_jll and GAP_lib_jll can be used to concurrently change GAP and GAP.jll as you described.

etc/run_with_override.jl
fingolfin: Just tried with Julia master and when running this code I actually got
```
ERROR: LoadError: type Nothing has no field uuid
```
which would suggest `pkgid === nothing`, but then also the `touch` below would be borked?

lgoettgens: if you print the pkgid, this succeeds (and prints something non-trivial), but i then get a segfault instead of a LoadError when accessing pkgid.uuid

fingolfin: Indeed: printing the `pkgid` makes it proceed -- that seems like a big regression in Julia, printing a value shouldn't change it form `nothing` to something, should it? In particular since *this* works in the REPL (without printing):
```julia
pkgid = Base.identify_package("GAP_jll"); pkgid.uuid
```
Perhaps we should open an issue?

Anyway, so I inserted the `print` and it proceeds, but I don't get a crash, I get this:
```
[ Info: Created temporary depot at /var/folders/d_/1zss2fnd6xdgclqnj8jj_27m0000gp/T/jl_7ABU3q
pkgname GAP has pkgid Base.PkgId(Base.UUID("5cd7a574-2c56-5be2-91dc-c8bc375b9ddf"), "GAP_jll")
pkgname GAP_lib has pkgid Base.PkgId(Base.UUID("de1ad85e-c930-5cd4-919d-ccd3fcafd1a3"), "GAP_lib_jll")
Precompiling GAP...
Info Given GAP was explicitly requested, output will be shown live
ERROR: LoadError: ArgumentError: Package GAP_jll [5cd7a574-2c56-5be2-91dc-c8bc375b9ddf] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.
```
which makes no sense, as we run `Pkg.instantiate()`. But I've inserted another `Pkg.instantiate()` invocation right after the `add_jll_override` calls.

fingolfin: Thanks for [reporting it upstream](https://github.com/JuliaLang/julia/issues/54599), it seems we've hit an IMHO rather serious looking code generation bug in Julia.

For now we can keep this workaround, but let's remove it again when they fix their bug.

etc/run_with_override.jl
```diff
@@ -12,21 +12,23 @@ gapoverride = abspath(gapoverride)
...
+import GAP_lib_jll
+
 #
 #
 #
-function add_jll_override(depot, pkgname, newdir)
+function add_jll_override(depot, pkgname, pkguuid, newdir)
     pkgid = Base.identify_package("$(pkgname)_jll")
-    uuid = string(pkgid.uuid)
+    # does not work with julia 1.12-dev
```
fingolfin, line 26:
```suggestion
    # does not work with julia 1.12-dev, see https://github.com/JuliaLang/julia/issues/54599
    # TODO: remove pkguui dargument again once that bug is fixed
```

fingolfin (approved):

PR #998, opened 2024-05-27
Fix force_symlink documentation, argument names

fingolfin:
The argument names now match those for Base.Filesystem.rename, and the documentation was fixed (it had the arguments reversed)


lgoettgens (approved):
the use-sites of this already seem to honour this new direction. this would lead to nasty  bugs if someone would have looked at this function...

fingolfin:
Yeah I wanted to use the function, looked at the comment and code and was reeeaaaaly confused

Issue #997, opened 2024-05-27
Using `etc/run_with_override.jl` should force use of a custom compiled JuliaInterface


fingolfin:
Basically `locate_JuliaInterface_so` should have some logic that makes it compiled and load the bundled JuliaInterface source code instead of the JLL. E.g. it could check an env variable `FORCE_JULIAINTERFACE_COMPILATION`, and `etc/run_with_override.jl` could set that.

PR #999, opened 2024-05-27
Add `FORCE_JULIAINTERFACE_COMPILATION`

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/997.

I verified by looking at debug output that just running the GAP.jl master uses the jll-version while running `julia --proj=override etc/run_with_override.jl /tmp/gap_jll_override --depwarn=error -e "ENV[\"JULIA_DEBUG\"] = all; using GAP"` prints `[Info: Compiling JuliaInterface...`

### DIFF
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -232,12 +232,12 @@ function locate_JuliaInterface_so(sysinfo::Dict{String, String})
     jll_hash = GitTools.tree_hash(joinpath(jll, "src"))
     bundled = joinpath(@__DIR__, "..", "pkg", "JuliaInterface")
     bundled_hash = GitTools.tree_hash(joinpath(bundled, "src"))
-    if jll_hash == bundled_hash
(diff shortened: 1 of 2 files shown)


src/setup.jl
```diff
@@ -232,12 +232,12 @@ function locate_JuliaInterface_so(sysinfo::Dict{String, String})
     jll_hash = GitTools.tree_hash(joinpath(jll, "src"))
     bundled = joinpath(@__DIR__, "..", "pkg", "JuliaInterface")
     bundled_hash = GitTools.tree_hash(joinpath(bundled, "src"))
-    if jll_hash == bundled_hash
+    if jll_hash == bundled_hash && get(ENV, "FORCE_JULIAINTERFACE_COMPILATION", "false") != "true"
```
fingolfin, line 235: Usually I'd just check for the presence of such a variable and not care about the content. But since this is not intended for user consumption, I am fine either way.

fingolfin (approved):

PR #1009, opened 2024-06-19
tell JuliaInterface to load GAPDoc

ThomasBreuer:
When `ENV["GAP_BARE_DEPS"]` is bound before GAP.jl gets loaded, the JuliaInterface package will be loaded before GAPDoc, and this causes a warning because JuliaInterface needs a variable from GAPDoc.
In order to avoid this, add GAPDoc to the dependencies of JuliaInterface.

### DIFF
--- a/pkg/JuliaInterface/PackageInfo.g
+++ b/pkg/JuliaInterface/PackageInfo.g
@@ -101,6 +101,7 @@ Dependencies := rec(
   GAP := ">= 4.11",    # need compatible code in GAP's src/julia_gc.c
   NeededOtherPackages := [ ],
   SuggestedOtherPackages := [ ],
+  OtherPackagesLoadedInAdvance := [ [ "GAPDoc", "1.6.6" ] ], # need StripEscapeSequences
   ExternalConditions := [ ],
 ),
 


pkg/JuliaInterface/PackageInfo.g
```diff
@@ -101,6 +101,7 @@ Dependencies := rec(
   GAP := ">= 4.11",    # need compatible code in GAP's src/julia_gc.c
   NeededOtherPackages := [ ],
   SuggestedOtherPackages := [ ],
+  OtherPackagesLoadedInAdvance := [ [ "GAPDoc", "1.6.6" ] ], # need StripEscapeSequences
```
fingolfin, line 104: I don't love this, `OtherPackagesLoadedInAdvance` is obscure and the code using `StripEscapeSequences` is unclear to me (i.e. why `StripEscapeSequences` is even needed there is not immediately visible). And to start with, why `StripEscapeSequences` isn't in the GAP library (which calls it, too! only in the help system, though).

But it's a pragmatic solution for the moment, so let's merge it.

ThomasBreuer: The point is that we cannot add GAPDoc to the needed packages of JuliaInterface because JuliaInterface is a needed package of GAP, and GAP does not allow us to silently add needed packages of GAP. (In this sense, the fact that using OtherPackagesLoadedInAdvance is possible at all can be regarded as a bug in GAP.)

Concerning the GAPDoc code in question, I agree that it should better be in the GAP library, but I cannot simply change this.

If I remember well then the idea behind needed packages of GAP was (besides the use case GAPDoc) that one could eventually split off more parts of the GAP library into packages.

Issue #1005, opened 2024-06-18
Run `JuliaInterface` tests as part of `Pkg.test`?


lgoettgens:
or at least in PkgEval runs?
In https://github.com/JuliaLang/julia/pull/54678, a PkgEval job was ran, and PRs to all broken packages were supplied. However, since this only ran the GAP.jl tests and not those of JuliaInterface, https://github.com/oscar-system/GAP.jl/issues/1004 was not found there.
I think it would make sense to have the `JuliaInterface` tests being run as part of PkgEval jobs to make the julia devs aware of possible future breakage they introduce.
What do you think?

fingolfin:
I have no idea how to run tests "at least in PkgEval runs".

If running JuliaInterface tests as part of `Pkg.test` seems useful, sure, why not (PR welcome ;-) ). But we also will want to keep the current test for them because they (a) test the generated `gap.sh` wrappers, and (b) provide code coverage information.

lgoettgens:
> I have no idea how to run tests "at least in PkgEval runs".

Iirc they set some environment variable. I'll look into this later

ThomasBreuer:
My interpretation of the proposal is as follows:

We extend the file `GAP.jl/test/runtests.jl` by a `@test` line that executes the contents of `GAP.jl/pkg/JuliaInterface/tst/testall.g`, except that GAP shall not be quit in the end, and that we need a return value.
This additional `@test` line shall be executed only under the condition `haskey(ENV, "JULIA_PKGEVAL")`.

If this is correct then `GAP.jl/test/runtests.jl` should be modified such that it can be used in both situations, i. e., when it is called from GAP or from Julia.

fingolfin:
I would prefer not to call `JULIA_PKGEVAL` and just always run those test `@test` lines: they add very little extra time for the tests, and doing it this way is much easier to verify.

To ensure GAP does not quit at the end of `tst/testall.g`, I see two options:
1. modify the two `testall.g` files to look for some global flag (e.g. check if `DONT_QUIT_AFTER_TEST` is bound and set to `true`) to modify the `exitGAP` flag value (and to also not execute the final `FORCE_QUIT_GAP`).  Then we still need the return value of `TestDirectory` which could be assigned to another global variable.
[...]

Issue #1011, opened 2024-06-20
 Error handling incomplete / clashes with GAP's Test/TestDirectory, leaves GAP in bad state, subsequent crash


fingolfin:
Tried to run the GAP test suite in Julia. This failed because it includes testing errors and we hijack them (so the code in GAP's `Test` which also "hijacks" error handling has no chance of running - here the very limited error handling model in GAP bites us, unlike try/catch in Julia there is no proper nesting mechanism... perhaps we could at least deal with this particular use case in `Test`, but that's a separate question for the future).

Anyway, after this I thought "maybe it works from a GAP prompt", so I launched one... and it crashes. Presumably because a bunch of state is not reset when we yank control away from `Test` / `RunTests` / `READ_STREAM_LOOP`:

```
julia> @time GAP.Globals.TestDirectory(GapObj("/Users/mhorn/Projekte/GAP/gap.spielwiese/tst/testinstall"))
Architecture: aarch64-apple-darwin20-julia1.10-64-kv9

testing: /Users/mhorn/Projekte/GAP/gap.spielwiese/tst/testinstall/ConjNatSym.tst
      31 ms (0 ms GC) and 24.1MB allocated for ConjNatSym.tst
testing: /Users/mhorn/Projekte/GAP/gap.spielwiese/tst/testinstall/DirectProductElement.tst
# line 105 of 131 (80%)ERROR: Error thrown by GAP: Error, <index> too large for <dpelm>, you may return another index

Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] ThrowObserver(depth::Int32)
   @ GAP ~/Projekte/OSCAR/GAP.jl/src/GAP.jl:89

julia> GAP.prompt()

[34215] signal (11.2): Segmentation fault: 11
in expression starting at REPL[3]:1
putToTheStream at /Users/mhorn/.julia/artifacts/2179238659fb71bb09503491ad5113d9c35daacc/lib/libgap.9.dylib (unknown line)
Pr at /Users/mhorn/.julia/artifacts/2179238659fb71bb09503491ad5113d9c35daacc/lib/libgap.9.dylib (unknown line)
GetLine at /Users/mhorn/.julia/artifacts/2179238659fb71bb09503491ad5113d9c35daacc/lib/libgap.9.dylib (unknown line)
[...]

PR #1015, opened 2024-06-21
Run JuliaInterface tests from runtests.jl via GAP integration

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/1005 via alternative 2 described in https://github.com/oscar-system/GAP.jl/issues/1005#issuecomment-2180401958.
Alternative to and thus closes https://github.com/oscar-system/GAP.jl/pull/1014.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -24,3 +24,13 @@ end
 @testset "manual examples" begin
   include("doctest.jl")
 end
+
+@testset "JuliaInterface tests" begin
+  GAP.Wrappers.Read(joinpath(dirname(dirname(pathof(GAP))), "pkg/JuliaInterface/tst/testsetup.g"))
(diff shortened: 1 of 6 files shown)


test/runtests.jl
```diff
@@ -24,3 +24,13 @@ end
 @testset "manual examples" begin
   include("doctest.jl")
 end
+
+@testset "JuliaInterface tests" begin
+  GAP.Wrappers.Read(joinpath(dirname(dirname(pathof(GAP))), "pkg/JuliaInterface/tst/testsetup.g"))
+  GAP.Globals.TestDirectory(GAP.Globals.dir, GapObj(Dict(:exitGAP => false, :testOptions => Dict(:compareFunction => GAP.Globals.compare)); recursive=true))
```
ThomasBreuer, line 30: `TestDirectory` returns `true` or `false`, thus I think we can call `@test GAP.Globals.TestDirectory(...)`.

ThomasBreuer:
The failures are due to the problem described in #1011: The JuliaInterface tests involve tests of GAP errors.

PR #1017, opened 2024-06-27
change the handling of `IncludeJuliaStartupFile` (for Julia 1.10)

ThomasBreuer:
The JuliaInterface package introduces a GAP user preference `IncludeJuliaStartupFile` that is intended to force loading Julia's startup file in the situation that GAP gets started from Julia via a gap.sh script.
The file `pkg/JuliaInterface/init.g` contains code that evaluates this user preferences and if necessary loads this file into the Julia session.

All this should be irrelevant if one first starts Julia and then loads GAP, because then the Julia startup file has been read already, but this is not true:
If one calls `using Oscar` in a fresh Julia session *and* if `Oscar.jl` has to get precompiled then the precompilation takes place in a Julia that has not read its startup file.
In Julia before 1.10, then JuliaInterface simply includes this startup file. In Julia 1.10, including this file fails (and then one gets a crash).

The current proposal is not yet good enough.
An info message is printed during precompilation, saying that the startup file could not be included. We can simply omit this message. However, I fear that the above analysis is not yet good enough, so perhaps we need another change.


ThomasBreuer:
Two tests with Julia nightly failed, I restarted them, then only one test failed ...

src/setup.jl
```diff
@@ -286,7 +286,13 @@ function create_gap_sh(dstdir::String)
         # But you can always instead load this file as if it was a .jl file via any
         # other Julia executable.
         #=
-        exec $(joinpath(Sys.BINDIR, Base.julia_exename())) --startup-file=no --project=$(dstdir) -i -- "$(gap_sh_path)" "\$@"
+        if [ "\$JULIA_STARTUP_FILE_IN_GAP" = "yes" ];
+        then
+            READ_STARTUP_FILE="yes";
+        else
+            READ_STARTUP_FILE="no";
```
fingolfin, line 293:
```suggestion
            READ_STARTUP_FILE="no"
```

src/setup.jl
```diff
@@ -286,7 +286,13 @@ function create_gap_sh(dstdir::String)
         # But you can always instead load this file as if it was a .jl file via any
         # other Julia executable.
         #=
-        exec $(joinpath(Sys.BINDIR, Base.julia_exename())) --startup-file=no --project=$(dstdir) -i -- "$(gap_sh_path)" "\$@"
+        if [ "\$JULIA_STARTUP_FILE_IN_GAP" = "yes" ];
+        then
+            READ_STARTUP_FILE="yes";
```
fingolfin, line 291:
```suggestion
            READ_STARTUP_FILE="yes"
```

src/setup.jl
```diff
@@ -286,7 +286,13 @@ function create_gap_sh(dstdir::String)
         # But you can always instead load this file as if it was a .jl file via any
         # other Julia executable.
         #=
-        exec $(joinpath(Sys.BINDIR, Base.julia_exename())) --startup-file=no --project=$(dstdir) -i -- "$(gap_sh_path)" "\$@"
+        if [ "\$JULIA_STARTUP_FILE_IN_GAP" = "yes" ];
```
fingolfin, line 289:
```suggestion
        if [ "\$JULIA_STARTUP_FILE_IN_GAP" = "yes" ]
```

fingolfin (approved):

PR #1021, opened 2024-07-05
Provide a stale_age to mkpidlock

fingolfin:
Normally the pidlock should be held for just a few milliseconds; waiting for its age to exceed 10 seconds until we consider it stale should be very safe. In fact Julia waits 5 times longer if the process creating the pid lock file seems to be still running.

On the other hand, without a stale age, the lock file is *never* considered stale, even if the process creating it definitely is gone, and so the user can get stuck, which obviously is very bad. To get unstuck they need to manually delete the lock file.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -243,7 +243,9 @@ function install(spec::String, version::String = "";
     # point PackageManager to the given pkg dir
     Globals.PKGMAN_CustomPackageDir = GapObj(pkgdir)
     mkpath(pkgdir)
-    Pidfile.mkpidlock("$pkgdir.lock") do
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
There is another call of `Pidfile.mkpidlock` in `src/packages.jl`. Should this be changed in the same way?

fingolfin:
@ThomasBreuer ah I somehow missed that one. I've now adjusted it as well but with a higher timeout (5 minutes resp. 25 minutes if the process creating the lock file is still running)

lgoettgens (approved):

ThomasBreuer (approved):

PR #1022, opened 2024-07-05
CI: test against Julia 1.11-rc1

fingolfin:
... and stop testing against 1.9


lgoettgens:
I think it would be great to get this in before the julia 1.11 release, so we still have some opportunity to report problems with the latest rc

Issue #1025, opened 2024-08-14
`GapObj` vs. `julia_to_gap`


ThomasBreuer:
Originally we installed methods for `GapObj` in situations where we wanted to *convert* the given Julia object to GAP.
Meanwhile we use `GapObj` in Oscar also for *fetching* the "underlying GAP object" stored in an Oscar object.

(On the one hand, this is better than accessing some field (with non-uniform name) in the object.
On the other hand, one cannot decide from the code whether a call of `GapObj` will just fetch a known value, or will compute the value in the first call (and then store it), or will create a new object on the GAP side in each call.)

For the "fetching" aspect of `GapObj`, sometimes we install in fact a `GAP.julia_to_gap(x)` method instead of a `GapObj(x)` method, in order to cover automatically also the situation `GapObj([x, y, z]; recursive = true)`.
The delegations between the methods in GAP.jl are not really nice:
Typically, `GapObj(x, recursive = false)` calls `julia_to_gap(x, IdDict(); recursive = false)`, and *if there is no special method for these arguments* then this delegates to the generic `julia_to_gap(x)`.

Here is a situation where this setup causes a problem:
```
using Oscar
g = symmetric_group(2);
T = right_transversal(g, g);
GapObj(T)
```
If a version of GAP.jl is used that includes #989 then we get a stack overflow error.
(Currently Oscar.jl wants an older version of GAP.jl, thus one cannot run into this problem without modifying Oscar.jl's `Project.toml`.)
The point is that the object `T` above is an `AbstractVector`, and the new GAP.jl provides a special `julia_to_gap` method for such objects.
An obvious solution is to provide also a `GapObj(T)` method in Oscar, but it would be better to provide better support in GAP.jl.

(The CI tests do not find the above problem. There is a test of GAP.jl together with the current Oscar.jl, but since Oscar.jl does not admit the dev version of GAP.jl, this test is skipped, and reports success.)

fingolfin:
@ThomasBreuer I believe this has been resolved by your PR #1029 ? If you agree, please close; otherwise: what exactly is missing?

PR #1029, opened 2024-08-30
rewrite `julia_to_gap`

ThomasBreuer:
This is a first attempt to issue #1025.
The ideas are the same as in pull request #777, which deals with "the other direction", that is, the conversion from GAP to Julia:

- Make the recursive conversion safe. Up to now, we have ad-hoc delegation from `julia_to_gap` methods with three arguments back to a `julia_to_gap` with only one argument.
- Make the conversion more efficient. Up to now, a dictionary for tracking subobjects is created by default in the methods in question, also in situations where this dictionary is not used later on.
- Make the installation of new conversion methods simpler. Up to now it happens (for example in Oscar.jl) that the installation of a `GapObj` method for some Oscar objects does not work as expected for vectors of these objects, or that changes in GAP.jl break the intended behaviour of `GapObj` methods in Oscar.jl (see #1025).

The conversion from Julia to GAP is simpler than that from GAP to Julia for example because we need not care about caching different Julia target types for the same GAP object. On the other hand, the direction from Julia to GAP shall cover also the constructor `GapObj` in situations where one wants to fetch an existing GAP object stored in a Julia object.

The current changes mean that all Julia to GAP conversions have to be installed via `GapObj_internal` methods that have three arguments.
Then `GapObj` needs only default methods that delegate to `GapObj_internal`.

If we proceed like this then these changes are breaking.
For Oscar.jl, they mean that the `julia_to_gap` and `GapObj` methods installed there have to be rewritten.
(Currently the master branch of Oscar.jl has 23 `julia_to_gap` methods, half of them in `src/GAP/oscar_to_gap`.)

Here are the technical details.

- introduce `GapObj_internal`, which takes always three arguments: the Julia object to be converted, the auxiliary cache object, and a Boolean indicating whether recursive conversion is wanted

- turn all `julia_to_gap` methods into `GapObj_internal` methods

- define default `GapObj` methods (with 1 to 3 arguments) that call `GapObj_internal`

- admit `nothing` as an allowed value for the auxiliary cache object, and create dictionary objects only inside the methods in question where they are needed; introduce the type `GapCacheDict` for that

- provide a macro for the installation of unary `GapObj` methods (i.e., methods that do not require the recursive tracking of subobjects w.r.t. object identity) via the syntax `GAP.@install GapObj(x::T) = f(x)`

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -15,6 +15,8 @@ An internal type of GAP.jl used for tracking conversion results in `gap_to_julia
 """
 const RecDict = IdDict{Any,Any}
 
+const GapCacheDict = Union{Nothing,RecDict}
+
 ## Conversion from GAP to Julia
 """
     gap_to_julia(type, x, recursion_dict::Union{Nothing,RecDict}=nothing; recursive::Bool=true)
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -85,27 +85,19 @@ function GapObj_internal end
 # to a GAP record.
 # Thus we have a default method returning `true`.
 #
-# Methods for those types that want `false` have to be installed;
-# the `GapObj` methods arising from `GAP.@install` calls are unary,
-# therefore the macro automatically installs such a
-# `_needs_tracking_julia_to_gap` method.
+# Methods for those types that want `false` have to be installed explicitly.
+# The `GapObj` methods arising from `GAP.@install` don't handle recursion
+# and the macro automatically installs such a `_needs_tracking_julia_to_gap`
+# method.
 _needs_tracking_julia_to_gap(::Any) = true
 
 
 GAP.@install GapObj(x::FFE) = x    # Default for actual GAP objects is to do nothing
 GAP.@install GapObj(x::Bool) = x    # Default for actual GAP objects is to do nothing
 
-## Integers: general case first deal with things that fit into immediate
-## integers, then falls back to converting to BigInt and calling into the GAP
-## kernel API.
-## TODO: we could provide more efficient conversion for UInt64, Int128, UInt128
-## which avoids the conversion to BigInt, if we wanted to.
-function GapObj_internal(x::Integer, cache::GapCacheDict, ::Val{recursive}) where recursive
-    # if it fits into a GAP immediate integer, convert x to Int64
-    x in -1<<60:(1<<60-1) && return Int64(x)
-    # for the general case, fall back to BigInt
-    return GapObj_internal(BigInt(x), cache, Val(recursive))
-end
+## Integers:
+## We do not want to track conversion for any concrete integer types.
+@install GapObj(x::Integer) = x in -1<<60:(1<<60-1) ? Int64(x) : GapObj(BigInt(x))
 
 ## Small integers types always fit into GAP immediate integers, and thus are
 ## represented by Int64 on the Julia side.
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -46,11 +46,11 @@ The following `julia_to_gap` conversions are supported by GAP.jl.
 | `UnitRange{T}`, `StepRange{T, S}`    | `IsRange`    |
 | `Function`                           | `IsFunction` |
 """
-julia_to_gap(x, cache::GapCacheDict = nothing; recursive::Bool = false) = julia_to_gap_internal(x, cache, recursive)
+julia_to_gap(x, cache::GapCacheDict = nothing; recursive::Bool = false) = GapObj_internal(x, cache, recursive)
 
-# The calls to `GAP.@install` install methods for `julia_to_gap_internal`
+# The calls to `GAP.@install` install methods for `GapObj_internal`
 # so we must make sure it is declared before
-function julia_to_gap_internal end
+function GapObj_internal end
 
 GAP.@install GapObj(x::FFE) = x    # Default for actual GAP objects is to do nothing
 GAP.@install GapObj(x::Bool) = x    # Default for actual GAP objects is to do nothing
@@ -60,11 +60,11 @@ GAP.@install GapObj(x::Bool) = x    # Default for actual GAP objects is to do no
 ## kernel API.
 ## TODO: we could provide more efficient conversion for UInt64, Int128, UInt128
 ## which avoids the conversion to BigInt, if we wanted to.
-function julia_to_gap_internal(x::Integer, cache::GapCacheDict, recursive::Bool)
+function GapObj_internal(x::Integer, cache::GapCacheDict, recursive::Bool)
     # if it fits into a GAP immediate integer, convert x to Int64
     x in -1<<60:(1<<60-1) && return Int64(x)
     # for the general case, fall back to BigInt
-    return julia_to_gap_internal(BigInt(x), cache, recursive)
+    return GapObj_internal(BigInt(x), cache, recursive)
 end
 
 ## Small integers types always fit into GAP immediate integers, and thus are
@@ -89,7 +89,7 @@ GAP.@install function GapObj(x::BigInt)
 end
 
 ## Rationals
-function julia_to_gap_internal(x::Rational{T}, cache::GapCacheDict, recursive::Bool) where {T<:Integer}
+function GapObj_internal(x::Rational{T}, cache::GapCacheDict, recursive::Bool) where {T<:Integer}
     denom_julia = denominator(x)
     numer_julia = numerator(x)
     if denom_julia == 0
@@ -99,8 +99,8 @@ function julia_to_gap_internal(x::Rational{T}, cache::GapCacheDict, recursive::B
             return -Globals.infinity
         end
     end
-    numer = julia_to_gap_internal(numer_julia, cache, recursive)
-    denom = julia_to_gap_internal(denom_julia, cache, recursive)
+    numer = GapObj_internal(numer_julia, cache, recursive)
+    denom = GapObj_internal(denom_julia, cache, recursive)
     return Wrappers.QUO(numer, denom)
 end
 
@@ -117,7 +117,7 @@ GAP.@install GapObj(x::AbstractString) = MakeString(string(x))
 GAP.@install GapObj(x::Symbol) = MakeString(string(x))
 
 ## Arrays (including BitVector)
-function julia_to_gap_internal(
+function GapObj_internal(
     obj::AbstractVector{T},
     recursion_dict::GapCacheDict,
     recursive::Bool,
@@ -141,7 +141,7 @@ function julia_to_gap_internal(
         end
         if recursive
             res = get!(recursion_dict::RecDict, x) do
-                julia_to_gap_internal(x, recursion_dict::RecDict, recursive)
+                GapObj_internal(x, recursion_dict::RecDict, recursive)
             end
         else
             res = x
@@ -152,7 +152,7 @@ function julia_to_gap_internal(
 end
 
 ## Convert two dimensional arrays
-function julia_to_gap_internal(
+function GapObj_internal(
     obj::Matrix{T},
     recursion_dict::GapCacheDict,
     recursive::Bool,
@@ -171,19 +171,19 @@ function julia_to_gap_internal(
     for i = 1:rows
         # Note that we need not check whether the row is in `recursion_dict`
         # because we are just now creating the object.
-        ret_val[i] = julia_to_gap_internal(obj[i, :], recursion_dict, recursive)
+        ret_val[i] = GapObj_internal(obj[i, :], recursion_dict, recursive)
     end
     return ret_val
 end
 
 ## Tuples
-function julia_to_gap_internal(
+function GapObj_internal(
     obj::Tuple,
     recursion_dict::GapCacheDict,
     recursive::Bool,
 )
     array = collect(Any, obj)
-    return julia_to_gap_internal(array, recursion_dict, recursive)
+    return GapObj_internal(array, recursion_dict, recursive)
 end
 
 ## Ranges
@@ -194,7 +194,7 @@ GAP.@install function GapObj(r::AbstractRange{<:Integer})
 end
 
 ## Dictionaries
-function julia_to_gap_internal(
+function GapObj_internal(
     obj::Dict{T,S},
     recursion_dict::GapCacheDict,
     recursive::Bool,
@@ -211,7 +211,7 @@ function julia_to_gap_internal(
         x = Wrappers.RNamObj(MakeString(string(x)))
         if recursive
             res = get!(recursion_dict::RecDict, y) do
-                julia_to_gap_internal(y, recursion_dict::RecDict, recursive)
+                GapObj_internal(y, recursion_dict::RecDict, recursive)
             end
         else
             res = y
@@ -227,7 +227,7 @@ end
 ## and if `obj` contains Julia subobjects;
 ## in this case, `obj` is a GAP list or record.
 ## An example of such an `obj` is `GAP.julia_to_gap([[1]])`.
-function julia_to_gap_internal(
+function GapObj_internal(
     obj::GapObj,
     recursion_dict::GapCacheDict,
     recursive::Bool,
@@ -244,7 +244,7 @@ function julia_to_gap_internal(
         for i = 1:len
             x = obj[i]
             ret_val[i] = get!(recursion_dict::RecDict, x) do
-                julia_to_gap_internal(x, recursion_dict::RecDict, recursive)
+                GapObj_internal(x, recursion_dict::RecDict, recursive)
             end
         end
     elseif Wrappers.IsRecord(obj)
@@ -257,7 +257,7 @@ function julia_to_gap_internal(
             x = Wrappers.RNamObj(xx)
             y = Wrappers.ELM_REC(obj, x)
             res = get!(recursion_dict::RecDict, y) do
-                julia_to_gap_internal(y, recursion_dict::RecDict, recursive)
+                GapObj_internal(y, recursion_dict::RecDict, recursive)
             end
             Wrappers.ASS_REC(ret_val, x, res)
         end
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -1,9 +1,9 @@
 ## Converters
 """
-    julia_to_gap(input, recursion_dict = IdDict(); recursive::Bool = false)
+    julia_to_gap(input, recursion_dict::GapCacheDict = nothing; recursive::Bool = false)
 
 Convert a julia object `input` to an appropriate GAP object.
-If `recursive` is set to `true`, recursive conversions on
+If `recursive` is set to `true`, recursive conversion on
 arrays, tuples, and dictionaries is performed.
 
 The input `recursion_dict` should never be set by the user, it is meant to keep egality
@@ -46,12 +46,10 @@ The following `julia_to_gap` conversions are supported by GAP.jl.
 | `UnitRange{T}`, `StepRange{T, S}`    | `IsRange`    |
 | `Function`                           | `IsFunction` |
 """
-function julia_to_gap end
-
-# default
 julia_to_gap(x, cache::GapCacheDict = nothing; recursive::Bool = false) = julia_to_gap_internal(x, cache, recursive)
 
 # The calls to `GAP.@install` install methods for `julia_to_gap_internal`
+# so we must make sure it is declared before
 function julia_to_gap_internal end
 
 GAP.@install GapObj(x::FFE) = x    # Default for actual GAP objects is to do nothing
@@ -142,11 +140,13 @@ function julia_to_gap_internal(
             continue
         end
         if recursive
-            x = get!(recursion_dict, x) do
-                julia_to_gap_internal(x, recursion_dict, recursive)
+            res = get!(recursion_dict::RecDict, x) do
+                julia_to_gap_internal(x, recursion_dict::RecDict, recursive)
             end
+        else
+            res = x
         end
-        ret_val[i] = x
+        ret_val[i] = res
     end
     return ret_val
 end
@@ -169,6 +169,8 @@ function julia_to_gap_internal(
         recursion_dict[obj] = ret_val
     end
     for i = 1:rows
+        # Note that we need not check whether the row is in `recursion_dict`
+        # because we are just now creating the object.
         ret_val[i] = julia_to_gap_internal(obj[i, :], recursion_dict, recursive)
     end
     return ret_val
@@ -208,11 +210,13 @@ function julia_to_gap_internal(
     for (x, y) in obj
         x = Wrappers.RNamObj(MakeString(string(x)))
         if recursive
-            y = get!(recursion_dict, y) do
-                julia_to_gap_internal(y, recursion_dict, recursive)
+            res = get!(recursion_dict::RecDict, y) do
+                julia_to_gap_internal(y, recursion_dict::RecDict, recursive)
             end
+        else
+            res = y
         end
-        Wrappers.ASS_REC(record, x, y)
+        Wrappers.ASS_REC(record, x, res)
     end
 
     return record
@@ -238,7 +242,10 @@ function julia_to_gap_internal(
         end
         recursion_dict[obj] = ret_val
         for i = 1:len
-             ret_val[i] = julia_to_gap_internal(obj[i], recursion_dict, recursive)
+            x = obj[i]
+            ret_val[i] = get!(recursion_dict::RecDict, x) do
+                julia_to_gap_internal(x, recursion_dict::RecDict, recursive)
+            end
         end
     elseif Wrappers.IsRecord(obj)
         ret_val = NewPrecord(0)
@@ -248,7 +255,11 @@ function julia_to_gap_internal(
         recursion_dict[obj] = ret_val
         for xx in Wrappers.RecNames(obj)::GapObj
             x = Wrappers.RNamObj(xx)
-            Wrappers.ASS_REC(ret_val, x, julia_to_gap_internal(Wrappers.ELM_REC(obj, x), recursion_dict, recursive))
+            y = Wrappers.ELM_REC(obj, x)
+            res = get!(recursion_dict::RecDict, y) do
+                julia_to_gap_internal(y, recursion_dict::RecDict, recursive)
+            end
+            Wrappers.ASS_REC(ret_val, x, res)
         end
     else
         ret_val = obj
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -1,31 +1,53 @@
 ## Converters
 """
-    julia_to_gap(input, recursion_dict::GapCacheDict = nothing; recursive::Bool = false)
+    GapObj(input, recursion_dict::GapCacheDict = nothing; recursive::Bool = false)
+    GapObj(input, recursive::Bool = false)
 
-Convert a julia object `input` to an appropriate GAP object.
-If `recursive` is set to `true`, recursive conversion on
-arrays, tuples, and dictionaries is performed.
+One can use the type [`GapObj`](@ref) as a constructor,
+in order to convert the julia object `input` to an appropriate GAP object.
+
+If `recursive` is set to `true`, recursive conversion of nested Julia objects
+(arrays, tuples, and dictionaries) is performed.
+
 
 The input `recursion_dict` should never be set by the user, it is meant to keep egality
 of input data, by converting equal data to identical objects in GAP.
 
 # Examples
 ```jldoctest
-julia> GAP.julia_to_gap(1//3)
+julia> GapObj(1//3)
 GAP: 1/3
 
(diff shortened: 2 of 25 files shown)


ThomasBreuer:
@fingolfin If it is of interest:
With GAP.jl as given by this pull request, I get reproducible segmentation faults on my notebook when I enter the following in Julia 1.10.4.
```
using GAP
using Test
include("packages/GAP.jl/test/runtests.jl")
```
The log looks similar to the one from the failed CI tests on github ([Julia 1.10 - ubuntu-latest](https://github.com/oscar-system/GAP.jl/actions/runs/10630211128/job/29468633163?pr=1029#logs)), but I get the crash a bit later: The GAP 4.13.1 can load its packages, processes a few testfiles, and crashes in `GAP.jl/pkg/JuliaInterface/tst/help.tst`.

test/macros.jl
```diff
@@ -59,3 +59,15 @@ end
...
+@testset "@install GapObj" begin
+    a = GAP.evalstr("(1,2)")
+
+    struct TestType1 X::GapObj end
+    GAP.@install GapObj(x::TestType1) = x.X
+    @test GapObj(TestType1(a)) === a
+
+    struct TestType2 X::GapObj end
+    GAP.@install function GapObj(x::TestType2) return x.X; end
+    @test GapObj(TestType2(a)) === a
```
fingolfin, line 72: Maybe also add a few tests for the error cases? I.e. trying to use the macro on functions other than `GapObj`, or on methods which have more or fewer arguments than expected

src/macros.jl
```diff
@@ -419,3 +419,47 @@ macro wrap(ex)
...
+macro install(ex)
+    errmsg = "GAP.@install must be applied to a unary method definition for GapObj"
+
+    # split the method definition
+    def_dict = try
+        MacroTools.splitdef(ex)
+    catch
+        error(errmsg)
+    end
+    def_dict[:name] === :GapObj || throw(ArgumentError(errmsg))
```
fingolfin, line 448: What if a user doesn't want to import `GapObj` and does
```julia
GAP.@install GAP.GapObj(x::T) = f(x)
```
Does that work as well? If not, can we make it work with reasonable effort?

src/julia_to_gap.jl
```diff
@@ -46,44 +46,52 @@ The following `julia_to_gap` conversions are supported by GAP.jl.
...
 | `Function`                           | `IsFunction` |
 """
-julia_to_gap(x::FFE) = x    # Default for actual GAP objects is to do nothing
-julia_to_gap(x::Bool) = x   # Default for actual GAP objects is to do nothing
+function julia_to_gap end
+
+# default
+julia_to_gap(x, cache::GapCacheDict = nothing; recursive::Bool = false) = julia_to_gap_internal(x, cache, recursive)
+
+# The calls to `GAP.@install` install methods for `julia_to_gap_internal`
```
fingolfin, line 54:
```suggestion
# The calls to `GAP.@install` install methods for `julia_to_gap_internal`
# so we must make sure it is declared before
```

src/julia_to_gap.jl
```diff
@@ -46,44 +46,52 @@ The following `julia_to_gap` conversions are supported by GAP.jl.
 | `UnitRange{T}`, `StepRange{T, S}`    | `IsRange`    |
 | `Function`                           | `IsFunction` |
 """
-julia_to_gap(x::FFE) = x    # Default for actual GAP objects is to do nothing
-julia_to_gap(x::Bool) = x   # Default for actual GAP objects is to do nothing
+function julia_to_gap end
+
+# default
+julia_to_gap(x, cache::GapCacheDict = nothing; recursive::Bool = false) = julia_to_gap_internal(x, cache, recursive)
```
fingolfin, line 52: Isn't this in fact the only remaining method for `julia_to_gap`? Then perhaps we can just do this:
```suggestion
julia_to_gap(x, cache::GapCacheDict = nothing; recursive::Bool = false) = julia_to_gap_internal(x, cache, recursive)
```

fingolfin: Note that `src/constructors.jl` has this:
```julia
## Handle "conversion" to GAP.Obj and GAP.GapObj (may occur in recursions).
Obj(x::Obj) = x
GapObj(x::GapObj) = x

## Handle conversion of Julia objects to GAP objects
Obj(obj; recursive::Bool = false) = julia_to_gap(obj, IdDict(); recursive)::Obj
GapObj(obj; recursive::Bool = false) = julia_to_gap(obj, IdDict(); recursive)::GapObj

Obj(obj, recursive::Bool) = julia_to_gap(obj, IdDict(); recursive)::Obj
GapObj(obj, recursive::Bool) = julia_to_gap(obj, IdDict(); recursive)::GapObj
```

These are now suboptimal because we create those `IdDict` unconditionally. I assume those could / should all be replaced by `nothing` now?

Alternatively one could rename the last `julia_to_gap` method here to `GapObj`, adjust `src/constructors.jl` accordingly, and either just drop `julia_to_gap` or turn it into an alias resp. mark it as deprecated.

At that point one could also consider renaming `julia_to_gap_internal` to `GapObj_internal`.

src/julia_to_gap.jl
```diff
@@ -133,7 +143,7 @@ function julia_to_gap(
         end
         if recursive
             x = get!(recursion_dict, x) do
```
fingolfin, line 145: The compiler can not infer the type for `recursion_dict` here, but we can
```suggestion
            x = get!(recursion_dict::RecDict, x) do
```

src/julia_to_gap.jl
```diff
@@ -133,7 +143,7 @@ function julia_to_gap(
...
-                julia_to_gap(x, recursion_dict; recursive)
+                julia_to_gap_internal(x, recursion_dict, recursive)
```
fingolfin, line 146:
```suggestion
                julia_to_gap_internal(x, recursion_dict::RecDict, recursive)
```

src/julia_to_gap.jl
fingolfin: Oh, and an existing issue: we should not use `x` both for the original value for the converted value. So it would be good to change `x = ` here to e.g. `res = ...` and then below change `ret_val[i] = x` o `ret_val[i] = res` (or whatever)

src/julia_to_gap.jl
```diff
@@ -142,66 +152,64 @@ function julia_to_gap(
...
     ret_val = NewPlist(rows)
     if recursive
+        if recursion_dict === nothing
+          recursion_dict = RecDict()
+        end
         recursion_dict[obj] = ret_val
     end
     for i = 1:rows
-        ret_val[i] = julia_to_gap(obj[i, :], recursion_dict; recursive)
+        ret_val[i] = julia_to_gap_internal(obj[i, :], recursion_dict, recursive)
```
fingolfin, line 172: Existing issue, but shouldn't this also do the `get!(recursion_dict::RecDict, x)` dance?

src/julia_to_gap.jl
```diff
@@ -215,26 +223,32 @@ end
...
     elseif Wrappers.IsList(obj)
         len = length(obj)
         ret_val = NewPlist(len)
+        if recursion_dict === nothing
+          recursion_dict = RecDict()
+        end
         recursion_dict[obj] = ret_val
         for i = 1:len
-             ret_val[i] = julia_to_gap(obj[i], recursion_dict; recursive)
+             ret_val[i] = julia_to_gap_internal(obj[i], recursion_dict, recursive)
```
fingolfin, line 241: Another case with missing `get!(recursion_dict::RecDict, x)` etc.?

test/macros.jl
ThomasBreuer: When we call the macro in error cases then we run into `LoadError` before `@test_throws` can do something.
The usual solution seems to put the code of the macro into a helper function, and to test this helper function in the error cases.

src/macros.jl
ThomasBreuer: O.k.

src/julia_to_gap.jl
ThomasBreuer: I think no.
The idea of the `get!` call is twofold.
1. Check whether the object in question is already in the dictionary, and if yes then take the value.
2. If the object is not yet in the dictionary then store it together with the converted value.
Situation 1. cannot occur because we are just now creating the Julia object anew.
And situation 2. would not help later on because no pointer to this newly created object can be found in the recursion.
(I will add a comment.)

src/julia_to_gap.jl
ThomasBreuer: Yes, good spot!
I will add a test for this situation (a GAP list containing identical Julia objects which themselves do *not* support/need recursion).

src/julia_to_gap.jl
ThomasBreuer: If we want to get rid of `julia_to_gap`, and recommend `GapObj` instead, also the documentation has to be adjusted.
The only logical problem is that then `GapObj` can return also objects of type `GAP.Obj` that are not of type `GapObj`. (We had talked about this.)
I will leave this renaming to a separate commit, and push the other changes now.

test/macros.jl
fingolfin: One can use `@macroexpand` for that purpose, e.g. in AbstractAlgebra we use this in a test:
```julia
        @test_throws ArgumentError @macroexpand @attr foo() = 1
```

src/julia_to_gap.jl
fingolfin: I feel there is something inconsistent here. Because then the object produced by recurively called `julia_to_gap_internal` is potentially never added to `recursion_dict`, is it? So if we had `T = BigInt ` we would not be caching the conversion results for numerator or denominator. While if we convert a list of `BigInt` they would be cached.

Perhaps we never want to cache a `BigInt`'s converted while. But then why does the `julia_to_gap_internal` method for `Rational{T}` provide the the full 3-argument form instead of just using `GAP.@install`?

src/gap_to_julia.jl
```diff
@@ -15,6 +15,8 @@ An internal type of GAP.jl used for tracking conversion results in `gap_to_julia
 """
 const RecDict = IdDict{Any,Any}
```
fingolfin, line 16: Thinking about this, can't this become
```suggestion
const RecDict = IdDict{Any,GAP.Obj}
```

src/julia_to_gap.jl
fingolfin: OK the crucial point I missed is that we convert row-wise, i.e. we convert `obj[i, :]` and not elementwise, i.e. `obj[i,j]`

src/julia_to_gap.jl
```diff
@@ -93,37 +99,39 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
 ) where {T}
 
+    if recursion_dict !== nothing && haskey(recursion_dict, obj)
+      return recursion_dict[obj]
+    end
     len = length(obj)
     ret_val = NewPlist(len)
     if recursive
+        if recursion_dict === nothing
+          recursion_dict = RecDict()
```
fingolfin, line 133: Another optimization note for future iterations on this: if we are given e.g. a `Vector{Int16}` and asked to recursively convert it to a `GapObj`, then we really don't need the `recursion_dict`, as there is no point storing the `Int16` values in it.

To get rid of such things, my "dual" code in PR #777 does this:
```
    if rec_dict === nothing && _needs_conversion_tracking(T)
        rec_dict = RecDict()
    end
```
where `_needs_conversion_tracking(T)` returns `true` if `T` is a type where the conversion results should be tracked.

Such an approach interacts nicely with my suggestion to "split" `GAP.@install`:
- `GAP.@install`  (or whatever it is called) could also produce a `_needs_conversion_tracking` method returning `false`
- `GAP.@install_rec`  (or whatever it is called) could also produce a `_needs_conversion_tracking` method returning `true`


Note: I guess we should have two different `_needs_conversion_tracking` functions, with different names, for the two direction: to and from Julia. Perhaps they'll end up identical but this is not a priori clear to me? Anyway there would be no harm in having two for the moment.

ThomasBreuer:
The current test failures are due to crashes (Julia 1.11, Julia nightly) and due to a problem with `@macroexpand` (Julia 1.6).

lgoettgens:
For the 1.6 failure: The place in AbstractAlgebra where the macroexpand idea came from limits to some julia versions. See https://github.com/Nemocas/AbstractAlgebra.jl/blob/025eabc50948f7db78bd5561acfb73bd5b340cc8/test/Attributes-test.jl#L228.

ThomasBreuer:
@lgoettgens Thanks.
Yes, I will do the same as AbstractAlgebra does.

src/gap_to_julia.jl
ThomasBreuer: Then `RecDict` cannot be used for the GAP to Julia conversion.

ThomasBreuer:
@fingolfin I think now it remains to decide where we really want recursion tracking.
Currently this is switched off for example for integers, via the various `GAP.@install GapObj` methods. In particular the additional `GapObj_internal(x::Integer, ...)` method does not strike.

src/julia_to_gap.jl
```diff
@@ -93,148 +131,235 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
+    recursive && recursion_dict !== nothing && haskey(recursion_dict, obj) && return recursion_dict[obj]
+
+    rows = size(obj, 1)
     ret_val = NewPlist(rows)
-    if recursive
-        recursion_dict[obj] = ret_val
+
+    rec = recursive && _needs_tracking_julia_to_gap(T)
+    if rec && recursion_dict === nothing
+        recursion_dict = RecDict()
```
fingolfin, line 254:
```suggestion
       recursion_dict = RecDict()
```
(Formatting still seems to mix 2-space and 3-space wide indents -- this line was 6 spaces intended, now it is 5?)

src/julia_to_gap.jl
```diff
@@ -93,148 +131,235 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
 ## Dictionaries
-function julia_to_gap(
+function GapObj_internal(
     obj::Dict{T,S},
-    recursion_dict::IdDict{Any,Any} = IdDict();
-    recursive::Bool = false,
-) where {S} where {T<:Union{Symbol,AbstractString}}
+    recursion_dict::GapCacheDict,
+    ::Val{recursive},
+) where {S} where {T<:Union{Symbol,AbstractString}} where {recursive}
```
fingolfin, line 295:
```suggestion
) where {S, T<:Union{Symbol,AbstractString}, recursive}
```

src/julia_to_gap.jl
```diff
@@ -93,148 +131,235 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
+
+    rec = recursive && _needs_tracking_julia_to_gap(T)
+    if rec && recursion_dict === nothing
+        recursion_dict = RecDict()
     end
+
+    if recursion_dict !== nothing
+      recursion_dict[obj] = ret_val
+    end
+
```
fingolfin, line 260: I wonder if we should do this:
```suggestion

    if !rec
      recursion_dict = nothing
    end

```
With that, I think we could replace the `if rec / else` in the for loop by the content of the `if` clause, i.e. just
```
    for i = 1:rows
        ret_val[i] = GapObj_internal(obj[i, :], recursion_dict::RecDict, Val(recursive))
    end
```

And then perhaps we could put this ugly but repetitive intro block into a macro or helper function?

src/julia_to_gap.jl
```diff
@@ -93,148 +131,235 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
     obj::GapObj,
-    recursion_dict::IdDict{Any,Any} = IdDict();
-    recursive::Bool = false,
-)
+    recursion_dict::GapCacheDict,
+    ::Val{recursive},
+) where {recursive}
     if ! recursive
         ret_val = obj
     elseif Wrappers.IsList(obj)
```
fingolfin, line 339: This method maybe needs
```julia
recursion_dict !== nothing && haskey(recursion_dict, obj) && return recursion_dict[obj]
```

Of course this then is an existing bug.

It might matter if converting e.g. a `Vector{GapObj}` recursively which contains the same GAP object multiple times, and that object needs conversion...

On the other I am not even sure anymore whether this entire method is desirable. There is an argument for stopping conversion once we reach a `GapObj`. (Of course one could allow a third value for `recursive` to distinguish between "stop at GAP objects" and "recurse through everything", but I really wonder how useful it is to recurse through GapObj as we do here. Do we have any real world example where this ever mattered?

ThomasBreuer: First of all: Yes, this extra line is needed if we want recursive conversion of `GapObj`'s to GAP; I will add a test for this situation.

Concerning the question whether we want this recursive conversion (and analogously recursive conversion of Julia objects to Julia), I think it would be not a problem for users to change the current setup, and define that recursive conversion to GAP stops at `GapObj`, and recursive conversion to Julia stops at non-`GapObj`.
Probably just the tests in GAP.jl will notice the difference.

src/julia_to_gap.jl
```diff
@@ -93,148 +131,213 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
+function GapObj_internal(
     obj::AbstractVector{T},
-    recursion_dict::IdDict{Any,Any} = IdDict();
-    recursive::Bool = false,
-) where {T}
+    recursion_dict::GapCacheDict,
+    ::Val{recursive},
+) where {T, recursive}
+
+    ret, rec, rec_dict = recursion_info(T, obj, recursive, recursion_dict)
```
fingolfin, line 178: "ret" and "rec" are very easy to mix up, they look so similar.

And then we also have `rec_dict` and `ret_val`...

I think we could use a macro to avoid introducing some of these. But all of this can be done in a follow-up PR (or not at all).

src/julia_to_gap.jl
```diff
@@ -93,148 +131,213 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
+    ret_val = NewPlist(length(obj))
+
+    recursion_dict = handle_recursion(obj, ret_val, rec, rec_dict)
+
+    for x in obj
+        if recursive
+            res = GapObj_internal(x, recursion_dict, Val(true))
+        else
+            res = x
         end
```
fingolfin, line 226: Here and elsewhere, might also write
```suggestion
        res = recursive ? GapObj_internal(x, recursion_dict, Val(true)) : x
```
but this, too, can wait for a future PR (or we just don't do it)

src/julia_to_gap.jl
```diff
@@ -93,148 +131,213 @@ function julia_to_gap(x::Rational{T}) where {T<:Integer}
...
+    recursion_dict::GapCacheDict,
+    ::Val{recursive},
+) where {T, recursive}
+
+    ret, rec, rec_dict = recursion_info(T, obj, recursive, recursion_dict)
+    ret && return rec_dict[obj]
+
+    ret_val = NewPlist(length(obj))
+
+    recursion_dict = handle_recursion(obj, ret_val, rec, rec_dict)
```
fingolfin, line 219: Just to record how I think this could look like with a macro *in the future*:
```suggestion
    recursion_dict = @handle_recursion(T, obj, recursive, recursion_dict) do
      NewPlist(length(obj))
    end
```
Except I would maybe rename the return value `recursion_dict = ...` to `rec_dict = ...` to ensure type stability (then of course the code below should be using `rec_dict`, too)

src/julia_to_gap.jl
fingolfin: We discussed this at the OSCAR workshop and I think we agreed that we can stop at GapObj, and that's it.

Can also be handled in a follow-up PR, though we should perhaps open an issue to remind us (unless you want to handle it immediately)

fingolfin (approved):

ThomasBreuer:
The required test `matching: [Oscar#master] - ubuntu-latest, julia ~1.10.0-0` does not want to start, close and reopen ...

ThomasBreuer:
@benlorenz thanks for your comment.

I had thought it would be a good idea to have corresponding branches with the same name in GAP.jl and Oscar.jl,
but apparently this only causes problems.
Should I just select "Merge without waiting for requirements to be met"?

And for the next breaking change in GAP.jl, should I better choose a different name for the corresponding branch in Oscar.jl?

fingolfin:
100% agree with @benlorenz

fingolfin:
Weird failure in the [OscarCI test on Julia 1.10](https://github.com/oscar-system/GAP.jl/actions/runs/10995846054/job/30527696059?pr=1029) (in Aqua tests?). What is this dependency "jl_Te1ctwpcZO [9e52b259-76e2-4c08-b46d-33fd3af550ab]" ? Memory corruption?

```
GC: pause 1103.08ms. collected 0.010208MB. incr 
Test Summary:                         | Pass  Broken  Total  Time
canonical maps and garbage collection |    9       1     10  7.9s
Test Summary: |Time
issue 3107    | None  1.8s
-> Testing test/Modules/UngradedModules.jl took: runtime 12.538 seconds + compilation 49.101 seconds + recompilation 0.0 seconds, 4.491 GiB
Starting tests for /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/test/Aqua.jl
[ Info: Compiling JuliaInterface ...

GC: pause 198.71ms. collected 517.464280MB. incr 
ERROR: The following 2 direct dependencies failed to precompile:

jl_Te1ctwpcZO [9e52b259-76e2-4c08-b46d-33fd3af550ab]

Failed to precompile jl_Te1ctwpcZO [9e52b259-76e2-4c08-b46d-33fd3af550ab] to "/home/runner/.julia/compiled/v1.10/jl_Te1ctwpcZO/jl_KlMALp".
#I  The Julia package 'Downloads' cannot be loaded.
ERROR: LoadError: UndefVarError: `@install` not defined
Stacktrace:
 [1] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [2] include(x::String)
   @ Oscar ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:20
 [3] top-level scope
   @ ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/Groups.jl:1
 [4] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [5] include(x::String)
   @ Oscar ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:20
 [6] top-level scope
   @ ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:256
 [7] include
   @ ./Base.jl:495 [inlined]
 [8] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
   @ Base ./loading.jl:2222
 [9] top-level scope
   @ stdin:3
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/types.jl:17
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/types.jl:17
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/Groups.jl:1
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:1
in expression starting at stdin:3
ERROR: LoadError: Failed to precompile Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] to "/home/runner/.julia/compiled/v1.10/Oscar/jl_3nxY3m".
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
...
```

In a previous run the error was a bit different:
```
GC: pause 1146.51ms. collected 0.010208MB. incr 
Test Summary:                         | Pass  Broken  Total  Time
canonical maps and garbage collection |    9       1     10  8.0s
Test Summary: |Time
issue 3107    | None  1.8s
-> Testing test/Modules/UngradedModules.jl took: runtime 14.151 seconds + compilation 49.45 seconds + recompilation 0.0 seconds, 4.502 GiB
Starting tests for /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/test/Aqua.jl
[ Info: Compiling JuliaInterface ...

GC: pause 169.75ms. collected 516.809388MB. incr 
ERROR: The following 2 direct dependencies failed to precompile:

Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13]

Failed to precompile Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] to "/home/runner/.julia/compiled/v1.10/Oscar/jl_LLtmII".
#I  The Julia package 'Downloads' cannot be loaded.
ERROR: LoadError: UndefVarError: `@install` not defined
Stacktrace:
 [1] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [2] include(x::String)
   @ Oscar ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:20
 [3] top-level scope
   @ ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/Groups.jl:1
 [4] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [5] include(x::String)
   @ Oscar ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:20
 [6] top-level scope
   @ ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:256
 [7] include
   @ ./Base.jl:495 [inlined]
 [8] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2222
 [9] top-level scope
   @ stdin:3
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/types.jl:17
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/types.jl:17
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/Groups.jl:1
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:1
in expression starting at stdin:3
jl_W5mlE9sv4h [30b189bf-1215-468d-8b2f-55ea7a2af89b]

Failed to precompile jl_W5mlE9sv4h [30b189bf-1215-468d-8b2f-55ea7a2af89b] to "/home/runner/.julia/compiled/v1.10/jl_W5mlE9sv4h/jl_XxEYcw".
#I  The Julia package 'Downloads' cannot be loaded.
ERROR: LoadError: UndefVarError: `@install` not defined
Stacktrace:
 [1] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [2] include(x::String)
   @ Oscar ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:20
 [3] top-level scope
   @ ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/Groups.jl:1
 [4] include(mod::Module, _path::String)
   @ Base ./Base.jl:495
 [5] include(x::String)
   @ Oscar ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:20
 [6] top-level scope
   @ ~/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:256
 [7] include
   @ ./Base.jl:495 [inlined]
 [8] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::String)
   @ Base ./loading.jl:2222
 [9] top-level scope
   @ stdin:3
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/types.jl:17
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/types.jl:17
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Groups/Groups.jl:1
in expression starting at /home/runner/work/GAP.jl/GAP.jl/oscar-dev/Oscar/src/Oscar.jl:1
in expression starting at stdin:3
ERROR: LoadError: Failed to precompile Oscar [f1435218-dba5-11e9-1e4d-f1a5fab5fc13] to "/home/runner/.julia/compiled/v1.10/Oscar/jl_7vg26X".

```

ThomasBreuer:
> which might cause this test-environment to use a different GAP version...

This would explain the error message

> ERROR: LoadError: UndefVarError: `@install` not defined

The macro `GAP.@install` gets introduced by this pull request, it is not known if any older GAP.jl is used.

fingolfin:
Then let's move forward with this and make a release today

ThomasBreuer:
Thanks for finally merging this pull request.

PR #1031, opened 2024-09-04
Update JLLs for Julia nightly compat & GC fix

fingolfin:
The fix in the JLL is to disable task stack scanning. This unfortunately has a severe performance penalty, but at least we can now test if it really resolves the GC crashes in all cases while I work on a more permanent solution.


ThomasBreuer (commented):
I have checked the effect of the change on my notebook (Ubuntu 20.04.1 LTS):
With this pull request, the GAP.jl test suite runs in Julia 1.10.4 without errors.
Without this pull request, Julia 1.10.4 crashes as described in a comment in #1029.

fingolfin:
Excellent. Unfortunately CI failed widely -- but it seems due to network issues (phew). I've restarted it now. If it passes I suggest we merge this for now. I'll work on a better fix, but this also contains fixes for Julia nightly so is useful anyway.

So if you agree, just approve this PR?

ThomasBreuer (approved):
Agreed.

lgoettgens:
The required CI jobs still wait for the 1.9 job, but this does not exist anymore due to https://github.com/oscar-system/GAP.jl/pull/1022. The list of required passing jobs needs to be adapted to reflect the changes there.

fingolfin:
Unfortunately there are still failing jobs, but it was merged anyway because I only remove the 1.9 job from the branch protection rules, but had not yet add the new checks I wanted to add (doing that now).

Anyway: e.g. https://github.com/oscar-system/GAP.jl/actions/runs/10696442564/job/29654943194?pr=1031 fails.

So it seems there is still more to uncover -- unfortunately for now I have run out of test cases that reproduce the crash locally for me. I'll continue my search :-(

PR #1033, opened 2024-09-04
Try to make `etc/ci_test.sh` work

lgoettgens:
`etc/ci_test.sh` gets called as part of the testsuite of GAP.jl.

Already for a long time, `gcov` prints some errors in CI, see e.g. https://github.com/oscar-system/GAP.jl/actions/runs/10700854945/job/29665504736#step:6:468. On newer gcov version (my system uses gcc/gcov 14.2.1, while the ubuntu CI runners use 11.2.0), instead of just printing the error, this returns a non-zero exit code and thus aborts the script.
This PR contains three small changes in the hopes to fix this:
1. The gcov contained a wrong path. This is changed for JuliaInterface. Since JuliaExperimental doesn't have any c files anymore, gcov cannot give any coverage for them. So this line gets disabled.
1. The build flags of JuliaInterface now contain the needed flag to create coverage files during compilation. This would of course be nice to have only in CI cases and not always, but I don't know how to do that.
1. I added a force re-compilation of JuliaInterface before running its tests. On CI, this shouldn't change anything, but locally, I had an issue that changes to the Makefile (see previous point) didn't get picked up automatically. This forces it instead.

Alternatively, we could just remove the two lines regarding `gcov` from `etc/ci_test.sh` completely.

cc @ThomasBreuer

### DIFF
--- a/etc/ci_test.sh
+++ b/etc/ci_test.sh
@@ -10,7 +10,7 @@ mkdir -p coverage
 cd pkg/JuliaInterface
 pwd
 # Force recompilation of JuliaInterface with coverage instrumentation
-FORCE_JULIAINTERFACE_COMPILATION=true JULIAINTERFACE_WITH_COVERAGE=true ${GAP} --nointeract
+CFLAGS="--coverage" LDFLAGS="--coverage" FORCE_JULIAINTERFACE_COMPILATION=true ${GAP} --nointeract
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaInterface.coverage tst/testall.g || AnyFailures=Yes
 gcov -o gen/src/ src/*.c*
--- a/etc/ci_test.sh
+++ b/etc/ci_test.sh
@@ -9,7 +9,8 @@ mkdir -p coverage
 #
 cd pkg/JuliaInterface
 pwd
-FORCE_JULIAINTERFACE_COMPILATION=true ${GAP} --nointeract # re-compile JuliaInterface
+# Force recompilation of JuliaInterface with coverage instrumentation
+CFLAGS=--coverage FORCE_JULIAINTERFACE_COMPILATION=true ${GAP} --nointeract
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaInterface.coverage tst/testall.g || AnyFailures=Yes
 gcov -o gen/src/ src/*.c*
@@ -20,7 +21,6 @@ cd pkg/JuliaExperimental
 pwd
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaExperimental.coverage tst/testall.g || AnyFailures=Yes
-#gcov -o gen/src/ src/*.c* # there is no src/ folder anymore
 cd ../..
 
 if [ ${AnyFailures} = Yes ]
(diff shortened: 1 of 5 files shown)


etc/ci_test.sh
```diff
@@ -9,17 +9,18 @@ mkdir -p coverage
...
+gcov -o gen/src/ src/*.c*
 cd ../..
 
 #
 cd pkg/JuliaExperimental
 pwd
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaExperimental.coverage tst/testall.g || AnyFailures=Yes
-gcov -o gen/src/.libs src/*.c*
+#gcov -o gen/src/ src/*.c* # there is no src/ folder anymore
```
fingolfin, line 23: No need to keep this for nostalgia :-)
```suggestion
```

fingolfin (commented):
Thank you!!

etc/ci_test.sh
```diff
@@ -9,17 +9,18 @@ mkdir -p coverage
 #
 cd pkg/JuliaInterface
 pwd
+# Force recompilation of JuliaInterface with coverage instrumentation
+CFLAGS=--coverage FORCE_JULIAINTERFACE_COMPILATION=true ${GAP} --nointeract
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaInterface.coverage tst/testall.g || AnyFailures=Yes
-gcov -o gen/src/.libs src/*.c*
+gcov -o gen/src/ src/*.c*
 cd ../..
```
lgoettgens, line 17:
```suggestion
# Force recompilation of JuliaInterface without coverage instrumentation
FORCE_JULIAINTERFACE_COMPILATION=true ${GAP} --nointeract
cd ../..
```

Maybe we should add something like this? Otherwise, someone running `]test` locally will end up with a JuliaInterface with coverage instrumentation. If this is not desireable, this patch should re-compile a "usual" JuliaInterface

lgoettgens:
Thanks for your changes @fingolfin. I verified that this still works locally on my system.

src/setup.jl
```diff
@@ -212,13 +212,17 @@ function build_JuliaInterface(sysinfo::Dict{String, String})
...
     # places; in any case, if anybody ever cares about this, we can work on
     # finding a better solution.
-    JULIA_CPPFLAGS = filter(c -> c != '\'', cflags())
+    JULIA_CFLAGS = filter(c -> c != '\'', cflags())
     JULIA_LDFLAGS = filter(c -> c != '\'', ldflags())
     JULIA_LIBS = filter(c -> c != '\'', ldlibs())
+    if get(ENV, "JULIAINTERFACE_WITH_COVERAGE", "false") == "true"
+        JULIA_CFLAGS *= " --coverage"
+        JULIA_LDFLAGS *= " --coverage"
+    end
```
fingolfin, line 221: Instead of this limited special case, I'd prefer if we changed this code to not overwrite existing CFLAGS/LDFLAGS in the `withenv` below but rather to append to them.

etc/ci_test.sh
fingolfin: Good point. But for that I think doing `make clean` would suffice?

src/setup.jl
lgoettgens: That was a 50/50 coin flip which of the two versions you prefered. This is in a stash and I can push it tomorrow

etc/ci_test.sh
lgoettgens: yeah it does. changed now

fingolfin (approved):

fingolfin:
Thank you @lgoettgens !

PR #1037, opened 2024-09-11
Disable AvailabilityTest checking for skipping install of packages

lgoettgens:
In some edge-cases a package is available, but one of its transitive dependencies is not. This piece of code would have skipped any installation in this case.
We (@ThomasBreuer and I) ran into this in https://github.com/oscar-system/Oscar.jl/pull/3688, because `recog` is available, but its transitive dependency `io` needs compilation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "GAP"
 uuid = "c863536a-3901-11e9-33e7-d5cd0df7b904"
 authors = ["Thomas Breuer <sam@math.rwth-aachen.de>", "Sebastian Gutsche <gutsche@mathematik.uni-siegen.de>", "Max Horn <horn@mathematik.uni-kl.de>"]
-version = "0.11.2"
+version = "0.11.3"
 
 [deps]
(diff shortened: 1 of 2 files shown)


lgoettgens:
We verified locally, that this PR gets rid of the `ERROR: LoadError: InitError: cannot load the GAP package recog` in https://github.com/oscar-system/Oscar.jl/pull/3688.

lgoettgens:
When I talked to @ThomasBreuer yesterday, our primary goal was to get https://github.com/oscar-system/Oscar.jl/pull/3688 merged, for the simple reason that after that we again have OscarCI that reports potential problems early.

ThomasBreuer:
The point is that just calling the `AvailabilityTest` function is not good enough.
Calling `TestPackageAvailability` would be better, but I have to check this.

lgoettgens:
> If this fixes more problems right now than it causes, go ahead, but we need to eventually solve the "no network connection" issues

(I don't have the right to proceed with merging and tagging a release of this.)

lgoettgens:
superceded by https://github.com/oscar-system/GAP.jl/pull/1039

PR #1039, opened 2024-09-12
fix the availability check in `Packages.install`

ThomasBreuer:
This is an alternative to #1037.

Here we try to keep the check whether a not yet loaded GAP package is already installed, i.e., `Packages.load` would return `true`.
The previous code was wrong in the sense that it did not consider the availability of the needed packages. Now we call the right GAP function, which tests also the needed packages.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -274,8 +274,8 @@ function install(spec::String, version::String = "";
           # then nothing is to do.
           for inforec in getproperty(info, spec)
             if version == string(getproperty(inforec, :Version))
-              fun = getproperty(inforec, :AvailabilityTest)
-              fun() && return true
+              res = Globals.TestPackageAvailability(GapObj(spec), GapObj(version))
+              res !== Globals.fail && return true
             end
           end
         end
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -272,12 +272,8 @@ function install(spec::String, version::String = "";
           # The package is not yet loaded but may be available.
           # If an available version has the required version number
           # then nothing is to do.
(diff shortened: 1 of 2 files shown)


src/packages.jl
```diff
@@ -274,8 +274,8 @@ function install(spec::String, version::String = "";
           # then nothing is to do.
           for inforec in getproperty(info, spec)
             if version == string(getproperty(inforec, :Version))
```
lgoettgens, line 276: these two lines are no longer needed due to the changes below

lgoettgens (approved):
This seems to solve the problem with `recog` in https://github.com/oscar-system/Oscar.jl/pull/3688 as well and seems to be a better solution than https://github.com/oscar-system/GAP.jl/pull/1037

fingolfin:
Thank you for the fix!

A very minor point, but: next time I would recommend to "squash merge" PRs with several "fixup" commits. To avoid a bunch of extra commits on master:
```
* 562986ee (2024-09-13 13:33:12 +0200) remove now unnecessary code lines <ThomasBreuer>
* 2d7d2b02 (2024-09-12 23:23:19 +0200) fix the availability check in `Packages.install` <ThomasBreuer>
* 0938c8ed (2024-09-12 21:47:25 +0200) fix the test ... <ThomasBreuer>
* 85ff8d11 (2024-09-12 16:37:38 +0200) add a test <ThomasBreuer>
* f78459b9 (2024-09-12 15:43:22 +0200) change the `show` method for GAP objects <ThomasBreuer>
```

PR #1041, opened 2024-09-18
Remove JuliaModule

fingolfin:
We didn't use it and JuliaModule("bla") can simply be replaced by Julia.bla


fingolfin:
1. This should also be mentioned in the changelog, but I'll do that just before the release as usual

PR #1049, opened 2024-09-23
Stop wrapping Julia modules in special objects

### LINKED ISSUE #1044: Replacing a Julia module is not handled well
I just run into this while testing some code:
```
gap> JuliaEvalString("module foo  x = 1 end");
<Julia: Main.foo>
gap> Julia.foo.x;
1
gap> JuliaEvalString("module foo  x = 2 end");
WARNING: replacing module foo.
<Julia: Main.foo>
gap> Julia.foo.x;
1
```

So what happens is that `Julia.foo` is cached, and so is not updated when the underlying module changes. We should deal with that better.

fingolfin:
This resolves issues with outdated caches in our module wrappers.

We added these wrappers to serve as a cache. But we already exempted Julia functions from this. For which we create wrapper objects -- so arguably the cache would make most sense for them... But this can cause inconsistencies, hence @ThomasBreuer remove this in PR #423 four years ago.

So the only thing we cached were the wrapper objects for Julia modules. And the only reasons we needed those wrappers were ...
1. to be able to do caching (haha),
2. to be able to supply the "fake" `Julia.GAP` as a way to access the `GAP` Julia module,
3. for some printing tweaks.

This PR achieves all of the above by adjusting the GAP type for Julia object slightly, we now indicate whether we are seeing a Julia module by setting a suitable filter. As an added bonus, this now also works with modules access by any other means, e.g. returned by a `JuliaEvalString` or a Julia function.

The part I like least is the new hack to make `Julia.GAP` possible. A long-term alternative to this might be to provide a different clean way to access this -- e.g. perhaps we should provide a global GAP variable `GAP_jl` which contains the GAP module? (We could also just call it `GAP` but I feel this might be reeeaaally confusing to GAP users, so I'd rather not). For symmetry in Oscar we could also add `Oscar_jl` as an alias for the `Oscar` global. 

But all of that is beyond this PR. We could file it as a separate issue if care enough.

Resolves #1044
Closes #1045

### DIFF
--- a/pkg/JuliaInterface/gap/JuliaInterface.gd
+++ b/pkg/JuliaInterface/gap/JuliaInterface.gd
@@ -109,8 +109,7 @@
 #! However not every object living on the Julia side is in this filter.
 #! For example Julia booleans and small <C>Int</C> values are directly
 #! translated to GAP booleans and small integers, while for Julia functions
-#! and wrappers dedicated wrappers are used for improved efficiency
-#! respectively additional features.
+#! dedicated wrappers are used to improve efficiency and add features.
 #! @BeginExampleSession
 #! gap> JuliaEvalString( "x = 4" );;
 #! gap> Julia.x;
@@ -118,8 +117,6 @@
 #! gap> IsJuliaObject( Julia.x );
 #! false
 #! gap> IsJuliaObject( Julia.sqrt );
-#! false
-#! gap> IsJuliaObject( Julia.Main );
 #! false
 #! @EndExampleSession
 DeclareCategory( "IsJuliaObject", IsObject );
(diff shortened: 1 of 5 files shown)


fingolfin:
(This should only be merged after #1029, as getting that one right is much harder)

lgoettgens (commented):
I don't feel qualified to comment on the changed internals here. Anyway, could you add a test for #1044, e.g. the one from https://github.com/oscar-system/GAP.jl/pull/1045/files#diff-f673fbc29b03e12764445c880278a94061eb0b24269b0fd2621c27c604d50b1e?

fingolfin:
@lgoettgens done, I shamelessly stole it from your PR :-)

lgoettgens:
> @lgoettgens done, I shamelessly stole it from your PR :-)

I stole it from your issue, so I think we're even now :)

PR #1052, opened 2024-09-24
Update CHANGES.md, set version to 0.12.0-DEV, augment a benchmark file

fingolfin:
Plan to release 0.12.0 later today, this prepares the path.


fingolfin:
I think that error will in fact best be fixed in Oscar.jl. I am working on it right now.

PR #1054, opened 2024-09-24
Hack to avoid warning about Downloads package

### LINKED ISSUE #1040: Warning "The Julia package 'Downloads' cannot be loaded" when starting GAP.jl in CI
This is printed by `JuliaImportPackage`.

That in turn is called by GAP's `PackageManage` which has a `Download.g` file containing this:
```
if IsBound(Julia) and JuliaImportPackage("Downloads") = true then
```

fingolfin:
The "proper" fix will come eventually with the new packagemanager. But in the meantime this should fix #1040 for users.


ThomasBreuer (approved):
Yes.
The warning disappears, nothing else changes.

PR #1059, opened 2024-10-24
Improve evalstr to show syntax warnings

### LINKED ISSUE #1058: irritating GAP warning/error handling
I stumbled over the following when dealing with #1057.

```
julia> f = GAP.evalstr("function() res:= 1; return res; end")
GAP: function(  ) ... end

julia> f()
1

julia> g"\\"
ERROR: Error thrown by GAP: Syntax warning: Unbound global variable in stream:1
function() res:= 1; return res; end;
           ^^^
Syntax warning: Unbound global variable in stream:1
function() res:= 1; return res; end;
                           ^^^
Syntax error: String must end with " before end of file in stream:1
[...]
```

The GAP syntax warning is not shown in the construction of the function but it is kept in some buffer.
For the later GAP error, the error message gets printed, but first the text of the old syntax warning appears.

fingolfin:
Resolves #1058

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -66,13 +66,19 @@ const last_error = Ref{String}("")
 
 const disable_error_handler = Ref{Bool}(false)
 
-function error_handler()
+function copy_gap_error_to_julia()
     global disable_error_handler
     if disable_error_handler[]
         return
     end
     last_error[] = String(Globals._JULIAINTERFACE_ERROR_BUFFER::GapObj)
     ccall((:SET_LEN_STRING, libgap), Cvoid, (GapObj, Cuint), Globals._JULIAINTERFACE_ERROR_BUFFER::GapObj, 0)
+end
+
+function get_and_clear_last_error()
+    err = last_error[]
+    last_error[] = ""
+    return err
 end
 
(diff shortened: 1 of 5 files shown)


ThomasBreuer (approved):
This is a good idea: Turn a real GAP error into a Julia error, and handle the GAP warning by just printing it.

As far as I understand, we have the following.
The syntax errors happen in the `evalstr_ex` call, and the buffer gets cleaned on the outer level, in `evalstr`.
Thus we will still keep the warning in the buffer if we call `evalstr_ex` directly. (Note that `eval_str_ex` is documented, and is mentioned in the docstring of `evalstr`.)

Logically, it would be better if `evalstr_ex` (or some helper function) would collect all GAP errors and warnings, clean the error buffer, and return also the collected error an warning information. Then the status of GAP would be clean after the `evalstr_ex` call. And `evalstr` would still be able to throw the Julia exception and to print the warning.

Is this reasonable?

fingolfin:
I like the idea of adding the last error as a second argument to `evalstr_ex`. However it would be breaking change. @ThomasBreuer discussed this and agreed to merge this as-is for now, and we'll change `evalstr_ex` in a later PR for the next release.

PR #1060, opened 2024-10-24
setup_override_dir.jl: allow passing args to configure

fingolfin:
Also use that to treat warnings as error in some CI tests.

### DIFF
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -63,11 +63,11 @@ jobs:
           mv GAPROOT /tmp/GAPROOT
           cd /tmp/GAPROOT
           ./autogen.sh
-          ./configure
+          ./configure --enable-Werror
           make -j`nproc`
       - name: "Override bundled GAP"
         run: |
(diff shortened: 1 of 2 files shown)


lgoettgens (approved):
seems reasonable

lgoettgens:
The failures:
- julia nightly, GAP master: will be fixed by https://github.com/JuliaLang/julia/pull/56319
- julia nightly, GAP stable-4.13: will be fixed by https://github.com/JuliaLang/julia/pull/56319
- julia 1.6, GAP stable-4.13: 
```
/tmp/GAPROOT/src/julia_gc.c:748:12: error: ‘gap_jl_boundp’ defined but not used [-Werror=unused-function]
  748 | static int gap_jl_boundp(jl_module_t *m, jl_sym_t *var)
      |            ^~~~~~~~~~~~~
   C       /tmp/GAPROOT/src/compstat_empty.c => build/obj/src/compstat_empty.c.o
   C       /tmp/GAPROOT/src/main.c => build/obj/src/main.c.o
cc1: all warnings being treated as errors
```

PR #1062, opened 2024-10-29
Run on RPTU servers

aaruni96:
I am not sure if this is actually correct, more explanation on why, if this doesn't work.

### DIFF
--- a/.github/workflows/oscar.yml
+++ b/.github/workflows/oscar.yml
@@ -57,7 +57,7 @@ jobs:
 
   test-oscar:
     needs: generatematrix
-    name: ${{ join(matrix.*.name) }} - ${{ runner.os }}, julia ${{ matrix.julia-version}}
+    name: ${{ join(matrix.*.name) }} - ${{ join(matrix.os) }}, julia ${{ matrix.julia-version}}
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.julia-version == 'nightly' }}
     env:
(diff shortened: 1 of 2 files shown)


aaruni96:
This currently spawns two values for runs-on: Linux, and RPTU. What we really want is one runs-on, with two labels: Linux and RPTU (such that github then forwards this job to only runners tagged with both of these labels). Only forwarding it to RPTU tagged runners means it might get run on one of the macos runners, and only forwarding it to Linux tagged runners leaves it open to running on other runners (in case github ever labels their runners as linux in addition to ubuntu).

cc @benlorenz @fingolfin any ideas ?

fingolfin:
@aaruni96 we should revive this for the OscarCI tests. If it works, then we should also use the same for Singular.jl and Polymake.jl

aaruni96:
This one job did not actually run, but does not show up as a failure : https://github.com/oscar-system/GAP.jl/actions/runs/16171366065/job/45645535770?pr=1062

https://github.com/oscar-system/GAP.jl/actions/runs/16171366065/job/45645535770?pr=1062#step:8:229

Is this expected / okay ?

lgoettgens:
> This one job did not actually run, but does not show up as a failure : https://github.com/oscar-system/GAP.jl/actions/runs/16171366065/job/45645535770?pr=1062
> 
> https://github.com/oscar-system/GAP.jl/actions/runs/16171366065/job/45645535770?pr=1062#step:8:229
> 
> Is this expected / okay ?

This is fine. (explanation: there has been a breaking GAP.jl release that the latest Oscar is not compatible with)

fingolfin:
Using `runner.os` did not work, so I went with the `join(matrix.os)` which seems reasonable enough.

Issue #1065, opened 2024-11-15
Systematic testing of all packages in the GAP distro


fingolfin:
We should test all packages in the GAP package distribution within GAP.jl. That means running `TestPackage("PKGNAME");`.

I am sure this will reveal a bunch of issues. For starters, we may need to add a `GAP.Packages.build()` command so we can use the bundled version of those packages without installing newer ones (without any network access in fact).

Some of this my be rendered moot by switching to `GAP_pkg_*` wrapper packages (but I view this issue as orthogonal to that. I.e. even if we had `GAP_pkg_*` wrappers in place right now, I'd still would want to run their *tests* -- it just would be a tad easier to do so because I would not have to worry about *building* them).

On the long run, these tests could be automated by CI. E.g. perhaps in the https://github.com/oscar-system/GAP_pkg repository there could be a CI job for each package (with a few disabled). 

Of course a few things never will work in their current form (e.g. I have no hopes for `xgap` and hence `itc`). That's fine. But yesterday I tried to install and run `nq` and `anupq` and embarrassingly ended up running headlong into issue #1064.

PR #1066, opened 2024-11-21
Add `GAP.Packages.build(name)`

lgoettgens:
Some progress towards https://github.com/oscar-system/GAP.jl/issues/1065.

As pointed out in https://github.com/oscar-system/GAP.jl/issues/1065#issuecomment-2491560696, this function may need to copy a package from an artifact directory to the user pkgdir to have it in a location that is both writable and `PackageManager:CompilePackage` does not complain about.


fingolfin:
Thanks! I overlooked this PR, sorry :-(

PR #1067, opened 2024-11-21
Add GAP distro tests

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/1065.

~~This is dependent on https://github.com/oscar-system/GAP.jl/pull/1066 and thus includes it.~~

I've used the ugly hack from https://github.com/oscar-system/GAP_pkg/pull/15 in `GAP.Packages.test` to avoid GAP packages exiting the process after running their tests. This function could be moved to a package extension on `Test` once the minimum required julia version is bumped to 1.8.


.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,75 @@
...
+          version: '1'
+      - name: "Cache artifacts"
+        uses: julia-actions/cache@v2
+      - name: "Build package"
+        uses: julia-actions/julia-buildpkg@v1
+      - name: Get list of GAP packages
+        id: set-matrix
+        run: julia --project=. -e '
+                  using Artifacts;
+                  output = sprint(print, "gap-packages=", filter(name -> startswith("d", name) || startswith("f", name), readdir(artifact"gap_packages")));
```
lgoettgens, line 36:
```suggestion
                  output = sprint(print, "gap-packages=", readdir(artifact"gap_packages"));
```
once we are happy with how this works, we can enable it to test the whole distro and not just ~10 packages

lgoettgens:
The remaining failures (for packages starting with `d` or `f` (random two letters)) seem to be more or less genuine:
- `design`: its (transitive) dep `GRAPE` has a build error
- `difsets`: its (transitive) dep `GRAPE` has a build error
- `fining`: its (transitive) dep `GRAPE` has a build error
- `francy`: its (transitive) dep `zeromqinterface` has a build error

So I think that this is ready to review now.

fingolfin:
Thank you for this @lgoettgens. Could you please rebase it? I'll try to have a closer in the coming week(s)

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,75 @@
...
+          version: '1'
+      - name: "Cache artifacts"
+        uses: julia-actions/cache@v2
+      - name: "Build package"
+        uses: julia-actions/julia-buildpkg@v1
+      - name: Get list of GAP packages
+        id: set-matrix
+        run: julia --project=. -e '
+                  using Artifacts;
+                  output = sprint(print, "gap-packages=", filter(startswith(r"d|f"), readdir(artifact"gap_packages")));
```
fingolfin, line 36: This doesn't work anymore now :-(

fingolfin: One could read Artifacts.toml, take all keys of the form `GAP_pkg_foo`, and extract the `foo`, I guess..

lgoettgens: I fixed it using #1090 (which is included in this PR here for the time being)

lgoettgens:
Hmm, I think the newly introduced `GAP.Packages.build` (ref #1066) and `GAP.Packages.build_recursive` currently don't work great with jll-backed packages. E.g. the test of `difsets` fails due to a build error in `GRAPE`, which should not be relevant anymore since #1086. I am not really sure what the `build` commands should do with these jll-backed package versions.

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,75 @@
...
+          version: ${{ matrix.julia-version }}
+          arch: ${{ matrix.julia-arch }}
+      - name: "Cache artifacts"
+        uses: julia-actions/cache@v2
+        with:
+          cache-scratchspaces: false
+      - name: "Build package"
+        uses: julia-actions/julia-buildpkg@v1
+      - name: "Build GAP package"
+        run: julia --color=yes --project=. -e 'using GAP; GAP.Packages.build_recursive("${{ matrix.gap-package }}") || exit(1)'
```
fingolfin, line 73: With the JLL work, do we even need to / want to build those packages?

On the one hand, yeah, it's useful to test whether building a package works

On the other hand, no, we really want to test that the *prebuilt* package JLLs work out of the box...

So I guess ideally we'd want to test both -- but for now my personal priority would be on the second version: making sure that the JLL binaries actually works is more important right now because the idea is that this is what most users will get.

So I suggest we focus on that for now -- later we could adjust this workflow to also test package building (or perhaps make another workflow for that)

lgoettgens: but you're right. now that almost all packages have a jll, we could skip the building altogether and instead fail the CI job if a package is not available immediately (aka the jll is not available of faulty). Is this what you had in mind?

fingolfin:
We also need an exclusion list: we won't be able to make use of e.g. xgap or itc at this point

lgoettgens:
Let me split this PR up into some smaller parts

src/packages.jl
```diff
@@ -408,6 +410,111 @@ function build(name::String; quiet::Bool = false,
...
+            end
+          end
+        end
+      end
+    end
+  finally
+    disable_error_handler[] = false
+  end
+
+  @test !error_occurred
```
fingolfin, line 515: Could we just do this and not use `@test` here and rely on the caller to do `@test GAP.Packages.test("bla")` ?
```suggestion
  return !error_occurred && (ended_using_QuitGap || result == true)
```

src/packages.jl
```diff
@@ -408,6 +410,111 @@ function build(name::String; quiet::Bool = false,
...
+
+  disable_error_handler[] = true
+  try
+    with_gap_var("ERROR_OUTPUT", Globals._JULIAINTERFACE_ORIGINAL_ERROR_OUTPUT) do
+      with_gap_var("QuitGap", fake_QuitGap) do
+        with_gap_var("QUIT_GAP", fake_QuitGap) do
+          with_gap_var("ForceQuitGap", identity) do
+            with_gap_var("FORCE_QUIT_GAP", identity) do
+              result = Globals.TestPackage(GapObj(name))
+              @test ended_using_QuitGap || result == true
```
fingolfin, line 503:
```suggestion
              return !error_occurred && (ended_using_QuitGap || result == true)
```
???

lgoettgens:
This now includes everything that we discussed today:
- adapted triggers (https://github.com/oscar-system/GAP.jl/pull/1067#discussion_r1852723867 needs to be applied once we are happy otherwise)
- everything about building packages has been moved to https://github.com/oscar-system/GAP.jl/pull/1097
- all failing packages have been added to the exclude list (with a short reason)

But now, a lot of *other* packages fail. I guess I could iterate adding everything failing to the exclusion list, but that does not seem future-proof...

fingolfin:
I'll look into the failing tests

fingolfin:
All test failures I looked at seem to fail because an external executable is not found / can not be launched (and skimming over list of names with failures, most indeed are on the list of such packages, e.g. ace, anupq, cohomolo, ...). On the other hand, `nq` passes... Huh.

Testing with your branch locally also does not reproduce the issues. E.g. I tried `julia -e 'using GAP, Test; @test GAP.Packages.test("ace")'` (within a suitable project environment) and it passed just fine.

So something is fishy.

I'll add some debugging code, and I'll also re-enable a few of the things you excluded to see why they are failing.

fingolfin:
Many of the failures are due to missing `Singular`. I'll add an override for that, too.

fingolfin:
Hrm...
```
┌ Debug: DirectoriesPackagePrograms override detected:
│     /home/runner/.julia/scratchspaces/c863536a-3901-11e9-33e7-d5cd0df7b904/gap_12610715269907574674-1.11/pkg/GAP_pkg_cohomolo/cohomolo-1.6.11/
│  => /home/runner/.julia/scratchspaces/386407dc-74c0-5367-b2ac-4ae54610b78b/binarywrappers_v1.11/bin/
└ @ GAP ~/work/GAP.jl/GAP.jl/src/GAP_pkg.jl:70

#I  Package ``cohomolo'': The program `extprun' (for example) is not compiled
#I  `cohomolo' is thus unavailable
```

Yet at least on my system I get:
```
$ ls -l ~/.julia/scratchspaces/386407dc-74c0-5367-b2ac-4ae54610b78b/binarywrappers_v1.11/bin/extprun
-rwxr-xr-x 1 mhorn staff 439 Dec 18 22:36 /Users/mhorn/.julia/scratchspaces/386407dc-74c0-5367-b2ac-4ae54610b78b/binarywrappers_v1.11/bin/extprun
```


Running this on GitHub gives
```
ls: cannot access '/home/runner/.julia/scratchspaces/*/binarywrappers_v1.11/bin': No such file or directory
```
so it looks as if the scratchspace is not created by BinaryWrappers.jl? Or at least not in the expected place?!?

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,117 @@
...
+      - uses: actions/checkout@v4
+      - name: "Set up Julia"
+        uses: julia-actions/setup-julia@v2
+        with:
+          version: ${{ matrix.julia-version }}
+          arch: ${{ matrix.julia-arch }}
+      - name: "Cache artifacts"
+        uses: julia-actions/cache@v2
+        with:
+          cache-scratchspaces: false
```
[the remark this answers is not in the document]
fingolfin: Thanks @benlorenz that explains it. I'll change the code to create the wrappers in `__init__` as this may otherwise also impact users.

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,115 @@
...
+      - uses: actions/checkout@v4
+      - name: "Set up Julia"
+        uses: julia-actions/setup-julia@v2
+        with:
+          version: ${{ matrix.julia-version }}
+          arch: ${{ matrix.julia-arch }}
+      - name: "Cache artifacts"
+        uses: julia-actions/cache@v2
+        with:
+          include-matrix: false # don't create a cache per GAP package
```
lgoettgens, line 108:
```suggestion
          include-matrix: false # don't create a cache per GAP package
          cache-scratchspaces: false
```
Let's try to re-enable this setting once https://github.com/oscar-system/GAP.jl/pull/1100 is merged.

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,116 @@
...
+          - gap-package: 'polenta'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'polycyclic'                 # ???
+          - gap-package: 'polymaking'                 # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'profiling'                  # segfaults during testing
+          - gap-package: 'radiroot'                   # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'ringsforhomalg'             # `Error, found no Singular executable in PATH while searching the following list`'
+          - gap-package: 'semigroups'                 # no jll
+          - gap-package: 'singular'                   # `Error, Singular executable file not found!`
+          - gap-package: 'toricvarieties'             # `Error, found no Singular executable in PATH while searching the following list`'
+          #- gap-package: 'typeset'                    # `ReadPackage could not read <digraphs>/gap/doc.g`
```
fingolfin, line 94: The warning is a red herring, the real problem is that it wants to run `dot2tex`. I guess we could make that work via some suitable `apt-get install` invocation?
```suggestion
          #- gap-package: 'typeset'                    # needs `dot2tex` executable
```

lgoettgens:
What do you think about merging this in the near future (e.g. after https://github.com/oscar-system/GAP.jl/pull/1108 in merged) and dealing with the remaining failures/excludes afterwards bit by bit?

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          - ubuntu-latest
+        gap-package: ${{fromJSON(needs.generate-matrix.outputs.gap-packages)}}
+        exclude:
+          - gap-package: 'alnuth'                     # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          #- gap-package: 'autodoc'                    # test failure
+          #- gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
+          - gap-package: 'curlinterface'              # no jll
+          - gap-package: 'example'                    # no jll
+          #- gap-package: 'examplesforhomalg'          # `Error, found no GAP executable in PATH`
+          #- gap-package: 'gapdoc'                     # test failure
```
fingolfin, line 70: gapdoc failure may be due to missing bibtex / latex

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          - gap-package: 'itc'                        # dependency `xgap` has no jll
+          #- gap-package: 'localizeringforhomalg'      # `Error, found no GAP executable in PATH`
+          #- gap-package: 'normalizinterface'          # `Error, Some error in the normaliz input data detected: Cone Property in last line not allowed for lattice ideals`
+          #- gap-package: 'packagemanager'             # test failure
+          - gap-package: 'polenta'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'polycyclic'                 # ???
+          - gap-package: 'polymaking'                 # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          #- gap-package: 'profiling'                  # segfaults during testing
+          - gap-package: 'radiroot'                   # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          #- gap-package: 'ringsforhomalg'             # `Error, found no GAP executable in PATH`
```
fingolfin, line 84: no GAP executable is something I could fix, but this can be postponed: the homalg team of @mohamed-barakat already has their own Julia versions of these packages, so getting the pure GAP version to work is not an immediate priority (so the perfectionist in me wants it anyway, but it can wait for a follow-up PR).

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          #- gap-package: 'examplesforhomalg'          # `Error, found no GAP executable in PATH`
+          #- gap-package: 'gapdoc'                     # test failure
+          - gap-package: 'guarana'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'hap'                        # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'hapcryst'                   # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          #- gap-package: 'io'                         # test failure
+          - gap-package: 'itc'                        # dependency `xgap` has no jll
+          #- gap-package: 'localizeringforhomalg'      # `Error, found no GAP executable in PATH`
+          #- gap-package: 'normalizinterface'          # `Error, Some error in the normaliz input data detected: Cone Property in last line not allowed for lattice ideals`
+          #- gap-package: 'packagemanager'             # test failure
```
fingolfin, line 78: this is weird. looking at why PackageManager tests fail, first is this:
```
LoadPackage("curlInterface", false);
# Expected output:
true
# But found:
fail
```
That's clear enough and is waiting for https://github.com/JuliaPackaging/Yggdrasil/pull/9999

But then:
```
# Input is:
InstallPackage("https://github.com/gap-packages/io.git");
# Expected output:
true
# But found:
#I  PackageInfo.g validation failed
#I  There may be problems with the package
true
########
```
I suspect what happens is that it fails to build the manual, or at least the PDF part of the manual, due to missing latex. Then you'd get something like this:
```
gap> ValidatePackageInfo("PackageInfo.g");
#E  component `PDFFile' must be bound to a string denoting a relative path to a readable file
false
gap> ValidatePackageInfo("PackageInfo.g" : quiet);
false
```

So to fix this, we'd have to install `latex` on the runner. That would probably also help GAPDoc and perhaps autodoc.

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          - gap-package: 'example'                    # no jll
+          #- gap-package: 'examplesforhomalg'          # `Error, found no GAP executable in PATH`
+          #- gap-package: 'gapdoc'                     # test failure
+          - gap-package: 'guarana'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'hap'                        # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'hapcryst'                   # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          #- gap-package: 'io'                         # test failure
+          - gap-package: 'itc'                        # dependency `xgap` has no jll
+          #- gap-package: 'localizeringforhomalg'      # `Error, found no GAP executable in PATH`
+          #- gap-package: 'normalizinterface'          # `Error, Some error in the normaliz input data detected: Cone Property in last line not allowed for lattice ideals`
```
fingolfin, line 77:
```suggestion
          - gap-package: 'normalizinterface'          # Tests fail: NormalizInterface currently bundles Normaliz 3.9.3 and its test suite is tuned to that, but our JLL builds it against normaliz_jll which has 3.10.2
```

This could be fixed by an update to NormalizInterface

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          #- gap-package: 'autodoc'                    # test failure
+          #- gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
+          - gap-package: 'curlinterface'              # no jll
+          - gap-package: 'example'                    # no jll
+          #- gap-package: 'examplesforhomalg'          # `Error, found no GAP executable in PATH`
+          #- gap-package: 'gapdoc'                     # test failure
+          - gap-package: 'guarana'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'hap'                        # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'hapcryst'                   # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          #- gap-package: 'io'                         # test failure
```
fingolfin, line 74:
```suggestion
          - gap-package: 'io'                         # segfaults, most likely due to child process handling
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          - gap-package: 'itc'                        # dependency `xgap` has no jll
+          #- gap-package: 'localizeringforhomalg'      # `Error, found no GAP executable in PATH`
+          #- gap-package: 'normalizinterface'          # `Error, Some error in the normaliz input data detected: Cone Property in last line not allowed for lattice ideals`
+          #- gap-package: 'packagemanager'             # test failure
+          - gap-package: 'polenta'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'polycyclic'                 # ???
```
fingolfin, line 80:
```suggestion
          - gap-package: 'polycyclic'                 # test suite currently broken and also disabled by GAP package distribution
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          - gap-package: 'polymaking'                 # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
```
fingolfin, line 81: We could make polymake available, too, but it requires more work, see https://github.com/oscar-system/Oscar.jl/pull/4132/files

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,108 @@
...
+          - '1'
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+        gap-package: ${{fromJSON(needs.generate-matrix.outputs.gap-packages)}}
+        exclude:
+          - gap-package: 'alnuth'                     # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          #- gap-package: 'autodoc'                    # test failure
+          #- gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
```
fingolfin, line 66: This one warrants a closer look but for now it is a genuine issue
```suggestion
          - gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
```

src/packages.jl
```diff
@@ -408,6 +408,66 @@ function build(name::String; quiet::Bool = false,
...
+  end
+
+  disable_error_handler[] = true
+  result = false
+  try
+    with_gap_var("ERROR_OUTPUT", Globals._JULIAINTERFACE_ORIGINAL_ERROR_OUTPUT) do
+      with_gap_var("QuitGap", fake_QuitGap) do
+        with_gap_var("QUIT_GAP", fake_QuitGap) do
+          with_gap_var("ForceQuitGap", identity) do
+            with_gap_var("FORCE_QUIT_GAP", identity) do
```
fingolfin, line 455:
```suggestion
          with_gap_var("ForceQuitGap", fake_QuitGap) do
            with_gap_var("FORCE_QUIT_GAP", fake_QuitGap) do
```
I think that would make WPE pass

lgoettgens: Seems reasonable

.github/workflows/CI-distro.yml
mohamed-barakat: Yes, not having these packages, or not having them tested is a nonissue for us. We clone them again anyway.

fingolfin:
Good: `gapdoc` and `wpe` pass now.

Bad: most other packages now fail despite reporting success:
```
#I  No errors detected while testing

Test Failed at none:1
  Expression: GAP.Packages.test("automata")

ERROR: There was an error during testing
```

lgoettgens:
This is probably due to the return value handling. I'll look into it tomorrow

lgoettgens:
Ok, so the problem is how gap packages try to exit after running their tests.

**Most packages** call `TestDirectory` with `exitGAP:=true` (e.g. https://github.com/gap-packages/datastructures/blob/24f8ea8d4265ffb64a6216a57fabf3aee739cc15/tst/testall.g#L4) (which internally calls `QuitGap` with an exit code. Up to [`5587398` (#1067)](https://github.com/oscar-system/GAP.jl/pull/1067/commits/5587398c955e00f352aa165523909006c545a9db) we used this exit code to determine if the tests succeeded. Afterwards, these packages call `ForceQuitGap(1)` as they think this is an unreachable statement and thus an error if reached (cf https://github.com/gap-packages/datastructures/blob/24f8ea8d4265ffb64a6216a57fabf3aee739cc15/tst/testall.g#L7). Due to  [`5587398` (#1067)](https://github.com/oscar-system/GAP.jl/pull/1067/commits/5587398c955e00f352aa165523909006c545a9db), we now fail due to this exit code of `1`.

**wpe** does not use `TestDirectory`, but instead calls `ForceQuitGap` with an exit code indicating if the tests succeeded (cf. https://github.com/gap-packages/WPE/blob/adc54d808f8b1634c227719f2b41a6fe81373c41/tst/testall.g#L27-L31).

Ways to solve this:
1. Patch wpe to use `QuitGap` instead of `ForceQuitGap`.
2. Only consider the very first call to any of `QuitGap` and `ForceQuitGap` as the try to convey an exit code and ignore all further calls (as they are unreachable in usual GAP execution anyway).

I will try to implement 2.

fingolfin:
Option 2 to me sounds like the way to go as it models more accurately what GAP does.

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,110 @@
...
+        uses: julia-actions/cache@v2
+      - name: "Build package"
+        uses: julia-actions/julia-buildpkg@v1
+      - name: Get list of GAP packages
+        id: set-matrix
+        run: julia --project=. -e '
+                  using Artifacts, TOML;
+                  output = sprint(print, "gap-packages=", map(name -> name[9:end], sort!(collect(filter(startswith("GAP_pkg_"), keys(TOML.parsefile(find_artifacts_toml(Base.active_project()))))))));
+                  println(output);
+                  open(ENV["GITHUB_OUTPUT"], "a") do io;
```
fingolfin, line 43: Not important but FYI I don't think you need the semicolons here

.github/workflows/CI-distro.yml
fingolfin: Indeed it passes now.
```suggestion
```

.github/workflows/CI-distro.yml
fingolfin: I think now only the missing `curlInterface` is an issue
```suggestion
          - gap-package: 'packagemanager'             # tests need curlInterface to passe, reactivate once we have a working JLL for that
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,110 @@
...
+          #- gap-package: 'packagemanager'             # test failure
+          - gap-package: 'polenta'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'polycyclic'                 # test suite currently broken and also disabled by GAP package distribution
+          - gap-package: 'polymaking'                 # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'profiling'                  # segfaults during testing
+          - gap-package: 'radiroot'                   # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'ringsforhomalg'             # `Error, found no GAP executable in PATH`
+          - gap-package: 'semigroups'                 # no jll
+          #- gap-package: 'typeset'                    # needs `dot2tex` executable
+          #- gap-package: 'wpe'                        # ???
```
fingolfin, line 87: WPE works now
```suggestion
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,110 @@
...
+          - gap-package: 'xgap'                       # no jll
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: "Install extra dependencies"
+        run: sudo apt-get install --no-install-recommends texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended
```
fingolfin, line 93: Perhaps `typeset` can be made to work by doing this:
```suggestion
        run: sudo apt-get install --no-install-recommends texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended dot2tex
```
But it might need more; in the GAP PackageDistro CI tests we also install `graphviz` and a few other things, see https://github.com/gap-system/PackageDistro/blob/35c6ee0f972a502ca916d5e7937deae557622d55/tools/gather_dependencies.py#L45

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,110 @@
...
+        uses: julia-actions/cache@v2
+        with:
+          include-matrix: false # don't create a cache per GAP package
+          cache-scratchspaces: false
+      - name: "Build package"
+        uses: julia-actions/julia-buildpkg@v1
+      - name: "Run GAP package tests"
+        run: |
+          JULIA_DEBUG=GAP julia --color=yes --project=. -e 'using GAP'  # ensure initialization
+          ls -l /home/runner/.julia/scratchspaces/*/binarywrappers_v1.11/bin
```
fingolfin, line 109: We don't need this debugging stuff anymore
```suggestion
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,110 @@
...
+          JULIA_DEBUG=GAP julia --color=yes --project=. -e 'using GAP, Test; @test GAP.Packages.test("${{ matrix.gap-package }}")'
```
fingolfin, line 110: We could drop `JULIA_DEBUG=GAP` here. Then again it might be useful for debugging CI failures in the future? Of course that could also be set about other CI tests, so perhaps not...
```suggestion
          julia --color=yes --project=. -e 'using GAP, Test; @test GAP.Packages.test("${{ matrix.gap-package }}")'
```

.github/workflows/CI-distro.yml
lgoettgens: I think I added these to easier copy paste this into my console for debugging ans not getting spammed with output

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,106 @@
...
+          - gap-package: 'normalizinterface'          # Tests fail: NormalizInterface currently bundles Normaliz 3.9.3 and its test suite is tuned to that, but our JLL builds it against normaliz_jll which has 3.10.2
+          - gap-package: 'packagemanager'             # tests need curlInterface to passe, reactivate once we have a working JLL for that
+          - gap-package: 'polenta'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'polycyclic'                 # test suite currently broken and also disabled by GAP package distribution
+          - gap-package: 'polymaking'                 # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'profiling'                  # segfaults during testing
+          - gap-package: 'radiroot'                   # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'ringsforhomalg'             # `Error, found no GAP executable in PATH`
+          - gap-package: 'semigroups'                 # no jll
+          #- gap-package: 'typeset'                    # needs `dot2tex` executable
```
fingolfin, line 85: Works now
```suggestion
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,106 @@
...
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+        gap-package: ${{fromJSON(needs.generate-matrix.outputs.gap-packages)}}
+        exclude:
+          - gap-package: 'alnuth'                     # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          #- gap-package: 'autodoc'                    # test failure
+          - gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
+          - gap-package: 'curlinterface'              # no jll
```
fingolfin, line 67: see PR #1109

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,106 @@
...
+          - '1'
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+        gap-package: ${{fromJSON(needs.generate-matrix.outputs.gap-packages)}}
+        exclude:
+          - gap-package: 'alnuth'                     # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          #- gap-package: 'autodoc'                    # test failure
+          - gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
```
fingolfin, line 66: Should also be fixed by PR see PR https://github.com/oscar-system/GAP.jl/pull/1109

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,106 @@
...
+        julia-version:
+          - '1'
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+        gap-package: ${{fromJSON(needs.generate-matrix.outputs.gap-packages)}}
+        exclude:
+          - gap-package: 'alnuth'                     # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          #- gap-package: 'autodoc'                    # test failure
```
fingolfin, line 65:
```suggestion
          - gap-package: 'autodoc'                    # tries and fails to build its own documentation, inside the (partially write protected) artifacts dir
```

.github/workflows/CI-distro.yml
lgoettgens:
```suggestion
          # - gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
```

.github/workflows/CI-distro.yml
lgoettgens:
```suggestion
          # - gap-package: 'curlinterface'              # no jll
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,105 @@
...
+          - '1'
+        julia-arch:
+          - x64
+        os:
+          - ubuntu-latest
+        gap-package: ${{fromJSON(needs.generate-matrix.outputs.gap-packages)}}
+        exclude:
+          - gap-package: 'alnuth'                     # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'autodoc'                    # tries and fails to build its own documentation, inside the (partially write protected) artifacts dir
+          # - gap-package: 'caratinterface'             # `Carat program Bravais_inclusions failed with error code 1`
```
lgoettgens, line 66:
```suggestion
```
Works

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,105 @@
...
+          # - gap-package: 'curlinterface'              # no jll
```
lgoettgens, line 67:
```suggestion
```
Works

lgoettgens:
The `help` package still has some failures. Could you have a look at it?

fingolfin:
Hmm, there also was a segfault in the `atlasrep` test suite (looks GC related :-/)

And another in `guava` which looks like maybe an encoding issue (?) but which I also can't reproduce locally
```
########> Diff in GAPROOT/pkg/GAP_pkg_guava/guava3.19/tst/guava.tst:649
# Input is:
p := CodeIsomorphism( C1, C2 );
# Expected output:
(2,13,7,10,8,3,5,4,14)(12,15)
# But found:
Syntax error: expression expected in /tmp/gaptempfile.5zTo2m:1
GUAVA_TEMP_VAR := � &
                  ^
Syntax error: ; expected in /tmp/gaptempfile.5zTo2m:3
false
########
```


In order to move things along, how about we disable those for now and merge this, then file issues about these failures and then can investigate these later.

lgoettgens:
After re-running it is only `help` failing. I'll go ahead disabling this and anything that randomly fails in the next few CI runs.

lgoettgens:
All distro tests are green now. The one failure is https://github.com/oscar-system/GAP.jl/issues/1111.

src/packages.jl
```diff
@@ -453,6 +453,69 @@ function build_recursive(name::String; quiet::Bool = false,
   return true
 end
 
+"""
+    test(name::String)
+
+Return a boolean indicating if the GAP package with name `name` succeeds
+in running its tests.
+
+It is inteded to be used with the `@test` macro from the `Test` package.
```
fingolfin, line 462:
```suggestion
It is intended to be used with the `@test` macro from the `Test` package.
```

src/packages.jl
```diff
@@ -453,6 +453,69 @@ function build_recursive(name::String; quiet::Bool = false,
   return true
 end
 
+"""
+    test(name::String)
+
+Return a boolean indicating if the GAP package with name `name` succeeds
+in running its tests.
```
fingolfin, line 460: I think we should not insist on using the `Return ...` format here, as the result is IMHO somewhat unnatural and makes it unclear whether the tests are actually run, which arguably is the main point of this function. How about this:
```suggestion
Run the tests of GAP package `name` and return a boolean indicating whether
they succeeded (`true`) or not.
```

src/packages.jl
```diff
@@ -453,6 +453,69 @@ function build_recursive(name::String; quiet::Bool = false,
...
+
+It is inteded to be used with the `@test` macro from the `Test` package.
+
+The function uses [the function `TestPackage`](GAP_ref(ref:TestPackage)).
```
fingolfin, line 464:
```suggestion
The function uses [the GAP function `TestPackage`](GAP_ref(ref:TestPackage)).
```

src/packages.jl
```diff
@@ -453,6 +453,69 @@ function build_recursive(name::String; quiet::Bool = false,
...
+"""
+function test(name::String)
+  global disable_error_handler
+
+  function with_gap_var(f, name::String, val)
+    gname = GapObj(name)
+    old_value = Globals.ValueGlobal(gname)
+    Globals.MakeReadWriteGlobal(gname)
+    Globals.UnbindGlobal(gname)
+    Globals.BindGlobal(gname, val)
```
fingolfin, line 474: I think this would also work?
```suggestion
    replace_global!(gname, val)
```

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,75 @@
+name: Test GAP package distro
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
fingolfin: There are >150 jobs but apparently they take less than 15 minutes to run? So perhaps we keep them on for now and see what happens?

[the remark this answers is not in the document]
lgoettgens: Fine with me

src/packages.jl
```diff
@@ -453,6 +453,65 @@ function build_recursive(name::String; quiet::Bool = false,
...
+they succeeded (`true`) or not.
+
+It is intended to be used with the `@test` macro from the `Test` package.
+
+The function uses [the GAP function `TestPackage`](GAP_ref(ref:TestPackage)).
+"""
+function test(name::String)
+  global disable_error_handler
+
+  function with_gap_var(f, name::String, val)
```
fingolfin, line 469:
```suggestion
  function with_gap_var(f, name::Symbol, val)
```

src/packages.jl
```diff
@@ -453,6 +453,65 @@ function build_recursive(name::String; quiet::Bool = false,
...
+    gname = GapObj(name)
+    old_value = Globals.ValueGlobal(gname)
+    replace_global!(gname, val)
+    try
+        f()
+    finally
+      replace_global!(gname, old_value)
```
fingolfin, line 476:
```suggestion
    replace_global!(name, val)
    try
        f()
    finally
      replace_global!(name, old_value)
```

src/packages.jl
```diff
@@ -453,6 +453,65 @@ function build_recursive(name::String; quiet::Bool = false,
...
+  end
+
+  disable_error_handler[] = true
+  result = false
+  try
+    with_gap_var("ERROR_OUTPUT", Globals._JULIAINTERFACE_ORIGINAL_ERROR_OUTPUT) do
+      with_gap_var("QuitGap", fake_QuitGap) do
+        with_gap_var("QUIT_GAP", fake_QuitGap) do
+          with_gap_var("ForceQuitGap", fake_QuitGap) do
+            with_gap_var("FORCE_QUIT_GAP", fake_QuitGap) do
```
fingolfin, line 498:
```suggestion
    with_gap_var(:ERROR_OUTPUT, Globals._JULIAINTERFACE_ORIGINAL_ERROR_OUTPUT) do
      with_gap_var(:QuitGap, fake_QuitGap) do
        with_gap_var(:QUIT_GAP, fake_QuitGap) do
          with_gap_var(:ForceQuitGap, fake_QuitGap) do
            with_gap_var(:FORCE_QUIT_GAP, fake_QuitGap) do
```

fingolfin (approved):

.github/workflows/CI-distro.yml
```diff
@@ -0,0 +1,105 @@
...
+          - gap-package: 'autodoc'                    # tries and fails to build its own documentation, inside the (partially write protected) artifacts dir
+          - gap-package: 'example'                    # no jll
+          - gap-package: 'examplesforhomalg'          # `Error, found no GAP executable in PATH`
+          - gap-package: 'guarana'                    # `AL_EXECUTABLE, the executable for PARI/GP, has to be set`
+          - gap-package: 'hap'                        # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'hapcryst'                   # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
+          - gap-package: 'help'                       # test failure in HeLP-4.0/tst/yes_4ti2.tst:39
+          - gap-package: 'io'                         # segfaults, most likely due to child process handling
+          - gap-package: 'itc'                        # dependency `xgap` has no jll
+          - gap-package: 'localizeringforhomalg'      # `Error, found no GAP executable in PATH`
```
fingolfin, line 75: See also issue #1105 (@ThomasBreuer that's another case where a package needs/wants a GAP executable)

PR #1068, opened 2024-11-29
More C code refactoring

fingolfin:
- instead of retrieving Julia modules by a string name and assuming they are bound in `Main`, add dedicated kernel functions for fetching `Main` and `Core` modules
- remove internal `get_module` helper

### DIFF
--- a/pkg/JuliaInterface/tst/utils.tst
+++ b/pkg/JuliaInterface/tst/utils.tst
@@ -49,15 +49,7 @@ gap> _JuliaGetGlobalVariableByModule(0, 0);
 Error, _JuliaGetGlobalVariableByModule: <name> must be a string (not the integ\
 er 0)
 gap> _JuliaGetGlobalVariableByModule("sqrt", 0);
-Error, _JuliaGetGlobalVariableByModule: <module> must be a string or a Julia m\
-odule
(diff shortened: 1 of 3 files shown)


pkg/JuliaInterface/tst/utils.tst
```diff
@@ -49,17 +49,7 @@ gap> _JuliaGetGlobalVariableByModule(0, 0);
...
-odule
-gap> _JuliaGetGlobalVariableByModule("sqrt", Julia.Base.sqrt);
-Error, _JuliaGetGlobalVariableByModule: <module> must be a string or a Julia m\
-odule
-gap> _JuliaGetGlobalVariableByModule("Base","sqrt");
-Error, sqrt is not a module
-gap> _JuliaGetGlobalVariableByModule("sqrt","Base");
-<Julia: sqrt>
-gap> _JuliaGetGlobalVariableByModule("sqrt", Julia.Base);
-<Julia: sqrt>
```
lgoettgens, line 62: This test is lines 61-62 still works and can thus be kept, maybe with `Julia` as the second arg instead of `Julia.Base`.

ThomasBreuer (approved):

Issue #993, opened 2024-05-22
Revise setup of the faux GAP root


fingolfin:
When starting multiple Julia's and using GAP.jl in each of them, there can be issues caused by the shared faux gaproot in a scratch space, despite pidlocking.

To quote myself from Slack:

> I guess there is a problem if process 1 loads GAP.jl and is past creating the scratch space and actively loading GAP; then process 2 deletes and recreates the scratch space. Then for a brief moment, the faux GAP root is gone and reads from paths relative to it will fail.
Strictly speaking this can of course also happen at any point after GAP initialization: if something deletes the scratch space it uses, then we have a problem.
>
> Two solutions comes to mind:
> - let each GAP.jl instance have its own faux gaproot. To that end, each process creates its own subdir of the GAP.jl scratch space (e.g. with a name based on its PID, possibly plus a random part). Use mktempdir(gap_scratch_space_dir; cleanup = true) to get the path, ensuring that it is deleted afterwards. In that case the pid locking could also go, as it would be "built in", so to say.
> - change the code re-creating the faux gaproot to be more defensive: force_symlink could be made less brutal: instead of force creating the new symlink, check if it already exists and points at the right path, if it does, just leave it alone. Likewise the code which recreates sysinfo.gap & gac could first assemble the new file content in a string; then read the existing file and compare; and only overwrite it if the content changed.
>
> With approach 2, there could still cause problems if there are actual changes in the symlink paths. To protect against that, the (hashes of) the paths to the relevant artifact  (gap_packages, GAP_jll, GAP_lib_jll) could become part of the Scratchspace name.

fingolfin:
Resolved by #1095: while we still have the faux/fake GAP root, we only need it for `sysinfo.gap` and `gac` and neither can cause the kind of problem mentioned here

PR #1069, opened 2024-12-07
Use one artifact for each deposited GAP package

fingolfin:
This is using the packages for GAP 4.14.0.

### DIFF
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -1,16 +1,17 @@
 module Setup
 
 using Pkg: GitTools
-import Artifacts: @artifact_str
+import Artifacts: find_artifacts_toml, @artifact_str
 import GAP_jll
 import GAP_lib_jll
 import GAP_pkg_juliainterface_jll
 import Scratch: @get_scratch!
 import Pidfile
+import TOML
 
 # to separate the scratchspaces of different GAP.jl copies and Julia versions
 # put the Julia version and the hash of the path to this file into the key
-const scratch_key = "gap_$(string(hash(@__FILE__)))_$(VERSION.major).$(VERSION.minor)"
+const scratch_key = "gap_$(hash(@__FILE__))-$(VERSION.major).$(VERSION.minor)"
 
 gaproot() = @get_scratch!(scratch_key)
 
@@ -195,8 +196,12 @@ function regenerate_gaproot()
         force_symlink("../../gac",
                       joinpath(gaproot_mutable, "bin", sysinfo["GAParch"], "gac"))
 
-        # create a `pkg` symlink to the GAP packages artifact
-        force_symlink(artifact"gap_packages", "$gaproot_mutable/pkg")
+        # create a `pkg` directory with symlinks to all the GAP packages artifacts
+        mkpath(joinpath(gaproot_mutable, "pkg"))
+        pkg_artifacts = filter(startswith("GAP_pkg_"),keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
+        for name in pkg_artifacts
+            force_symlink(@artifact_str(name), joinpath(gaproot_mutable, "pkg", name))
+        end
 
     end # mkpidlock
 
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -198,7 +198,7 @@ function regenerate_gaproot()
 
         # create a `pkg` directory with symlinks to all the GAP packages artifacts
         mkpath(joinpath(gaproot_mutable, "pkg"))
-        pkg_artifacts = filter(startswith("GAP_pkg_"),keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
+        pkg_artifacts = filter(startswith("GAP_pkg_"), keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
         for name in pkg_artifacts
             force_symlink(@artifact_str(name), joinpath(gaproot_mutable, "pkg", name))
         end
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,6 +1,1619 @@
-[gap_packages]
-git-tree-sha1 = "a24cceaecaac91adaf0f67802e4670ff5de4afa7"
+[GAP_pkg_4ti2interface]
+git-tree-sha1 = "0cfc2ae54301261dad1f4fa233047ded8e719cfc"
 
-    [[gap_packages.download]]
-    sha256 = "77b1dfaf3ec6cc9e45aab2ed3f33187fc507e18f3c0bc13bbf6741748f0c6e7c"
-    url = "https://github.com/gap-system/PackageDistro/releases/download/v4.13.1/packages.tar.gz"
+    [[GAP_pkg_4ti2interface.download]]
+    sha256 = "6716431f63a5b29e20333b2bc4b432198528e9b50c28dea988658962857659e7"
+    url = "https://github.com/homalg-project/homalg_project/releases/download/4ti2Interface-2024.11-01/4ti2Interface-2024.11-01.tar.gz"
+    [[GAP_pkg_4ti2interface.download]]
+    sha256 = "6716431f63a5b29e20333b2bc4b432198528e9b50c28dea988658962857659e7"
+    url = "https://files.gap-system.org/pkg/4ti2Interface-2024.11-01.tar.gz"
+
+[GAP_pkg_ace]
+git-tree-sha1 = "c67becc10447fae9af91c022bc44f9ea17157934"
+
+    [[GAP_pkg_ace.download]]
+    sha256 = "9cc94cd6327147b8b538341ba7594e4252a80ca7c960ac6749224efa437c9b66"
+    url = "https://github.com/gap-packages/ace/releases/download/v5.6.2/ace-5.6.2.tar.gz"
+    [[GAP_pkg_ace.download]]
+    sha256 = "9cc94cd6327147b8b538341ba7594e4252a80ca7c960ac6749224efa437c9b66"
+    url = "https://files.gap-system.org/pkg/ace-5.6.2.tar.gz"
+
+[GAP_pkg_aclib]
+git-tree-sha1 = "e91708d23a46159c10291635d44472f2ea9d10a4"
+
+    [[GAP_pkg_aclib.download]]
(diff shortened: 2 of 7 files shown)


mohamed-barakat:
An artifact of a GAP-package with name `XY` will not hijack the name for a Julia package `XY.jl`, right?

I am asking because we are gradually moving some GAP packages to Julia with no `GAP.jl` dependency. `CAP.jl` ist already an official Julia package.

fingolfin:
Artifact names are internal to a package and do not affect other packages.

lgoettgens:
This is the debug log of the download failures that we talked about earlier today (at the example of `gap_pkg_homalg` which was the first one attempted by my julia 1.11.2):
```julia
(GAP) pkg> resolve
┌ Debug: Downloading artifact from Pkg server
│   name = "GAP_pkg_homalg"
│   artifacts_toml = "/home/lgoe/code/julia/GAP.jl/Artifacts.toml"
│   platform = Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.11.2, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.30}
│   url = "https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13"
└ @ Pkg.Artifacts ~/.julia/juliaup/julia-1.11.2+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/src/Artifacts.jl:421
  Downloading artifact: GAP_pkg_homalg
┌ Debug: download and verify failed on attempt 1/3
│   url = "https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13"
│   dest = "/tmp/jl_WN2HIJEJQN-download.gz"
│   err = RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13
└ @ Pkg.PlatformEngines ~/.julia/juliaup/julia-1.11.2+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/src/PlatformEngines.jl:358
┌ Debug: download and verify failed on attempt 2/3
│   url = "https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13"
│   dest = "/tmp/jl_WN2HIJEJQN-download.gz"
│   err = RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13
└ @ Pkg.PlatformEngines ~/.julia/juliaup/julia-1.11.2+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/src/PlatformEngines.jl:358
┌ Debug: download and verify failed on attempt 3/3
│   url = "https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13"
│   dest = "/tmp/jl_WN2HIJEJQN-download.gz"
│   err = RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13
└ @ Pkg.PlatformEngines ~/.julia/juliaup/julia-1.11.2+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/src/PlatformEngines.jl:358
┌ Debug: download_artifact error
│   tree_hash = SHA1("2881b49ed9fc87cd0040df78e3c9bde8f5151f13")
│   tarball_url = "https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13"
│   tarball_hash = nothing
│   err = RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13
  Failure artifact: GAP_pkg_homalg
┌ Debug: Failed to download artifact from Pkg server
│   download_success = RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/2881b49ed9fc87cd0040df78e3c9bde8f5151f13
└ @ Pkg.Artifacts ~/.julia/juliaup/julia-1.11.2+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/src/Artifacts.jl:430
┌ Debug: Downloading artifact
│   name = "GAP_pkg_homalg"
│   artifacts_toml = "/home/lgoe/code/julia/GAP.jl/Artifacts.toml"
│   platform = Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.11.2, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.30}
│   url = "https://github.com/homalg-project/homalg_project/releases/download/homalg-2024.01-01/homalg-2024.01-01.tar.gz"
└ @ Pkg.Artifacts ~/.julia/juliaup/julia-1.11.2+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/src/Artifacts.jl:446
  Downloading artifact: GAP_pkg_homalg
  Downloaded artifact: GAP_pkg_homalg
```
We can see that Pkg tries to find the artifact in the Pkg servers three times, and only then starts using your provided urls.

lgoettgens:
> When downloading and installing the `socrates` artifact, URLs will be attempted in order until one succeeds.
> (https://pkgdocs.julialang.org/v1/artifacts/#Artifact-types-and-properties)

so the order is indeed relevant

fingolfin:
@lgoettgens yeah, I verified all that before submitting this PR ;-) thanks for double checking, never hurts

fingolfin:
Initial script for importing the artifacts data is now in place, too.

I don't understand right now why it fails to build in Julia 1.6, but I'll wait with looking into that until also the JLLs are updated.

fingolfin:
Hu, weird, the Julia 1.6 failures are also on master and started yesterday (first failure on master [here](https://github.com/oscar-system/GAP.jl/actions/runs/12209497467/job/34064385387#step:10:163) while it was OK the day before [in this log](https://github.com/oscar-system/GAP.jl/actions/runs/12192337764))

src/GAP.jl
```diff
@@ -118,6 +118,9 @@ function initialize(argv::Vector{String})
     # (which is created by `setup.jl`).
     append!(argv, ["--systemfile", abspath(@__DIR__, "..", "gap", "systemfile.g")])
 
+    # disable loading of all but the absolutely required packages
+    push!(argv, "-A")
```
ThomasBreuer, line 122: The addition of `-A` under the condition `haskey(ENV, "GAP_BARE_DEPS")` should then be removed from `__init__`

lgoettgens:
With the change of GAP from `4.13` to `4.14.0`, we should change the used GAP branch in `gap.yml` from `stable-4.13` to `stable-4.14` as well.

fingolfin:
I've now reduced this again to *just* replace the single monolithic package artifact with one artifact per package. I think this was already working reasonably well (?) but we'll see.

src/setup.jl
```diff
@@ -1,16 +1,17 @@
...
 import GAP_lib_jll
 import GAP_pkg_juliainterface_jll
 import Scratch: @get_scratch!
 import Pidfile
+import TOML
 
 # to separate the scratchspaces of different GAP.jl copies and Julia versions
 # put the Julia version and the hash of the path to this file into the key
-const scratch_key = "gap_$(string(hash(@__FILE__)))_$(VERSION.major).$(VERSION.minor)"
+const scratch_key = "gap_$(hash(@__FILE__))-$(VERSION.major).$(VERSION.minor)"
```
lgoettgens, line 14: IIuc this change is here to not use the same scratchspace as before as that could lead to conflicts, right? Wouldn't it make sense to either include some kind of increasing number in either the key or in the hash input, so we can bump that again for the next incompatible change in the future?

src/setup.jl
```diff
@@ -195,8 +196,12 @@ function regenerate_gaproot()
         force_symlink("../../gac",
                       joinpath(gaproot_mutable, "bin", sysinfo["GAParch"], "gac"))
 
-        # create a `pkg` symlink to the GAP packages artifact
-        force_symlink(artifact"gap_packages", "$gaproot_mutable/pkg")
+        # create a `pkg` directory with symlinks to all the GAP packages artifacts
+        mkpath(joinpath(gaproot_mutable, "pkg"))
+        pkg_artifacts = filter(startswith("GAP_pkg_"),keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
```
lgoettgens, line 201:
```suggestion
        pkg_artifacts = filter(startswith("GAP_pkg_"), keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
```

src/setup.jl
fingolfin: I actually had an intermediate version where I computed a more complicated hash, which involved the paths of the various artifacts (!) as those can also change; but also checksums of e.g. `setup.jl` and other files, but then realized this doesn't solve all the problems either... 

Here are some further alternatives:
- as above, include more data into the hash (e.g. GAP.jl version, location of the various JLLs and artifacts we link to, ...)
- stop using a scratch space, instead just create a fresh "mutable root dir" each time the package is loaded in a fresh temp dir (of course that means a *lot* of temp dirs would accumulate... we could add an `atexit` handler to delete them at the end?)
   - would be interesting how much time we spend creating the mutable root dir
- we could adjust the setup code so that whenever it e.g. creates a dir or symlink it checks if there is already an object of the correct type there -- i.e. if we created a dir `pkg` and there is already a dir `pkg`, do nothing; but if we create a symlink `pkg` pointing to `targetA` and there is instead a directory `pkg`; or a symlink pointing to `anotherTarget`, then delete `pkg`.
  - this seems much more complicated and in every way I can think of inferior to the solution of just rebuilding it all from scratch, *except* that at least the path to this root dir is somewhat stable and predictable
  - strictly speaking we'd also have to scan for obsolete file system objects that should be removed... making this even more complicated...

lgoettgens:
The timeouts are probably happening as the new artifacts are not cached yet, and (even worse) Pkg will first try to download them from the Pkg servers 3 times before using the provided github source. This overhead should reduce once there is a release with these new artifacts.

src/setup.jl
fingolfin: See also https://github.com/oscar-system/GAP.jl/issues/993

lgoettgens (approved):

Issue #1071, opened 2024-12-09
CI tests against Julia 1.6 started to fail around 7th December 2024


fingolfin:
Julia 1.6 failures are also on master and started Friday (first failure on master [here](https://github.com/oscar-system/GAP.jl/actions/runs/12209497467/job/34064385387#step:10:163) while it was OK the day before [in this log](https://github.com/oscar-system/GAP.jl/actions/runs/12192337764))

_Originally posted by @fingolfin in https://github.com/oscar-system/GAP.jl/issues/1069#issuecomment-2525308610_

PR #1073, opened 2024-12-09
Add libstdc++ workaround for CI with julia 1.6

lgoettgens:
Maybe (hopefully) resolves https://github.com/oscar-system/GAP.jl/issues/1071.

This just adds the patch from https://github.com/oscar-system/GAP.jl/issues/1071#issuecomment-2527552438.


lgoettgens:
The remaining failure with GAP master looks like it is due to https://github.com/gap-system/gap/pull/5859, but that is a different topic.

fingolfin (approved):
Thanks for debugging this and making a patch!

PR #1074, opened 2024-12-09
Version 0.13.0-DEV and switch to GAP 4.14.0

fingolfin:
It will take some more time to hone PR #1069. In the meantime I'd like to already switch master to using GAP 4.14.0 to get the gap-system/gap CI tests to pass again.


lgoettgens (approved):

fingolfin:
The failure in `CI with GAP / Julia 1 - GAP master - macOS-latest (pull_request)` really confuses me:
```
GAP --code-coverage=none --color=auto --check-bounds=yes --warn-overwrite=yes --depwarn=error --inline=yes --startup-file=no --track-allocation=none 

Failed to precompile GAP [c863536a-3901-11e9-33e7-d5cd0df7b904] to "/var/folders/0w/4z5l9vds32nbkz7l22n8j6s80000gn/T/jl_K2kFra/compiled/v1.11/GAP/jl_GqF5bM".
ERROR: LoadError: InitError: could not load library "/tmp/gap_jll_override/lib/libgap.9.dylib"
dlopen(/tmp/gap_jll_override/lib/libgap.9.dylib, 0x0001): tried: '/tmp/gap_jll_override/lib/libgap.9.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/tmp/gap_jll_override/lib/libgap.9.dylib' (no such file), '/tmp/gap_jll_override/lib/libgap.9.dylib' (no such file), '/private/tmp/gap_jll_override/lib/libgap.9.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/private/tmp/gap_jll_override/lib/libgap.9.dylib' (no such file), '/private/tmp/gap_jll_override/lib/libgap.9.dylib' (no such file)
```
Now of course GAP 4.14.0 now has `libgap.10`, so indeed there won't be a `libgap.9`.

But I can't see why anything would even *try* to use the old libgap.9. 

Note that the corresponding job on Ubuntu passes, as does `CI with GAP / Julia 1 - GAP stable-4.14 - macOS-latest `.

lgoettgens:
It's this line:
https://github.com/JuliaBinaryWrappers/GAP_jll.jl/blob/ce44bbd8a4af7f955dfa0aa13fe287529948f20c/src/wrappers/aarch64-apple-darwin-julia_version%2B1.11.1.jl#L14

`run_with_override.jl` (the hacky script used in the `CI with GAP` jobs) only replaces the used artifacts. However, the wrapper code in `GAP_jll.jl` has the kernel version hardcoded. The analogous file for linux is https://github.com/JuliaBinaryWrappers/GAP_jll.jl/blob/main/src/wrappers/aarch64-linux-gnu-julia_version%2B1.11.1.jl#L14 and does not refer to the kernel version at this place. Probably this is a reason why the linux jobs still succeed

fingolfin:
That explains it -- so GAP_jll contains nonsense. But how is *that* happening now? Why does the GAP_jll wrapper for GAP 4.14.0 have `libgap.9` when it should be `libgap.10`.

And I still wonder why the same problem does not occur with the `CI with GAP / Julia 1 - GAP stable-4.14 - macOS-latest`

lgoettgens:
> That explains it -- so GAP_jll contains nonsense. But how is _that_ happening now? Why does the GAP_jll wrapper for GAP 4.14.0 have `libgap.9` when it should be `libgap.10`.
> 
> And I still wonder why the same problem does not occur with the `CI with GAP / Julia 1 - GAP stable-4.14 - macOS-latest`

That's because https://github.com/gap-system/gap/pull/5859 was only merged to the gap master AFTER you branched off the stable-4.14 branch. Thus GAP 4.14.0 has dylib version 9.1, but GAP 4.15.0-DEV has dylib version 10.

fingolfin:
Yeah I just realized this, too :facepalm: I had it backwards.

So I guess we have to either disable testing against GAP master, or implement a workaround.

lgoettgens:
> Yeah I just realized this, too 🤦 I had it backwards.
> 
> So I guess we have to either disable testing against GAP master, or implement a workaround.

I am playing around with a workaround right now

PR #1075, opened 2024-12-10
CI: don't specify runner arch

fingolfin:
macOS runners use native ARM these days, don't request x86 on them


lgoettgens (approved):
taking the default sounds sensible to do

PR #1076, opened 2024-12-10
Use JLLs to provide compiled binaries for some GAP packages

### LINKED ISSUE #852: Again trouble starting OSCAR while offline due to GAP issues
Reported by @ederc (who'll provide details if he still has them).

Ideal solution is to finally get GAP_pkg_*.jl out of the door....

### LINKED ISSUE #928: Trouble installing GAP package `nq` (and hence `hap`) on macOS

### LINKED ISSUE #1064: Error installing nq, anupq because GMP can't be found
```
julia> GAP.Packages.install("nq"; debug=true)
#I  Getting PackageInfo URLs...
#I  Retrieving PackageInfo.g from https://gap-packages.github.io/nq/PackageInfo.g ...
#I  The newest version of package "nq" is already installed
#I  Package ``nq'': The executable program is not available
#I  Package ``nq'': The executable program is not available
#I  Running compilation script on /Users/mhorn/.julia/gaproot/v4.13/pkg/nq-2.5.11/ ...
#I  Possible error detected, see log:
#I    + GAPROOT=/Users/mhorn/.julia/scratchspaces/c863536a-3901-11e9-33e7-d5cd0df7b904/gap_7216319432009064789_1.11
#I    + PKGDIR=/Users/mhorn/.julia/gaproot/v4.13/pkg/nq-2.5.11/
#I    + command -v gmake
...
+ ./configure --with-gaproot=/Users/mhorn/.julia/scratchspaces/c863536a-3901-11e9-33e7-d5cd0df7b904/gap_7216319432009064789_1.11
checking for a BSD-compatible install... /opt/homebrew/bin/ginstall -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /opt/homebrew/bin/gmkdir -p
checking for gawk... no
...
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for gmp.h... no
configure: error: Could not locate GMP, the GNU multiprecision library

#I  Package ``nq'': The executable program is not available
#I  Package ``nq'': The executable program is not available
#I  Checking dependencies for nq...
#I    polycyclic 2.11: true
#I  Package ``nq'': The executable program is not available
#I  Package ``nq'': The executable program is not available
#I  Package availability test failed
```

I tried fixing this by modifying `setup.jl` to do this (after suitably implementing `gmp_artifact_dir`)
```
  sysinfo["GMP_PREFIX"] = gmp_artifact_dir()
```
This then builds, however the resulting `nq` executable does not run standalone on macOS (which is what I expected):
```
dyld[56596]: Library not loaded: '@rpath/libgmp.10.dylib'
  Referenced from: '/Users/mhorn/.julia/gaproot/v4.13/pkg/nq-2.5.11/bin/aarch64-apple-darwin20-julia1.11-64-kv9/nq'
  Reason: tried: '/usr/local/lib/libgmp.10.dylib' (no such file), '/usr/lib/libgmp.10.dylib' (no such file)
Abort trap: 6
```

Perhaps I can fix this by ensuring the `nq` executable gets a suitable `@rpath` definition embedded... gotta play with `LDFLAGS` etc. (and then make sure I can pass them on to the `nq` build system... huh). 

Of course the real plan is still to use JLLs to deal with this issue (instead of running a C compiler); however the binaries in the JLL will actually have the same issue. So we may need to also use `BinaryWrappers.jl` (in either `GAP_pkg_nq.jl`, or `GAP.jl`, or even both?!?)

fingolfin:
- **Stop autoloading for most GAP packages**
- **Use JLLs to provide compiled binaries for some GAP packages**
- **HACK HACK HACK do not merge**

Resolves #1064
Resolves #928
Resolves #852


.github/workflows/CI.yml
```diff
@@ -69,18 +69,14 @@ jobs:
       - name: "GAP tests"
         run: |
           julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")'
-          export GAP="/tmp/gap.sh -A --quitonbreak --norepl"
+          export GAP="/tmp/gap.sh --quitonbreak --norepl"
```
fingolfin, line 72: So this is a nasty hack: because we specify `-A` now any places that already specifies `-A` needs to remove it... But IMHo that solution doesn't scale well.

Alternative ideas for resolving this:
1. we filter the args being passed in to remove any `-A` arguments
2. we don't use `-A` to reduce the loaded packages, but rather `--bare`
   - optionally we could afterwards "manually" load the required packages gapdoc/smallgrp/transgrp/primgrp
3. we don't use `-A` nor `--bare` and instead patch the GAP library directly to support the "overrides" (i.e. `GAP_lib_jll` for now and possibly later upstream GAP)
   - if we do it right then `DirectoriesPackageProgramsOverrides` exists by the time `gap/systemfile.g` is executed and we can fill it then. 
   - this would also allow us to load the default set of GAP packages again (I am torn whether we want that, or want `--bare`, or `-A` ...)
4. as a variant of that we could also use another dirty hack and run the code from `gap/pkg.g` (which patches `DirectoriesPackagePrograms`) from inside `pkg/JuliaInterface/PackageInfo.g` after making sure that `JuliaInterface` is the first package to be loaded...

ThomasBreuer: Currently my point of view is:

When starting GAP from GAP.jl then we should  load JuliaInterface but no other GAP packages.
This way we allow Oscar.jl (or others) to prepare special settings that will influence what happens when certain GAP packages will get loaded, and it is possible to provide/prescribe special packages versions and perhaps to load these packages.

We can provide a small function in GAP.jl that triggers GAP's `LoadPackages` for those packages that are "wanted by GAP" and that have not yet been loaded.
The question is then how this function can be called automatically but late enough such that the above order of package loading does not get distorted.

fingolfin:
Summary discussion with @ThomasBreuer: we plan to let GAP.jl use the equivalent of `--bare` for GAP, i.e. doing `using GAP` will only load `JuliaInterface` (and *maybe* GAPDoc to enable help? or maybe we delay this until someone accesses the help, dunno).

For the `gap.sh` scripts we could still load the default "autoload" packages, to be decided.

This PR however does *not* do that right now, it is as close as possible to the current status quo -- my idea here being that we can merge this independently from the other change, this is simpler to review and may also help to bisect issues.

fingolfin:
I think this is ready for merging now

lgoettgens:
I don't understand enough about GAP package loading to give educated feedback on this.

Maybe a stupid question: The long term plan is still to have `GAP_pkg_foo.jl` packages that wrap `GAP_pkg_foo_jll.jl`, right? And this here is more of an intermediate solution

fingolfin:
The long term plan still is to introduce `GAP_pkg_foo.jl`, but there are known problems with that (the Julia and GAP package system are *very* different), so it makes sense to have this "intermediate" solution -- it may be what we use for a year or longer; and perhaps we come up with a completely different solution; we'll see. But most likely whatever we use will use the same artifacts and JLLs, which is good.

Issue #840, opened 2022-11-11
Factor out check for C/C++ compiler (so it doesn't slow down our startup each time)


fingolfin:
Not yet sure how to best do it. Perhaps we should use [`Preferences.jl`](https://github.com/JuliaPackaging/Preferences.jl) ? 

Or perhaps we compute them lazily, by moving the checks into `gac` ?

fingolfin:
Another option would be to compute/update these only inside `Packages.install`.

PR #1079, opened 2024-12-10
Use temp dir (no scratch space) for mutable GAP root

### LINKED ISSUE #993: Revise setup of the faux GAP root

fingolfin:
Resolves #993 -- I measured that it is not slower than the current approach which fully regenerates the GAP root each time anyway.

Specifically on my test machines it takes about 0.09 - 0.11 seconds. Which is not nothing, but still relatively little.

An alternative would be to keep using a scratch space but be more aggressive in switching to a new one when anything changes, i.e. incorporating a hash value into its name that involves more components, such as
- a hash of the content of `setup.jl`
- hashes of the artifact dirs of all involved artifacts / JLLs
- ...


src/packages.jl
```diff
@@ -239,7 +237,7 @@ For details, please refer to its documentation.
 function install(spec::String, version::String = "";
                                interactive::Bool = true, quiet::Bool = false,
                                debug::Bool = false,
-                               pkgdir::AbstractString = DEFAULT_PKGDIR[])
+                               pkgdir::AbstractString = GAP.sysinfo["DEFAULT_PKGDIR"])
```
lgoettgens, line 240:
```suggestion
                               pkgdir::AbstractString = sysinfo["DEFAULT_PKGDIR"])
```
this is a submodule of `GAP`, but `sysinfo` gets already imported

src/packages.jl
```diff
@@ -306,7 +304,7 @@ For details, please refer to its documentation.
 """
 function update(spec::String; interactive::Bool = true, quiet::Bool = false,
                               debug::Bool = false,
-                              pkgdir::AbstractString = DEFAULT_PKGDIR[])
+                              pkgdir::AbstractString = GAP.sysinfo["DEFAULT_PKGDIR"])
```
lgoettgens, line 307:
```suggestion
                              pkgdir::AbstractString = sysinfo["DEFAULT_PKGDIR"])
```

src/packages.jl
```diff
@@ -343,7 +341,7 @@ For details, please refer to its documentation.
 """
 function remove(spec::String; interactive::Bool = true, quiet::Bool = false,
                               debug::Bool = false,
-                              pkgdir::AbstractString = DEFAULT_PKGDIR[])
+                              pkgdir::AbstractString = GAP.sysinfo["DEFAULT_PKGDIR"])
```
lgoettgens, line 344:
```suggestion
                              pkgdir::AbstractString = sysinfo["DEFAULT_PKGDIR"])
```

src/packages.jl
```diff
@@ -379,7 +377,7 @@ The info messages shown by this function can be suppressed by passing
 """
 function build(name::String; quiet::Bool = false,
                              debug::Bool = false,
-                             pkgdir::AbstractString = DEFAULT_PKGDIR[])
+                             pkgdir::AbstractString = GAP.sysinfo["DEFAULT_PKGDIR"])
```
lgoettgens, line 380:
```suggestion
                             pkgdir::AbstractString = sysinfo["DEFAULT_PKGDIR"])
```

src/packages.jl
fingolfin: Sure, but I'd like to keep this in sync with the docstring, to avoid confusion

lgoettgens:
> Specifically on my test machines it takes about 0.09 - 0.11 seconds. Which is not nothing, but still relatively little.

I compared de7bed99 (tip of this PR) to 0412fa6b (the master commit this branch is based on) w.r.t. `@time @time_imports using GAP` (after doing `] up`, waiting for precompilation and running multiple times to simulate a non-dev usage). There were no statistically significant changes between the two commits. This may be due to the removal of a pid-file usage that could counter the times you measured.

The only thing I am a little bit worried about is that some OSs may clear the `/tmp` directory after a set number of days. If this is done while a long running GAP.jl/OSCAR computation is still running, this may break.

fingolfin:
Oh, good point about stuff in `/tmp` or `/var/tmp` potentially being deleted by cron jobs. That's indeed a thing.

So how about instead we keep the scratch space and create the temp dirs in it (`mktempdir` has an option for that). The scratch space name then could stay relatively fixed. The tempdir names could consist of the current date&time in ISO8601 format followed by a random string (I am suggesting to put in time because that way it's trivial to spot any such dirs that should have been deleted but were not e.g. due to a crash. It might also help recognize the dirs for the "long running" jobs).

lgoettgens:
I don't understand what you would want to use mktempdir for? Shouldn't we let Scratch.jl create these directories using `get_scratch!`? This way, Scratch.jl would take care of garbage collecting old unused folders.

fingolfin:
I want a *different* directory for each julia process loading GAP.jl. Scratchspaces alone do not provide for that.

lgoettgens:
No, it does. If you call `get_scratch!(some_unique_string_that_contains_datetime)`, this creates a *unique* folder (as the key is unique). However, Scratch.jl will track that usage dates of these folders and gc them after some time

fingolfin:
@lgoettgens updated now as discussed on #1083

lgoettgens:
Why did you drop the changes in https://github.com/oscar-system/GAP.jl/compare/de7bed9931a62d2bcd003442268379f59359d274..d8459a9b7730af2650c491fe0b1e709f6b756d1c#diff-68a57defa4eef4087b4707a47026264bd208020e98af8639ab658853d0f1f35d (packages.jl)? It seems to me that this could be related to the failures here

fingolfin:
Nope the failure is caused by the use of our `gap.sh` wrapper, which now every time has a fresh GAP root. That messes with the build system of JuliaInterface: the configure script read `sysinfo.gap` and uses that to generate `Makefile` from `Makefile.in`. But also the path to the `sysinfo.gap` file is put into `Makefile`, so facilitate re-generation of everything if the GAP config changes. As a result, when we run `make clean` in `etc/ci_test.sh` it runs into an error because it can't fight `sysinfo.gap`.

Gotta think a bit about it...

fingolfin:
This is now reduced to generating an additional GAP root dir containing a `sysinfo.gap` and a `gac` executable. 

Perhaps time now to transform this:
- produce a function which takes a directory path and puts `sysinfo.gap` and `gac` (and whatever
- then only run this on a fresh temp dir just before invoking `GAP.Packages.install` (perhaps caching the resulting dir for the duration of the GAP.jl session)
- perhaps `create_gap_sh` could also make use of it, and produce a `gac` and `sysinfo.gap` which fit together with the produce `gap` shell script (towards resolving #1105)
- actually with that last bit we could circle back to the top: before calling `GAP.Packages.install` perhaps we should call this enhance `create_gap_sh` and make sure `PackageManger` "sees" its `sysinfo.gap` (and of course `gac`, `gap`)

fingolfin:
This would also hopefully allow us to deal with the three packages listed under "Missing Software: GAP" in #1122

fingolfin:
The stuff I just mentioned probably also would resolve #840

PR #1080, opened 2024-12-11
Load JuliaInterface before all other packages

fingolfin:
I think this is the right thing to do no matter what

### DIFF
--- a/gap/systemfile.g
+++ b/gap/systemfile.g
@@ -5,11 +5,10 @@
     # early enough for being required by other GAP packages,
     # and such that it is available when user files get read
     # via the GAP command line.
-    deps:= SHALLOW_COPY_OBJ( GAPInfo.Dependencies.NeededOtherPackages );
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
O.k., makes sense.
I think the idea of the old variant was that the other packages to be loaded do not depend on JuliaInterface.

lgoettgens (approved):
sounds sensible

fingolfin:
Yeah the old code worked fine because the other packages do not depend on JuliaInterface. But I don't think we put JI last on purpose, it was more a case of "just add it to the list" without reflecting on it further.

This way is better because e.g. a future version of `transgrp` might include optional Julia code for, say, managing additional data via artifacts, and then it should be loaded after `JuliaInterface` (well, OK, it probably would be using a package extension so the order is not as important, but at least logically it seems best to have the `JuliaInterface` functionality available as early as possible)

PR #1083, opened 2024-12-12
Use different scratchspace dir for every session

lgoettgens:
This implements my suggestion from https://github.com/oscar-system/GAP.jl/pull/1079#issuecomment-2537263244.

The other parts of #1079 (e.g. regarding the `sysinfo` handling) are not covered here and instead may be split off into their own PR or rebased or...


lgoettgens:
This currently does not create a new scratch folder for every session, but only every time precompilation runs. Do you have an idea on how to fix that?

lgoettgens:
I see you point. The only problem with `mktempdir` is that the cleanup does not work properly if the process does not end gracefully, e.g. due to a segmentation fault or an oom kill

fingolfin:
Exactly. That's why I wanted to put the dates in there, so that at least I can do something like `ls -d ~/.julia/scratchspaces/*/gap_*` to see "old" scratch spaces and clean them manually. 

But actually, perhaps we also use a scratch space name that involves the current *month*, so something like `gap-2024-12` -- then that scratch space is only used during that month and cleaned out roughly a month later. At least assuming a recompilation happens. If not it'll stay on this scratch space for a longer time. But that's till OK, it still means that any update to GAP.jl or a dependency is going to cause "old" scratch spaces to be orphaned and eventually recycled.

lgoettgens:
> But actually, perhaps we also use a scratch space name that involves the current _month_, so something like `gap-2024-12` -- then that scratch space is only used during that month and cleaned out roughly a month later. At least assuming a recompilation happens. If not it'll stay on this scratch space for a longer time. But that's till OK, it still means that any update to GAP.jl or a dependency is going to cause "old" scratch spaces to be orphaned and eventually recycled.

sounds good to me

lgoettgens:
closing in favor of https://github.com/oscar-system/GAP.jl/pull/1079

PR #1085, opened 2024-12-12
GAP: add and use ReplaceBinding helper

fingolfin:
This is a variant of Oscar's `ReplaceGapFunc` but I figure it is
be useful for other code.


lgoettgens (approved):
no idea about these low-lever GAP functions, but seems to work

ThomasBreuer (approved):
Yes, why not.

If the variable in question has no value (yet) then `VAL_GVAR` will run into an error, which is fine.

PR #1090, opened 2024-12-13
Fix dir names inside `gaproot/pkg/`

lgoettgens:
This changes it back from this dir containing dirs of the form `GAP_pkg_foobar` to just `foobar`. I suspect that this was an unintended side-effect from #1069.

Note: After this PR is merged, everyone with a dev version of GAP should remove all scratchspaces that have been used between #1069 was merged and this one here. Otherwise, GAP may get confused because all distro packages are available twice.

### DIFF
--- a/src/setup.jl
+++ b/src/setup.jl
@@ -199,8 +199,9 @@ function regenerate_gaproot()
         # create a `pkg` directory with symlinks to all the GAP packages artifacts
         mkpath(joinpath(gaproot_mutable, "pkg"))
         pkg_artifacts = filter(startswith("GAP_pkg_"), keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
-        for name in pkg_artifacts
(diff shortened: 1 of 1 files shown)


fingolfin:
Actually, I was aware of this and kept it as it is on purpose: because there is no real downside I am aware of, and the code simpler, and that means "easier to understand debug".

Moreover, the change you propose doesn't really make things much more natural. My understanding is this (please correct me if I am wrong):
- Current: `~/.julia/scratchspaces/ABC/gap_XXX-1.11/pkg/GAP_pkg_edim/EDIM-1.3.8/`
- With this patch: `~/.julia/scratchspaces/ABC/gap_XXX-1.11/pkg/edim/EDIM-1.3.8/`

lgoettgens:
I didn't notice that inside `pkgname` (resp. `GAP_pkg_pkgname`) there is another dir named `pkgname-version`. In this case this change does not make much sense, agreed.

PR #1094, opened 2024-12-16
Revamp GAP package overrides

fingolfin:
Previously we only checked for the package name to decide whether to
override the result of DirectoriesPackagePrograms. However, this is
problematic if e.g. we provide a kernel extension for version 1.0 of a
GAP package, but the user tries to load version 2.0 which may have an
incompatible kernel extension.

Therefore the new code is much stricter: we check whether the specific
version of the package for which `DirectoriesPackagePrograms` was called
is the one we provide via an artifact, and only then do we activate the
override.

### DIFF
--- a/pkg/JuliaInterface/read.g
+++ b/pkg/JuliaInterface/read.g
@@ -12,21 +12,31 @@ ReadPackage( "JuliaInterface", "gap/utils.gi");
 ReadPackage( "JuliaInterface", "gap/helpstring.g");
 ReadPackage( "JuliaInterface", "gap/juliahelp.g");
 
-# HACK: we need to set up the GAP package override system as early as possible.
-# Since JuliaInterface is loaded before all other packages, putting it here
-# ensures that
-BindGlobal("DirectoriesPackageProgramsOverrides", rec());
-
-BindGlobal("_DirectoriesPackageProgramsOriginal", DirectoriesPackagePrograms);
 MakeReadWriteGlobal("DirectoriesPackagePrograms");
 DirectoriesPackagePrograms := function(name)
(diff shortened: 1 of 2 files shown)


pkg/JuliaInterface/read.g
```diff
@@ -12,21 +12,31 @@ ReadPackage( "JuliaInterface", "gap/utils.gi");
...
 ReadPackage( "JuliaInterface", "gap/juliahelp.g");
 
-# HACK: we need to set up the GAP package override system as early as possible.
-# Since JuliaInterface is loaded before all other packages, putting it here
-# ensures that
-BindGlobal("DirectoriesPackageProgramsOverrides", rec());
-
-BindGlobal("_DirectoriesPackageProgramsOriginal", DirectoriesPackagePrograms);
 MakeReadWriteGlobal("DirectoriesPackagePrograms");
 DirectoriesPackagePrograms := function(name)
```
fingolfin, line 16: This could also use `ReplaceBinding` from PR #1085 once that is merged.

In any case, I think I should add a comment above this code to indicate why it is there and point to the other parts of it (in `src/GAP_pkg.jl`).

fingolfin: also the comment could indicate that most of the code below is copied from GAP's `DirectoriesPackagePrograms`

ThomasBreuer (approved):
Yes, this is how I understand the discussion we had on this topic.

pkg/JuliaInterface/read.g
lgoettgens: I still think this comment and usage of ReplaceBinding would be great

Issue #1092, opened 2024-12-13
Add `ExtendPackageDirectories` to GAP so we can get rid of mutable GAP root


fingolfin:
I.e. resurrect https://github.com/gap-system/gap/pull/933

@ThomasBreuer and me discussed today extensively why we even need the "mutable GAP root". A large part is so that we can collect the packages together.

A bunch of other things we have in there turn out to be not necessary anymore (I'll submit a PR to clean some out).

The main part left are the custom `sysinfo.gap` and `gac`, which are only relevant for compiling GAP packages (but then of course are crucial). But if we moved this out into a separate method, say `gaproot_for_building()`, then we could put this into a more permanent location (a genuine scratch space), and indeed only generate / update it when `GAP.Packages` functions get called. This would then also help resolve issue #840.

If we do that all that we may be able to eventually eliminate the mutable GAP root created during startup.

PR #1095, opened 2024-12-18
Use gap's `--packagedirs` instead of symlinking in gaproot

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/1092.

This needs https://github.com/gap-system/gap/pull/5873 to be available in the used GAP. Thus, I only expect the `CI with GAP / julia * / GAP master` jobs to succeed.


fingolfin:
This looks good, clean and simple.

But unfortunately it will have to wait some time, until `--packagedirs` arrives in our GAP... Normally this would have to wait for GAP 4.15, i.e. a couple months... but I guess we could also patch our GAP to already contain this code... But maybe we should first get 0.13.0 released.

lgoettgens:
Do we want to proceed with this before GAP 4.15 gets released, i.e. with patching https://github.com/gap-system/gap/pull/5873 into GAP_lib_jll? If yes, I can prepare the yggdrasil PR for that.

fingolfin:
@lgoettgens yes please

I also would like to release GAP 4.14.1 in the next 2-3 weeks (CC @ThomasBreuer) to get some GAP library & package bug fixes (but that's orthogonal)

Project.toml
```diff
@@ -61,7 +61,7 @@ BinaryWrappers = "0.1.3"
 Compat = "4.11"
 Downloads = "1.4.3"
 GAP_jll = "~400.1400.002"
-GAP_lib_jll = "~400.1400.002"
+GAP_lib_jll = "~400.1401.000"
```
lgoettgens, line 64:
```suggestion
GAP_lib_jll = "~400.1400.003"
```

lgoettgens:
I had to add a patch containing https://github.com/gap-system/gap/pull/5873 here as well, so we can patch gap `stable-4.14` in the `CI with GAP` CI job.

fingolfin (approved):
Looks good to me. Thanks a ton!

PR #1096, opened 2024-12-18
Use JLL wrappers for caratinterface, zeromqinterface

fingolfin:


lgoettgens:
These two jll wrappers should also be mentioned in the changelog

PR #1099, opened 2024-12-18
Enable GAP package Singular

fingolfin:


fingolfin:
We should first merge #1100 then I can easily adapt this PR

pkg/JuliaInterface/read.g
```diff
@@ -40,3 +40,6 @@ DirectoriesPackagePrograms := function(name)
     return [ Directory( override ) ];
 end;
 MakeReadOnlyGlobal("DirectoriesPackagePrograms");
+
+# ensure GAP package 'singular' sees Singular executablef rom Singular_jll
```
lgoettgens, line 44:
```suggestion
# ensure GAP package 'singular' sees Singular executable from Singular_jll
```

fingolfin:
@ThomasBreuer so this sets up the GAP package `singular` to use `Singular_jll` instead of doing it in Oscar.jl. On the long run we maybe want to change this, but for now I think doing it here is best because then we can run the corresponding package distro tests (see PR #1067)

PR #1106, opened 2024-12-20
Test `GAP.prompt()` and tab completion via `expect`

### LINKED ISSUE #460: Add automated CI tests for the GAP.prompt()
Right now I am not aware of a "good" way to test GAP.prompt(); but I'd really like to have one, so that when an issue like #458 pops up, then the fix (here in #459) can also include a test that verifies the issue is gone.

One way to go about this might be to use [pexpect](https://pexpect.readthedocs.io/en/stable/index.html) which can be used to automate interactive command line utilities. (Of course if somebody knows a pure-Julia way to do the same, I'd also be happy to use that). This can also be handy for other kinds of tests, as it allows to emulate what an actual user would experience very closely.

I think we need to do roughly this:
1. add `pip install pexpect` into `.travis.yml`
2. add a script to `etc` which performs the required test using `pexpect`
3. call that script from `.travis.yml`

fingolfin:
Resolves #460

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,6 +49,7 @@ jobs:
           # be able to properly deal with PRs / merges
           fetch-depth: 2
       - name: "Install extra dependencies"
+        if: runner.os == 'Linux'
         run: sudo apt-get install expect
       - name: "Set up Julia"
         uses: julia-actions/setup-julia@v2
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -48,6 +48,8 @@ jobs:
           # For Codecov, we must also fetch the parent of the HEAD commit to
           # be able to properly deal with PRs / merges
           fetch-depth: 2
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -48,6 +48,8 @@ jobs:
           # For Codecov, we must also fetch the parent of the HEAD commit to
           # be able to properly deal with PRs / merges
           fetch-depth: 2
+      - name: "Install extra dependencies"
+        run: sudo apt-get install expect
```
fingolfin, line 52: Gotta adapt this for macos 😆

lgoettgens: Is `expect` pre-installed on the macOS runners?

lgoettgens (approved):

.github/workflows/CI.yml
fingolfin: Yes, in fact it ships with macOS (resp. with Xcode)

PR #1112, opened 2025-01-06
Optimize conversion of GAP strings to Julia

fingolfin:
... be it to Julia strings, symbols, or `Vector{UInt8}`

Multiple things were optimized:

- For conversion to `Vector{UInt8}` we used to do call `deepcopy` on the output of `unsafe_wrap`, but `copy` is sufficient and avoids allocating an `IdDict`; but even better is to create a `Vector{UInt8}` of the right size and then use `unsafe_copyto!`.
- We also called `deepcopy` on the output of `unsafe_string`, but that was completely unnecessary as `unsafe_string` actually copies the data passed to it (unlike `unsafe_wrap`).
- A new internal helper `UNSAFE_CSTR_STRING` avoid using `ccall` to get the data inside a GAP string
- For conversion to symbols we now `ccall` the Julia kernel function `jl_symbol_n`
- Writing `Vector{UInt8}(s)` was slower than `gap_to_julia(Vector{UInt8}, s)` because the former invoked `gap_to_julia(Vector{UInt8}, s; recursive=true)` and this then went into a generic conversion method; by extending the method for `gap_to_julia(Vector{UInt8}, s)` to also take (and then ignore) the `recursive` keyword argument, this overhead (and the creation of yet another dictionary) could be avoided.

As a result we can replace the kernel function `JuliaSymbol` by `Julia.Symbol` as that is at least as efficient now.

Benchmarks using `ChairMarks.jl`, before this patch:

    julia> s = GAP.Obj("abc")
    GAP: "abc"

    julia> @b String($s)
    94.880 ns (4 allocs: 368 bytes)

    julia> @b GAP.gap_to_julia(String, $s)
    94.531 ns (4 allocs: 368 bytes)

    julia> @b Symbol($s)
    163.811 ns (4 allocs: 368 bytes)

    julia> @b GAP.gap_to_julia(Symbol, $s)
    161.397 ns (4 allocs: 368 bytes)

    julia> @b Vector{UInt8}($s)
    1.148 μs (10 allocs: 480 bytes)

    julia> @b GAP.gap_to_julia(Vector{UInt8}, $s)
    142.219 ns (7 allocs: 480 bytes)

After this patch:

    julia> s = GAP.Obj("abc")
    GAP: "abc"

    julia> @b String($s)
    14.297 ns (1 allocs: 24 bytes)

    julia> @b GAP.gap_to_julia(String, $s)
    14.310 ns (1 allocs: 24 bytes)

    julia> @b Symbol($s)
    72.098 ns

    julia> @b GAP.gap_to_julia(Symbol, $s)
    72.044 ns

    julia> @b Vector{UInt8}($s)
    21.397 ns (2 allocs: 64 bytes)
    
    julia> @b GAP.gap_to_julia(Vector{UInt8}, $s)
    21.409 ns (2 allocs: 64 bytes)

### DIFF
--- a/pkg/JuliaInterface/gap/adapter.gi
+++ b/pkg/JuliaInterface/gap/adapter.gi
@@ -65,7 +65,7 @@ BindGlobal("_JL_RNAM_TO_JULIA_SYMBOL", function(rnam)
     if ISB_REC(_JL_RNAM_TO_JULIA_SYMBOL_CACHE, rnam) then
         symbol := ELM_REC(_JL_RNAM_TO_JULIA_SYMBOL_CACHE, rnam);
     else;
-        symbol := JuliaSymbol( NameRNam( rnam ) );
+        symbol := Julia.Symbol( NameRNam( rnam ) );
         ASS_REC(_JL_RNAM_TO_JULIA_SYMBOL_CACHE, rnam, symbol);
     fi;
     return symbol;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,6 +39,7 @@
   - simpcomp
   - singular
   - zeromqinterface
+- Optimize conversion of GAP strings to Julia strings, symbols or `Vector{UInt8}`
 
 ## Version 0.12.3 (released 2025-01-01)
 
--- a/pkg/JuliaExperimental/gap/numfield.g
+++ b/pkg/JuliaExperimental/gap/numfield.g
@@ -809,7 +809,7 @@ InstallMethod( Characteristic,
       # We need this for matrix groups over residue class rings.
       # Nemo does not support it.
       return JuliaToGAP( IsInt,
-                 Julia.Base.getfield( R, JuliaSymbol( "n" ) ) );
+                 Julia.Base.getfield( R, Julia.Symbol( "n" ) ) );
     else
       return JuliaToGAP( IsInt, Julia.Nemo.characteristic( R ) );
     fi;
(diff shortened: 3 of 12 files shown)


pkg/JuliaInterface/gap/adapter.gi
```diff
@@ -65,7 +65,7 @@ BindGlobal("_JL_RNAM_TO_JULIA_SYMBOL", function(rnam)
     if ISB_REC(_JL_RNAM_TO_JULIA_SYMBOL_CACHE, rnam) then
         symbol := ELM_REC(_JL_RNAM_TO_JULIA_SYMBOL_CACHE, rnam);
     else;
-        symbol := JuliaSymbol( NameRNam( rnam ) );
+        symbol := Julia.Symbol( NameRNam( rnam ) );  # TODO: use rnam -> symbol cache
```
ThomasBreuer, line 68: Concerning the caching idea:
Isn't this the `if` case, using the record `_JL_RNAM_TO_JULIA_SYMBOL_CACHE`?
Or would a Julia dictionary be more appropriate?

ThomasBreuer (approved):
- Yes, the previous idea to delegate between `gap_to_julia` methods was not good.
- Luckily, `JuliaSymbol` was not documented.
- Just one comment about caching.

pkg/JuliaInterface/gap/adapter.gi
fingolfin: Yes precisely -- those comments slipped in by accident (I made them during development then realized they are pointless and *thought* I had removed them... but clearly not *sigh*)

PR #1113, opened 2025-01-06
Remove call to `Base.GC.gc(true)` in `__init__`

fingolfin:
This was a workaround for weird GC related errors I was seeing for a while. My guess is that these were ramifications of the argv issue recently fixed by adding `_saved_argv` (see PR #1110).

We'll see if CI passes. If not, time to investigate this more.


lgoettgens (approved):
Let's run CI once again and merge if still everything green

PR #1115, opened 2025-01-07
Ensure all gap_to_julia method handle 'recursive' kwarg

fingolfin:
... instead of relying on a dodgy catch-all method which handles that kwarg

The required changes for Oscar.jl are in https://github.com/oscar-system/Oscar.jl/pull/4419 

Also update CHANGES.md to point out breaking changes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,12 +2,17 @@
 
 ## Version 0.13.0-DEV (released YYYY-MM-DD)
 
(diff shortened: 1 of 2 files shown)


ThomasBreuer (approved):
wYes, no all `gap_to_julia` methods support `recursive`

lgoettgens (approved):

PR #1121, opened 2025-01-09
Version 0.13.0

### LINKED ISSUE #1114: Releasing GAP.jl 0.13.0
What do we need for releasing GAP.jl 0.13.0 ?
- [x] merge PR #1067
- [x] try to use Oscar.jl with new GAP.jl to look out for serious issues we need to address first?
      - a full PR adapting Oscar.jl can come later, but it would be nice to know it is possible, instead of having to first release 0.13.1. Then again, that won't kill us either

Anything else, @ThomasBreuer @lgoettgens ?

We could also go for "only load a minimal set of packages" / "no packages" (that was my original plan). But that's another biggish change, so perhaps it is better to not do this now, but try it for 0.14.0 which could come quickly after 0.13.0 ?

Other things that would be nice but are not necessary from my POV:
- PR #1082 -- someone feeling like adding `GAP.versioninfo()` ?
- PR #1112 -- optimizations are always nice
  - if we merge this we must update `GAP_pkg_juliainterface_jll`:
    - [x] https://github.com/JuliaPackaging/Yggdrasil/pull/10213
    - [x] https://github.com/oscar-system/GAP.jl/pull/1116

fingolfin:
Resolves #1114


lgoettgens:
Do you wanna fix https://github.com/oscar-system/GAP.jl/issues/1114#issuecomment-2581146637 first?

ThomasBreuer (commented):
I am confused.
The description says that #1114 gets resolved, but #1114 describes some (new?) issue?

lgoettgens:
> I am confused. The description says that #1114 gets resolved, but #1114 describes some (new?) issue?

#1114 is the issue "release 0.13", and in a comment there I posted the errors that the Oscar CI has with the current GAP.jl master. If you click on the link, you get redirected to the stack trace

CHANGES.md
```diff
@@ -1,15 +1,17 @@
 # Changes in GAP.jl
 
-## Version 0.13.0-DEV (released YYYY-MM-DD)
+## Version 0.13.0 (released 2025-01-09)
```
lgoettgens, line 3: Reminder: Needs to be updated again

Project.toml
```diff
@@ -60,8 +60,8 @@ Artifacts = "1.6"
 BinaryWrappers = "0.1.3"
 Compat = "4.11"
 Downloads = "1.4.3"
-GAP_jll = "~400.1400.000"
-GAP_lib_jll = "~400.1400.000"
+GAP_jll = "~400.1400.002"
+GAP_lib_jll = "~400.1400.001"
```
lgoettgens, line 64:
```suggestion
GAP_lib_jll = "~400.1400.002"
```

lgoettgens: Require backport of https://github.com/gap-system/gap/pull/5903

lgoettgens (approved):
Apart from the release date this looks good to me. The OscarCI with the current GAP.jl master is happy as well (cf https://github.com/oscar-system/Oscar.jl/actions/runs/12725036717?pr=4421)

Issue #1122, opened 2025-01-13
Enable more packages in the GAP distro CI tests


lgoettgens:
https://github.com/oscar-system/GAP.jl/pull/1067 added CI tests for all packages in the GAP package distribution (ref. https://github.com/oscar-system/GAP.jl/issues/1065).
However, many packages are still excluded due to various reasons.

The current list and some notes on why they are excluded can be found in https://github.com/oscar-system/GAP.jl/blob/master/.github/workflows/CI-distro.yml#L63-L86.

- [ ] `alnuth`
- [ ] `atlasrep`
- [ ] `autodoc`
- [ ] `example`
- [ ] `examplesforhomalg`
- [ ] `guarana`
- [ ] `guava`
- [ ] `hap`
- [ ] `hapcryst`
- [ ] `help`
- [ ] `io`
- [ ] `itc`
- [ ] `localizeringforhomalg`
- [ ] `normalizinterface`
- [ ] `packagemanager`
- [ ] `polenta`
- [ ] `polycyclic`
- [ ] `polymaking`
- [ ] `profiling`
- [ ] `radiroot`
- [ ] `ringsforhomalg`
- [ ] `semigroups`
- [ ] `xgap`

fingolfin:
To get the GAP package 'polymaking' working (and hence `hap`, `hapcryst`) we need access to polymake. This has a JLL and we could use `BinaryWrappers.jl`. However, this is complicated by the fact that the `polymake` executable is just a perl script, which begins with `#!/usr/bin/env perl`. Hence the following doesn't work:

    GAP.Globals.POLYMAKE_COMMAND = GapObj(joinpath(polymake_binpath, "polymake"))

... as this ends up trying to use a system install version of Perl (which may not even exist). So we need to use the perl from our `Perl_jll`. E.g this works to launch polymake:

    import Oscar.Polymake: polymake_jll
    run(addenv(`$(polymake_jll.Perl_jll.perl()) $(polymake_jll.polymake_path)`,
               polymake_jll.JLLWrappers.LIBPATH_env=>polymake_jll.LIBPATH[]))

But GAP needs a path to an executable and we can't tell it to set environment variables. So we either need to generate our own wrapper shell script; or someone needs to generalize BinaryWrappers to teach it to generate wrappers that set up environment variables based on one JLL (here: `polymake_jll`), but executes something from another (here: perl from `Perl_jll`).

[...]

PR #1124, opened 2025-01-15
CI: enable packages using pari

fingolfin:
Progress towards https://github.com/oscar-system/GAP.jl/issues/1122.

### DIFF
--- a/.github/workflows/CI-distro.yml
+++ b/.github/workflows/CI-distro.yml
@@ -91,6 +91,7 @@ jobs:
           version: ${{ matrix.julia-version }}
           arch: ${{ matrix.julia-arch }}
       - name: 'Install PARI/GP'  # for alnuth, and thus guarana, polenta, radiroot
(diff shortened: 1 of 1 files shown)


.github/workflows/CI-distro.yml
```diff
@@ -94,6 +90,10 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
           arch: ${{ matrix.julia-arch }}
+      - name: 'Install PARI/GP'  # for alnuth, and thus guarana, polenta, radiroot
```
lgoettgens, line 93:
```suggestion
      - name: 'Install PARI/GP'  # for alnuth, and thus guarana, polenta, radiroot
        if: matrix.gap-package == 'alnuth' || matrix.gap-package == 'guarana' || matrix.gap-package == 'polenta' || matrix.gap-package == 'radiroot'
```
we could install it only conditionally like this, but no strong preference

lgoettgens (approved):

Issue #1053, opened 2024-09-24
Replace `Julia.GAP` by a cleaner solution?


fingolfin:
Right now we use a hack to make sure `Julia.GAP` always refers to the `GAP` julia module, even if it was not loaded into Julia's `Main` module.

This requires some hacky code. It also means that `Main.GAP` in Julia and `Julia.GAP` in GAP can reference different objects (mind you, that's not a major concern as it is easy to avoid for a user, but it's still annoying).

So perhaps we can provide an additional alternative means to access it (and then phase out use of `Julia.GAP` over time. E.g. we could add a global variable `GAP_jl` which contains the GAP module. Suggestions for better names welcome. (I would not call it `GAP`, though, as I believe this would be very confusing for GAP users).

If we went with `GAP_jl`, then for symmetry in Oscar we could also provide `Oscar_jl` as an alias for the `Oscar` global variable.

PR #1126, opened 2025-01-24
Introduce global gap variable `GAP_jl`

lgoettgens:
Initial step towards https://github.com/oscar-system/GAP.jl/issues/1053.

I would only call it resolved, once the special case for `Julia.GAP` is indeed removed.


fingolfin (approved):
Thinking this through: the immediate name for this object that comes to mind is just `GAP`, on the top level. But that somehow "feels" wrong, or at least pretentious for us to claim that name. It feels as if a global variable with this name should either not exist, or "belong" to the GAP library. 

So `GAP_jl` then seems rather natural. I can't say I love it (no offense), but as my language makes clear, everything I am saying here is about "feeling" and so extremely subjective anyway.

Of course it somewhat breaks with the usual GAP naming conventions that suggest `CamelCase` for globals, but `GAPJl` or `GAPJL` don't seem seem better, and `GAPJulia` is worse.

I am pretty sure I'll get used to `GAP_jl` if we use it *shrug*.

If we do this, then we should also consider renaming the global `Oscar` object in OscarInterface to `Oscar_jl` (or rather: have both names, at least for a while, but move everything to use `Oscar_jl`).


All that said, I'd really like to hear from @ThomasBreuer, perhaps he sees something I am not or has some other ideas and suggestions.

ThomasBreuer:
I would expect that the variable name `GAP` in a GAP session describes some important object of this GAP session, for example we could have called the `GAPInfo` record just `GAP`.
Inside a GAP session, using the name `GAP` for a Julia module would indeed be irritating.

Since we really need a variable for this module, I think `GAP_jl` is fine.

I found `Julia.GAP` very elegant: It expresses that it is something in the Julia world, and relative to that, `GAP` is exactly what one means.

I think there is no naming problem in the other direction: In Julia, `GAP` stands for the Julia module `GAP`, and `GAP.Globals` stands for the objects on the `GAP` side --the only problem is the ambiguity of `GAP` in sentences like this.

ThomasBreuer (approved):
see the comment which I just made

(I had thought that I had approved this pull request on Friday.)

PR #1127, opened 2025-01-24
Simplify doc building by removing support for Documenter pre-1.0

lgoettgens:
We require Documenter.jl of at least version 1.0 already since Nov 2023. Thus I believe this code path is no longer needed.

The support for a `mkdocs.yml` file was removed in Documenter.jl in https://github.com/JuliaDocs/Documenter.jl/pull/1826, which was also released as part of Documenter.jl 1.0. So we can remove this file as well.

Preview: https://oscar-system.github.io/GAP.jl/previews/PR1127/

~~Currently draft, as this contains https://github.com/oscar-system/GAP.jl/pull/1128. Will be un-drafted once rebased.~~


lgoettgens:
The nightly failures are all https://github.com/oscar-system/GAP.jl/issues/1132

fingolfin (approved):

PR #1128, opened 2025-01-24
Add docs preview for PRs

lgoettgens:
This is just copy-pasted from the Oscar repo.

It would be great to have something like this to make sure that e.g. https://github.com/oscar-system/GAP.jl/pull/1127 work as expected and do not break the documentation completely, before merging such PRs.


lgoettgens:
The preview works. For this PR, it can be accessed via https://oscar-system.github.io/GAP.jl/previews/PR1128/.

fingolfin (approved):

Issue #964, opened 2024-02-02
problem with local builds


ThomasBreuer:
The CI tests take care of copying certain files in `pkg/JuliaInterface/doc/` to `docs/src/assets/html/JuliaInterface/`.
These files are needed by Documenter.jl for the creation of cross-references to the JuliaInterface manual after the JuliaInterface manual has been built.

When one builds the GAP.jl documentation locally then there is no such automatism yet. If all cross-referenced files exist in `docs/src/assets/html/JuliaInterface/` then everything is fine, it not (for example because new cross-references have been added) then Documenter.jl complains (and then it is not obvious from the error messages how one should fix this problem).

Logically, `docs/make.jl` would be the right place for a command that copies the files.
Should the JuliaInterface manual get processed before that, or would that be too involved?

fingolfin:
We can add the copying code to `docs/make.jl` but then we should indeed also run `makedoc.g` in JuliaInterface. And then the whole "Build GAP manual" step in `.github/workflows/CI.yml` could go.

It shouldn't be too hard to do that. If anyone is interested in doing it, please go ahead.

PR #1130, opened 2025-01-24
Move JuliaInterface manual building to `docs/make.jl`

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/964.

~~This PR depends on and thus includes #1127, #1128, and #1129. I'll drop the respective commits once they are merged. (Draft until this is done)~~

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -69,8 +69,8 @@ function compute_GAP_URL(url::String)
 end
 
 
-# The code was adapted from `expand_citations`
-# in DocumenterCitations.jl/src/citations.jl.
+# The code was adapted from `expand_citation`
+# in DocumenterCitations.jl/src/expand_citations.jl.
 Document = Documenter.Document
 
 # Execute our step before the computation of cross-references (3.0)


docs/make.jl
```diff
@@ -42,104 +68,71 @@ function compute_GAP_URL(url::String)
...
-        for (docstr, meta) in zip(node.element.mdasts, node.element.metas)
-          compute_external_reference(doc, page, docstr)
-        end
-      elseif node.element isa Documenter.MarkdownAST.Link
-        compute_external_reference(node, page.globals.meta, page, doc)
-      end
-    end
-  end
+# The code was adapted from `expand_citations`
+# in DocumenterCitations.jl/src/citations.jl.
```
ThomasBreuer, line 73: There is no longer a file `src/citations.jl`, and there is no function `expand_citations` anymore in DocumenterCitations.jl.

ThomasBreuer (approved):
Looking at the many changes, I am surprised that the GAP.jl manual was not completely broken before.

docs/make.jl
fingolfin: This is part of existing code that was just moved around (and that bit was in another PR and thus is already merged). I believe you wrote it, @ThomasBreuer ? If this is outdated, can you please open a PR updating it?

docs/make.jl
```diff
@@ -42,8 +69,8 @@ function compute_GAP_URL(url::String)
 end
 
 
-# The code was adapted from `expand_citations`
-# in DocumenterCitations.jl/src/citations.jl.
+# The code was adapted from `expand_citation`
+# in DocumenterCitations.jl/src/expand_citations.jl.
```
fingolfin, line 73: Ah maybe @lgoettgens already updated it here (CC @ThomasBreuer )

fingolfin (approved):

fingolfin:
> Looking at the many changes, I am surprised that the GAP.jl manual was not completely broken before.

I don't understand this remark? As far as I can tell, nothing was broken, and there is no claim here that anything is being fixed?

This PR, and the three other documentation related PRs that were part of it initially, mostly move some things around, untangle bits, add a few small settings (which is not meant to take away from this nice work @lgoettgens has done here, I really appreciate it). Now the final bit is clearly visible by being on its own, and it mainly converts some shell script that was stuck in a CI.yml file to (better) Julia code. That makes it now trivial to buld the full docs locally, and in exactly the same way as CI does. Nice!

docs/make.jl
lgoettgens: Already did that in [`3326f7b` (#1130)](https://github.com/oscar-system/GAP.jl/pull/1130/commits/3326f7b7ac3412829bda74d0aac30d62ce8ad193)

ThomasBreuer: (Sorry, this was a misunderstanding:
Yes, I had added this comment when I added the code. With the current changes of the code, the comment was no longer relevant; this would hold even if the claimed reference would still exist.)

ThomasBreuer:
> > Looking at the many changes, I am surprised that the GAP.jl manual was not completely broken before.
> 
> I don't understand this remark? As far as I can tell, nothing was broken, and there is no claim here that anything is being fixed?

I meant the code from #1127 that deals with external references. This code got rewritten, not just moved around. I assume that the new code is the current state of the art (how one should do it with the current Documenter and DocumenterCitattions), and I was surprised that the old code did still work.

lgoettgens:
> > > Looking at the many changes, I am surprised that the GAP.jl manual was not completely broken before.
> > 
> > 
> > I don't understand this remark? As far as I can tell, nothing was broken, and there is no claim here that anything is being fixed?
> 
> I meant the code from #1127 that deals with external references. This code got rewritten, not just moved around. I assume that the new code is the current state of the art (how one should do it with the current Documenter and DocumenterCitattions), and I was surprised that the old code did still work.

The only thing I did in #1127 was to remove the branch for Documenter 0.27. This is easier to see with whitespace changes hidden, i.e. https://github.com/oscar-system/GAP.jl/pull/1127/files?w=1

fingolfin (approved):

PR #1139, opened 2025-02-02
Update GAP_pkg_juliainterface_jll

fingolfin:


lgoettgens (approved):
The `ctbllib` CI job has a segfault, but that should be unrelated to this change.

fingolfin:
@lgoettgens what confuses me are the failures in Julia nightly which again report worldage failures. I wonder if that means something went wrong compiling the JLL that caused it to no pick up `JULIA_VERSION_MINOR` and `JULIA_VERSION_MAJOR` correctly... huh

fingolfin:
I can reproduce this locally so it is not something weird in the CI.

But I also used a disassembler to look at the `JuliaInterface.so` in the JLL -- and thus verified that its `Func_JuliaGetGlobalVariableByModule` really does contain the new code which updates the world age. 

However comparing it with a locally built `JuliaInterface.so` it appears the offset to world age is different. Conclusion: `libjulia_jll` needs to be updated with a recent Julia master snapshot, then we can rebuild `GAP_pkg_juliainterface_jll` and hopefully it'll work as desired.

lgoettgens:
julia nightly now refers to 1.13-DEV, for which we don't have a libjulia yet (x-ref https://github.com/JuliaPackaging/Yggdrasil/pull/10429). After https://github.com/oscar-system/GAP.jl/pull/1141 is merged, we can at least see CI running on 1.12-DEV again.

lgoettgens:
The 1.12 tests only show warnings about `docmeta`, i.e. https://github.com/JuliaDocs/Documenter.jl/issues/2621.

I am looking into the 1.13 failure, this probably has to do with Aqua.jl.

lgoettgens (approved):
This looks good now.
Two things to note:
- The name of the 1.11 CI job seems to have changed. @fingolfin could you please update the list of required CI jobs to reflect that?
- There are still some errors in `nightly` everytime we try to access singular (ref https://github.com/oscar-system/GAP.jl/actions/runs/13178758178/job/36784187046?pr=1139#step:11:314). This will probably fix itself once `libsingular_julia_jll` gets rebuild with julia 1.13 support.

PR #1141, opened 2025-02-04
Run CI both against 1.12 and 1.13 nightly

lgoettgens:


lgoettgens:
nightly (1.13) is expected to fail until https://github.com/JuliaPackaging/Yggdrasil/pull/10429 is merged and everything dependent is rebuilt. 1.12-nightly may fail as well until https://github.com/oscar-system/GAP.jl/pull/1139 is merged.

Issue #1146, opened 2025-02-06
The function `CSTR_STRING` exists, but no method is defined for this combination of argument types.


lgoettgens:
new failure on julia nightly (1.13), first seen on https://github.com/oscar-system/GAP.jl/actions/runs/13182335604/job/36795983847#step:8:259 with julia 1.13.0-DEV.11, Commit https://github.com/JuliaLang/julia/commit/6f992ab0fc5.

```julia
basics: Error During Test at /home/runner/work/GAP.jl/GAP.jl/test/basics.jl:24
  Test threw exception
  Expression: string(GapObj("x")) == "x"
  MethodError: no method matching CSTR_STRING(::String)
  The function `CSTR_STRING` exists, but no method is defined for this combination of argument types.
  
  Closest candidates are:
    CSTR_STRING(::GapObj)
     @ GAP ~/work/GAP.jl/GAP.jl/src/ccalls.jl:207
  
  Stacktrace:
   [1] string(obj::GapObj)
     @ GAP ~/work/GAP.jl/GAP.jl/src/adapter.jl:30
   [2] top-level scope
     @ ~/work/GAP.jl/GAP.jl/test/basics.jl:2
   [3] macro expansion
     @ /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.13/Test/src/Test.jl:1771 [inlined]
   [4] macro expansion
     @ ~/work/GAP.jl/GAP.jl/test/basics.jl:24 [inlined]
   [5] macro expansion
     @ /opt/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.13/Test/src/Test.jl:676 [inlined]
Test Summary: | Pass  Error  Total  Time
basics        |   15      1     16  2.9s
RNG of the outermost testset: Random.Xoshiro(0xa32af379724aa8f3, 0x0782dfb5c86325bb, 0xf92fa2c162bdc01b, 0xc5c383ef86abf180, 0xfffa5bee67dc752b)
ERROR: LoadError: Some tests did not pass: 15 passed, 0 failed, 1 errored, 0 broken.
in expression starting at /home/runner/work/GAP.jl/GAP.jl/test/basics.jl:1
in expression starting at /home/runner/work/GAP.jl/GAP.jl/test/runtests.jl:8
Package GAP errored during testing
```

lgoettgens:
I found the following difference:
```julia
[...]

PR #1147, opened 2025-02-06
Work around resolvedness underspecification

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/1146.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -416,6 +416,7 @@ macro wrap(ex)
 
     return esc(MacroTools.@qq begin
        @eval const $newsym = Ref{GapObj}()
+       global $name;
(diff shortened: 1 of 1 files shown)


src/macros.jl
```diff
@@ -416,6 +416,7 @@ macro wrap(ex)
 
     return esc(MacroTools.@qq begin
        @eval const $newsym = Ref{GapObj}()
+       global $name;
```
fingolfin, line 419: Any deep reason why a semicolon is needed? If yes it should be documented in a comment. If not, let's drop it
```suggestion
       global $name
```

fingolfin (approved):

PR #1150, opened 2025-02-07
Move code out of regenerate_gaproot()

fingolfin:
Computing the list of GAP roots and the package dir now are
separate from regenerating sysinfo.gap

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -27,6 +27,7 @@ import Artifacts: find_artifacts_toml, @artifact_str
 import TOML
 
 import GAP_jll: GAP_jll, libgap
+import GAP_lib_jll
 
 GAP_jll.is_available() ||
    error("""This platform or julia version is currently not supported by GAP:
@@ -63,6 +64,7 @@ import Random
 # the startup time a little bit
 const sysinfo = Setup.regenerate_gaproot()
 
+const GAP_VERSION = VersionNumber(sysinfo["GAP_VERSION"])
 
 include("types.jl")
 
@@ -253,13 +255,22 @@ function __init__()
 
     real_JuliaInterface_path[] = Setup.locate_JuliaInterface_so(sysinfo)
 
-    gaproots = sysinfo["GAPROOTS"]
-    cmdline_options = ["", "-l", gaproots]
+    roots = [
+            # GAP root with custom sysinfo.gap
+            Setup.gaproot(),
+            # GAP root for the the actual GAP library, from GAP_lib_jll
+            abspath(GAP_lib_jll.find_artifact_dir(), "share", "gap"),
+            # GAP root into which PackageManager installs packages by default
+            Packages.gap_packages_rootdir(),
+            ]
+    cmdline_options = ["", "-l", join(roots, ";")]
 
     # tell GAP about all artifacts that contain GAP packages
     pkg_artifacts = filter(startswith("GAP_pkg_"), keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
-    pkgdirs = join((realpath(@artifact_str(name)) for name in pkg_artifacts), ';')
-    append!(cmdline_options, ["--packagedirs", pkgdirs])
+    pkgdirs = [realpath(@artifact_str(name)) for name in pkg_artifacts]
+    push!(pkgdirs, abspath(@__DIR__, "..", "pkg", "JuliaInterface"))
+    push!(pkgdirs, abspath(@__DIR__, "..", "pkg", "JuliaExperimental"))
+    push!(cmdline_options, "--packagedirs", join(pkgdirs, ';'))
 
     if isdefined(Main, :__GAP_ARGS__)
         # we were started via gap.sh, handle user command line arguments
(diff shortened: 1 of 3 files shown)


src/GAP.jl
```diff
@@ -253,8 +255,16 @@ function __init__()
...
     real_JuliaInterface_path[] = Setup.locate_JuliaInterface_so(sysinfo)
 
-    gaproots = sysinfo["GAPROOTS"]
-    cmdline_options = ["", "-l", gaproots]
+    gaproot_gapjl = abspath(@__DIR__, "..")
+    gap_lib_dir = abspath(GAP_lib_jll.find_artifact_dir(), "share", "gap")
+    mkpath(Packages.DEFAULT_PKGDIR[])
+    roots = [
+            gaproot_gapjl,      # for JuliaInterface and JuliaExperimental
+            Packages.DEFAULT_PKGDIR[],
```
ThomasBreuer, line 263: `Packages.DEFAULT_PKGDIR[]` is a path with suffix `/pkg`. The corresponding root directory is the path without this suffix.

lgoettgens:
```suggestion
            abspath(Packages.DEFAULT_PKGDIR[], ".."),
```

lgoettgens: Alternatively, we could pass the `DEFAULT_PKGDIR[]` via `--packagedirs` instead of `-l`. The same would be possible for `gaproot_gapjl` as well, since this is only included in the list for providing the `pkg/` subdir.

ThomasBreuer: Yes, eventually the packages should be handled via package directories, not root directories.
(I was not sure whether the current GAP.jl must still support the "old" setup where we cannot assume package directories independent of root directories.)

src/packages.jl
```diff
@@ -5,9 +5,11 @@ import Downloads
 import Pidfile
 import Pkg
 import ...GAP: disable_error_handler, Globals, GapObj, replace_global!, RNamObj, sysinfo, Wrappers
-import ...GAP: gap_pkg_jlls, Compat, Setup
+import ...GAP: gap_pkg_jlls, Compat, Setup, GAP_VERSION, GAP_jll, GAP_lib_jll
 
-const DEFAULT_PKGDIR = Ref{String}()
+gap_pkgdir() = joinpath(Base.DEPOT_PATH[1], "gaproot", "v$(GAP_VERSION.major).$(GAP_VERSION.minor)")
```
lgoettgens, line 10: I find it a bit confusing that this function is called `pkgdir` although it contains the path to a GAP root path and *not* to a GAP packagedir.

src/GAP.jl
lgoettgens: Since GAP_lib_jll 400.1400.003 the PR to gap is included, so we can assume this.

fingolfin:
Sorry forgot to mark this as draft. Really no point in commenting on it right now

fingolfin:
OK this is now ready for reviews :-)

src/GAP.jl
```diff
@@ -253,13 +255,22 @@ function __init__()
...
+            ]
+    cmdline_options = ["", "-l", join(roots, ";")]
 
     # tell GAP about all artifacts that contain GAP packages
     pkg_artifacts = filter(startswith("GAP_pkg_"), keys(TOML.parsefile(find_artifacts_toml(@__FILE__))))
-    pkgdirs = join((realpath(@artifact_str(name)) for name in pkg_artifacts), ';')
-    append!(cmdline_options, ["--packagedirs", pkgdirs])
+    pkgdirs = [realpath(@artifact_str(name)) for name in pkg_artifacts]
+    push!(pkgdirs, abspath(@__DIR__, "..", "pkg", "JuliaInterface"))
+    push!(pkgdirs, abspath(@__DIR__, "..", "pkg", "JuliaExperimental"))
```
lgoettgens, line 272:
```suggestion
    push!(pkgdirs, abspath(@__DIR__, "..", "pkg"))
```
alternatively, you could do this, but there is no difference whatsoever as no more packages are put in `../pkg/`

src/packages.jl
lgoettgens: This is still the same issue for me

lgoettgens (approved):

src/packages.jl
fingolfin: changed it, please have another look

lgoettgens: The new name is less confusing, thanks!

lgoettgens:
What I just noticed: the code now no longer sets `sysinfo["GAPROOTS"]`, and this isn't written to the sysinfo.gap anymore.
Does this make any difference for the involved GAP processes?

ThomasBreuer:
> What I just noticed: the code now no longer sets `sysinfo["GAPROOTS"]`, and this isn't written to the sysinfo.gap anymore. Does this make any difference for the involved GAP processes?

Only GAP.jl needed `sysinfo["GAPROOTS"]`, with the current changes this is no longer needed.

ThomasBreuer (approved):

PR #1155, opened 2025-02-10
Avoid OtherPackagesLoadedInAdvance

fingolfin:
Recent GAP versions include _StripEscapeSequences, use that instead of
StripEscapeSequences from GAPDoc


ThomasBreuer (approved):
Very good.
(When we removed the `OtherPackagesLoadedInAdvance` from GAP, we had just checked that this removal does not affect packages that are distributed with GAP, and JuliaInterface is not a distributed package.)

PR #1156, opened 2025-02-10
Remove build step from setup job in `CI-distro.yml`

lgoettgens:
All the job needs to do is to find the `Artifact.toml` and parse some things there. This does not require us to build GAP.jl beforehand, saving some few CI minutes and a CI cache.

### DIFF
--- a/.github/workflows/CI-distro.yml
+++ b/.github/workflows/CI-distro.yml
@@ -30,10 +30,6 @@ jobs:
         uses: julia-actions/setup-julia@v2
         with:
           version: '1'
-      - name: "Cache artifacts"
-        uses: julia-actions/cache@v2
-      - name: "Build package"
-        uses: julia-actions/julia-buildpkg@v1
       - name: Get list of GAP packages
         id: set-matrix
         run: julia --project=. -e '


.github/workflows/CI-distro.yml
```diff
@@ -30,10 +30,6 @@ jobs:
         uses: julia-actions/setup-julia@v2
         with:
           version: '1'
-      - name: "Cache artifacts"
-        uses: julia-actions/cache@v2
```
fingolfin, line 34: Are you sure this is an improvement? This caches the artifacts, so instead of downloading 160+ artifacts individually from Julia servers, this downloads a single cache file with all the artifacts.

It is difficult to compare individual CI runs on GitHub VMs, but the last ["CI 1.10 ubuntu" run on master](https://github.com/oscar-system/GAP.jl/actions/runs/13235926442/job/36940686024) took 4.5 minutes while [its counterpart here](https://github.com/oscar-system/GAP.jl/actions/runs/13242017330/job/36959330692) took 7 minutes.

lgoettgens: This is not related at all, since I am only touching the `distro CI` job, not the "normal" testing `CI`.

What this change essentially does is reducing the setup job from > 1 min (e.g. https://github.com/oscar-system/GAP.jl/actions/runs/13235926447/job/36940685206 last run on master) to ~15 sec (https://github.com/oscar-system/GAP.jl/actions/runs/13242017346/job/36959328839?pr=1156).

Issue #1105, opened 2024-12-20
Invalid path to GAP executable in sysinfo


fingolfin:
We currently do this in `regenerate_gaproot()`:
```julia
sysinfo["GAP"] = joinpath(gaproot_mutable, "bin", "gap.sh")
```
However, there is no `bin/gap.sh` in `gaproot_mutable`.

This is irrelevant most of the time, but when building packages it may be important, as some package use that variable in their build system to find a GAP executable, e.g. so they can build their manuals.

So we could use `create_gap_sh()` or a variant of that to provide that. (A variation because at the very least, we'd want to silence the `@info` messages there. But also this command creates a fresh Julia environment which is comparatively slow. Perhaps instead we could just generate a script that uses the currently active environment, or something like that?)

fingolfin:
One place where this came up is in this comment: https://github.com/oscar-system/GAP.jl/pull/1079#issuecomment-2540148097

Also in PR #1067 (see comments in the diff of that PR) we see that some GAP packages (mostly from the homalg project?) would like to launch GAP processes and fail if they can't find one. However, those don't use `sysinfo.gap`, so it is a somewhat different but still highly related issue.

lgoettgens:
> So we could use `create_gap_sh()` or a variant of that to provide that. (A variation because at the very least, we'd want to silence the `@info` messages there. But also this command creates a fresh Julia environment which is comparatively slow. Perhaps instead we could just generate a script that uses the currently active environment, or something like that?)

This approach (using the currently active environment) seems very prone to race conditions to me. If multiple julia sessions with GAP.jl loaded share the same `gaproot_mutable` (as it is currently the case, cf. https://github.com/oscar-system/GAP.jl/pull/1079), to which "currently active environment" should the file in `joinpath(gaproot_mutable, "bin", "gap.sh")` point to?

Issue #1158, opened 2025-02-10
Revise how we handle `sysinfo.gap` and `gac` / replace `Setup.regenerate_gaproot`


fingolfin:
Based on a comment on PR #1079, I think it is time to revise or indeed replace `Setup.regenerate_gaproot`. It is now reduced to generating an additional GAP root dir containing a `sysinfo.gap` and a `gac` executable. 

It wastes some time each time GAP is started to find a C and C++ compiler, even if none is needed.

Perhaps time now to transform this:
- produce a function which takes a directory path and puts `sysinfo.gap` and `gac` (and whatever else) into it
- then only run this function on a fresh temp dir just before invoking `GAP.Packages.install` (perhaps caching the resulting dir for the duration of the GAP.jl session)
- perhaps `create_gap_sh` should also make use of it, and produce a `gac` and `sysinfo.gap` which fit together with the produce `gap` shell script (towards resolving #1105)
- actually with that last bit we could circle back to the top: before calling `GAP.Packages.install` perhaps we should call this enhance `create_gap_sh` and make sure `PackageManger` "sees" its `sysinfo.gap` (and of course `gac`, `gap`)

This would also hopefully allow us to deal with the three packages listed under "Missing Software: GAP" in #1122

This would probably also resolve #840

PR #1162, opened 2025-02-12
Replace `Setup.regenerate_gaproot()` by `Setup.gaproot_for_building()`

lgoettgens:
Implements the second bullet point of https://github.com/oscar-system/GAP.jl/issues/1158 in the following way:

- Add a new function `gaproot_for_building()` that creates a temporary directory just containing `sysinfo.gap`, `gac` (as they are currently in `gaproot_mutable`) and a slim `gap.sh`. This directory is cached for the duration of the current session, but each time we want to use it, we check that the content hasn't been deleted by some OS cronjob that cleans `/tmp` regularly.
- This `gaproot_for_building` is injected both into the build process of `JuliaInterface` and the PackageManager right before calling `InstallPackage` or `CompilePackage`.
- Remove all mentions of the `gaproot_mutable` that we had previously.

Resolves https://github.com/oscar-system/GAP.jl/issues/840, resolves https://github.com/oscar-system/GAP.jl/issues/1105.

For testing the compilation of packages, we need a package with a kernel extension that has no jll (and will never get one). The prime candidate would be `example`. However, its `TestPackageAvailability` does not tell us anything about the build status of its kernel extension, so I don't see a way to tell if compiling it has actually worked. For the meantime, I disabled `GAP_pkg_json_jll` and use `json` for testing, but this should be changed before merging. @fingolfin do you have an idea how to do a proper test here?


lgoettgens:
Uuh, Oscar also wants to put things in a gap root, notably https://github.com/oscar-system/Oscar.jl/blob/7c2d569c224a462f05d0e2d50ebf0da024b0dc3e/src/GAP/utils.jl#L2 which is called from the initialization code of the perfect groups library.
I think the cleanest solution for this would be to instead let Oscar create its own gap root in a scratch space with these files and then call `GAP.Globals.ExtendRootPaths` during `__init__` to hook this into GAP. I'll try to get this working in a matching branch tomorrow, so we can let CI test it together.

In any case, this PR will need a breaking bump in GAP.jl's Version number.

fingolfin:
As for testing we could use a package that is guaranteed to never be shipped in the distro. https://github.com/gap-packages/RegisterPackageTNUMDemo might be perfect candidate. I just made a fresh release of it.

fingolfin:
Yeah it would be good to get rid of `__init_extraperfect`. In this case I actually think there is a very simple solution on the Oscar.jl side: Right now we do
```
function __init_extraperfect()
  for i in [27, 33]
    _write_gap_file(
      "grp/perf$(i).grp",
      "Read(JuliaToGAP(IsString, Oscar._path_extraperfect($(i))));\n",
    )
  end
end
```
(Hm, shouldn't that be using `Oscar_jl.` now?)

Instead of creating these files dynamically, we could just put them all permanently into `gap/grp/` in the Oscar.jl repository, and add that dir as a root path to GAP. (If we then move `OscarInterface` into a new `pkg` subdir we then could also get rid of the `GAP.Globals.SetPackagePath` call for it in `src/Oscar.jl`). 

Having such a GAP root in Oscar might be useful to other things, too (e.g. we can use it to overload specific GAP library files.

Anyway, just a thought. Using a Julia scratch space or an OS temp dir created via `mktemp` both would also work just fine, I'd be happy with any of those solutions.

src/GAP.jl
```diff
@@ -59,12 +59,7 @@ include("setup.jl")
 import Libdl
 import Random
 
-# setup the initial sysinfo dictionary; we'll update this later in __init__
-# this also ensures that Setup.regenerate_gaproot gets precompiled, reducing
-# the startup time a little bit
-const sysinfo = Setup.regenerate_gaproot()
-
-const GAP_VERSION = VersionNumber(sysinfo["GAP_VERSION"])
+const GAP_VERSION = VersionNumber(Setup.read_sysinfo_gap(joinpath(GAP_jll.find_artifact_dir(), "lib", "gap", "sysinfo.gap"))["GAP_VERSION"])
```
fingolfin, line 62: I actually have written code in the past that used `GAP.sysinfo`. From that POV it would be nice to at least retain it, e.g.
```suggestion
const sysinfo = Setup.read_sysinfo_gap(joinpath(GAP_jll.find_artifact_dir(), "lib", "gap", "sysinfo.gap"))

const GAP_VERSION = VersionNumber(sysinfo["GAP_VERSION"])
```

src/setup.jl
```diff
@@ -193,18 +200,21 @@ function build_JuliaInterface(sysinfo::Dict{String, String})
...
                 "LDFLAGS" => JULIA_LDFLAGS * " " * JULIA_LIBS) do
-            run(pipeline(`./configure $(gaproot())`, stdout="build.log"))
+            run(pipeline(`./configure $(gaproot)`, stdout="build.log"))
             run(pipeline(`make V=1 -j$(Sys.CPU_THREADS)`, stdout="build.log", append=true))
         end
     end
 
+    sysinfo = read_sysinfo_gap(joinpath(gaproot, "sysinfo.gap"))
+
     return normpath(joinpath(jipath, "bin", sysinfo["GAParch"]))
```
fingolfin, line 214: If we add the global `sysinfo` back, then this could also be
```suggestion
    return normpath(joinpath(jipath, "bin", GAP.sysinfo["GAParch"]))
```
assuming that something like `import ...GAP: GAP` is added at the top of this file

fingolfin (commented):
Thank you for working on this!

src/GAP.jl
lgoettgens: The difference would be that `GAP.sysinfo` used to be the sysinfo with updated fields that was created during startup, but with this change it would instead be the sysinfo that is bundled with `GAP_jll`.

lgoettgens:
> As for testing we could use a package that is guaranteed to never be shipped in the distro. [gap-packages/RegisterPackageTNUMDemo](https://github.com/gap-packages/RegisterPackageTNUMDemo?rgh-link-date=2025-02-13T12%3A35%3A34Z) might be perfect candidate. I just made a fresh release of it.

I am not sure this is the right candidate, because it does not check if its kernel extension is indeed available in `TestPackageAvailability` (https://github.com/gap-packages/RegisterPackageTNUMDemo/blob/af1401c5bdc10872ccb6169ddf0f7dae99bc33be/PackageInfo.g#L94-L96). Or is this something that should be changed in `RegisterPackageTNUMDemo` since it would be helpful to have it there as well?

lgoettgens:
> Instead of creating these files dynamically, we could just put them all permanently into `gap/grp/` in the Oscar.jl repository, and add that dir as a root path to GAP. (If we then move `OscarInterface` into a new `pkg` subdir we then could also get rid of the `GAP.Globals.SetPackagePath` call for it in `src/Oscar.jl`).
> 
> Having such a GAP root in Oscar might be useful to other things, too (e.g. we can use it to overload specific GAP library files.

I really like that idea. Didn't know that this would work, but I'll give it a go.

> Anyway, just a thought. Using a Julia scratch space or an OS temp dir created via `mktemp` both would also work just fine, I'd be happy with any of those solutions.

If we would ever want to change any files in there, and people use multiple different Oscar versions on the same system, we get basically to the same discussion that we already had in https://github.com/oscar-system/GAP.jl/pull/1079. I.e. how to parameterize the scratch keys, and when to clean up and how to prevent the OS from cleaning up too early.

lgoettgens:
This now contains https://github.com/oscar-system/GAP.jl/pull/1164 as well, which takes care of the timeouts during the REPL tests.

There is one final small change needed to be done once https://github.com/gap-packages/RegisterPackageTNUMDemo/pull/9 is released, otherwise this is ready from my POV.

fingolfin:
Merged and released your `RegisterPackageTNUMDemo` PR, and merged your `expect` PR here.

lgoettgens:
> Merged and released your `RegisterPackageTNUMDemo` PR, and merged your `expect` PR here.

Great! I would like to also wait for the juliainterface update to verify that we indeed do not setup anything in the case that juliainterface is recent enough, and don't search for compiler etc.

lgoettgens:
This works as expected, i.e. no compiler search shenanigans on startup when JuliaInterface is not touched

src/setup.jl
```diff
@@ -267,8 +281,8 @@ function create_gap_sh(dstdir::String)
             READ_STARTUP_FILE="yes"
         else
             READ_STARTUP_FILE="no"
-        fi
-        exec $(joinpath(Sys.BINDIR, Base.julia_exename())) --startup-file=\$READ_STARTUP_FILE --project=$(dstdir) -i -- "$(gap_sh_path)" "\$@"
+        fi
```
fingolfin, line 284:
```suggestion
        fi
```

fingolfin:
Fails with OSCAR release:
```
ERROR: LoadError: InitError: UndefVarError: gaproot not defined
Stacktrace:
  [1] getproperty
    @ ./Base.jl:26 [inlined]
  [2] _write_gap_file(relpath::String, content::String)
    @ Oscar ~/.julia/packages/Oscar/Z2pwW/src/GAP/utils.jl:2
  [3] __init_extraperfect()
    @ Oscar ~/.julia/packages/Oscar/Z2pwW/src/Groups/libraries/perfectgroups.jl:253
  [4] __init_group_libraries()
    @ Oscar ~/.julia/packages/Oscar/Z2pwW/src/Groups/libraries/libraries.jl:72
  [5] __init__()
    @ Oscar ~/.julia/packages/Oscar/Z2pwW/src/Oscar.jl:118
```

fingolfin:
(which is OK and expected as it is `gaproot`)

lgoettgens:
> (which is OK and expected as it is `gaproot`)

This is kind of expected, and the reason why this PR has a "breaking" label. It is already adapted in Oscar master, but we must forbid older Oscar version (1.3.x) to ever use this PR, as that the produces the above error

etc/ci_test.sh
```diff
@@ -15,7 +15,8 @@ ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaInterface.coverage tst/testall.g || AnyFailures=Yes
 gcov -o gen/src/ src/*.c*
 # Remove JuliaInterface with coverage instrumentation
-make clean
+# (`make clean` does not work because the `gaproot_for_building` only exists while GAP.jl runs)
```
fingolfin, line 18: Won't this comment be unclear if one looks at this file in the future? Perhaps like so, with the lines above:
```
# We should clean JuliaInterface and also e.g. delete coverage instrumentation
# now, by calling `make clean`. However, this does not work because the
# `gaproot_for_building` only exists while GAP.jl runs. Since this script is
# mostly used in the CI, and everything will be cleaned out there anyway, we
# simply don't bother.
```

etc/ci_test.sh
```diff
@@ -15,7 +15,8 @@ ${GAP} makedoc.g
...
+FORCE_JULIAINTERFACE_COMPILATION=true ${GAP} --nointeract
```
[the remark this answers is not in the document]
lgoettgens: it rebuilds JuliaInterface with default flags (i.e. without coverage instrumentation). I guess I should really put a comment here.

PR #1171, opened 2025-04-29
Version 0.13.2

fingolfin:
We accumulated quite some changes but made no release for some time. Should do that now, esp. to be ready for Julia 1.12.


lgoettgens (approved):

lgoettgens:
`nconvex` and `toricvarieties` have segmentation faults. I would suggest to just exclude these two for the meantime, and instead add them to https://github.com/oscar-system/GAP.jl/issues/1122

PR #1172, opened 2025-05-07
Move `gap_packages_rootdir` into a scratch space

fingolfin:
... instead of `~/.julia/gaproot/v4.14/`

Pro: scratch spaces are garbage collected, while I'd still have `~/.julia/gaproot/v4.12` around if I hadn't manually deleted it.

Con: harder for users to "find" this dir if they need to (it is now `~/.julia/scratchspaces/c863536a-3901-11e9-33e7-d5cd0df7b904/v4.14/`), though of course they can call `GAP.Packages.gap_packages_rootdir()`

(But I know the location `~/.julia/gaproot/` "by heart" while I am sure I'll have to look up the UUID each time).

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -3,10 +3,11 @@ module Packages
 
 import Downloads
 import Pidfile
+import Scratch: @get_scratch!
 import ...GAP: disable_error_handler, Globals, GapObj, replace_global!, RNamObj, Wrappers
 import ...GAP: gap_pkg_jlls, Compat, GAP_VERSION, GAP_jll, GAP_lib_jll
 
-gap_packages_rootdir() = joinpath(Base.DEPOT_PATH[1], "gaproot", "v$(GAP_VERSION.major).$(GAP_VERSION.minor)")
+gap_packages_rootdir() = @get_scratch!("v$(GAP_VERSION.major).$(GAP_VERSION.minor)")
 
 const DEFAULT_PKGDIR = Ref(joinpath(gap_packages_rootdir(), "pkg"))
 const DOWNLOAD_HELPER = Ref{Downloads.Downloader}()


src/packages.jl
```diff
@@ -3,10 +3,11 @@ module Packages
 
 import Downloads
 import Pidfile
+import Scratch: @get_scratch!
 import ...GAP: disable_error_handler, Globals, GapObj, replace_global!, RNamObj, Wrappers
 import ...GAP: gap_pkg_jlls, Compat, GAP_VERSION, GAP_jll, GAP_lib_jll
 
-gap_packages_rootdir() = joinpath(Base.DEPOT_PATH[1], "gaproot", "v$(GAP_VERSION.major).$(GAP_VERSION.minor)")
+gap_packages_rootdir() = @get_scratch!("v$(GAP_VERSION.major).$(GAP_VERSION.minor)")
```
lgoettgens, line 10:
```suggestion
gap_packages_rootdir() = @get_scratch!("gap_packagedir_v$(GAP_VERSION.major).$(GAP_VERSION.minor)")
```
I think it would make sense to somehow name this. Two benefits:
1. you are able to find this inside your `~/.julia/scratchspaces` with an appropriate `find` call
2. the same folder also contains the gaproots (at least until https://github.com/oscar-system/GAP.jl/pull/1162)

lgoettgens (commented):
is this an intended use of scratchspaces? as far as i can remember, deleting a packages scratchspace and recompiling the package should make no semantical difference to the package.
But in this case, it would remove installed gap packages; thus potentially breaking code

lgoettgens:
> is this an intended use of scratchspaces? as far as i can remember, deleting a packages scratchspace and recompiling the package should make no semantical difference to the package.
> But in this case, it would remove installed gap packages; thus potentially breaking code

After looking through `Scratch.jl` and the implementation of `Pkg.gc()` this seems to be pretty safe to use. IIUC, Pkg will just clear scratchspaces whose corresponding `Project.toml` that called them is no longer present on disk. Ie: scratchspaces of deved packages are never cleared. scratchspaces of stable versions are cleared when the stable version gets removed due to no manifest no longer using it.

Anyway, it might make sense to document in `GAP.Packages.install` that we don't guarantee that installed packages stay (across sessions / across patch releases of GAP.jl / etc)

lgoettgens:
> Concerning the argument that the outdated scratchspace for GAP 4.12 automatically gets garbage collected when GAP.jl uses a newer GAP version, how does this work?

For everyone using a dev version of GAP.jl (aka us three), it simply doesn't.
For everybody else, julia keeps track which Project.toml files correspond to a package that used the scratchspace for (let's say gap 4.14). This will (for the normal user) be something like `.julia/packages/somehashforthecurrentGAPjlversion/Project.toml`. When we now update the underlying gap, we will release a new version of GAP.jl which has a different somehash value. Thus, once the previous GAP.jl version gets gced, the scratchspace for 4.14 has no corresponding project anymore and can thus get removed as well (after some grace period).

Of course this just works for future updates to gap. The `.julia/gaproot` and all of the old folders therein must be removed manually (or just kept there forever)

ThomasBreuer:
Thanks @lgoettgens.
So the garbage collection depends on the GAP.jl version not on the GAP version.

Then the GAP version number makes sense as a part of the name of the scratchspace only because after an update to a newer GAP, the outdated scratchspace does *not* automatically get garbage collected.

lgoettgens:
Bump @fingolfin. Iirc this looked good, apart from my comment about the scratch name: https://github.com/oscar-system/GAP.jl/pull/1172#discussion_r2079168193
It would be great if we could include this into v0.14.0

fingolfin:
Applied your patch, fine to go from my side.

Issue #1170, opened 2025-03-31
Regression GAP.jl loading got slower going from 0.12.3 to 0.13.x / master, and also from Julia 1.10 to 1.11


fingolfin:
Thanks to @albinahlback for pointing out this regression to me.

With GAP.jl 0.12.3 and Julia 1.10.9, on my M1 MacBook Pro, I get this:
```
julia> @time using GAP
 ┌───────┐   GAP 4.13.1 of 2024-06-11
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: aarch64-apple-darwin20-julia1.10-64-kv9
 Configuration:  gmp 6.2.1, Julia GC, Julia 1.10.9, readline
 Loading the library and packages ...
 Packages:   AClib 1.3.2, Alnuth 3.2.1, AtlasRep 2.1.8, AutPGrp 1.11, CRISP 1.4.6,
             Cryst 4.1.27, CrystCat 1.1.10, CTblLib 1.3.9, FactInt 1.6.3, FGA 1.5.0,
             Forms 1.2.11, GAPDoc 1.6.7, genss 1.6.9, IRREDSOL 1.4.4,
             JuliaInterface 0.12.3, LAGUNA 3.9.6, orb 4.9.0, Polenta 1.3.10,
             Polycyclic 2.16, PrimGrp 3.4.4, RadiRoot 2.9, recog 1.4.4,
             ResClasses 4.7.3, SmallGrp 1.5.3, Sophus 1.27, SpinSym 1.5.2,
             StandardFF 1.0, TomLib 1.2.11, TransGrp 3.6.5, utils 0.85
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
  1.919437 seconds (6.41 M allocations: 585.247 MiB, 18.17% gc time, 13.90% compilation time)
```

With GAP.jl 0.12.3 and Julia 1.11.4, on my M1 MacBook Pro, I get this:
```
julia> @time using GAP
 ┌───────┐   GAP 4.13.1 of 2024-06-11
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: aarch64-apple-darwin20-julia1.11-64-kv9
 Configuration:  gmp 6.3.0, Julia GC, Julia 1.11.4, readline
 Loading the library and packages ...
 Packages:   AClib 1.3.2, Alnuth 3.2.1, AtlasRep 2.1.8, AutPGrp 1.11, Browse 1.8.21,
[...]

PR #1179, opened 2025-05-17
Stop using Pkg

fingolfin:
... by copying the code for computing tree hashes from it.

I wrote this back in February, and also asked if there was interest in moving the treehash code into a separate package but the reaction to that was rather negative.

While cleaning out old branches in my clone of GAP.jl I found it again and thought it might help a bit with issue #1170... I  haven't done any benchmarks, though, not sure it is worth it (but since I wrote the code already, I thought there is no harm in at least putting it up here)


Project.toml
```diff
@@ -98,9 +98,9 @@ MacroTools = "0.5.13"
 Markdown = "1.6"
 Ncurses_jll = "6.4.1"
 Pidfile = "1.3"
-Pkg = "1.6"
 REPL = "1.6"
 Random = "1.6"
+SHA = "<0.0.1, 1"
```
lgoettgens, line 103:
```suggestion
SHA = "<0.0.1, 0.7, 1"
```
this was available as v0.7 in some julia versions. Don't ask why, but this should fix the failures

src/treehash.jl
```diff
@@ -0,0 +1,174 @@
+# The code in this file is taken from Pkg.GitTools
```
lgoettgens, line 1: Could you add a github permalink here that refers to the concrete repository, file, and hash this is taken from? My main point for this is to make it easier in the future to check for possible bugfixes to the upstream code and allowing it to copy an updated version of the code over to this file

lgoettgens (approved):
Looks good, thanks! One minor comment

Issue #1178, opened 2025-05-16
`expect` tests occasionally fail


fingolfin:
So far always on [macOS like in this log](https://github.com/oscar-system/GAP.jl/actions/runs/15070461748/job/42365248765?pr=1177) or in [this log](https://github.com/oscar-system/GAP.jl/actions/runs/15070461748/job/42365249180?pr=1177).

fingolfin:
Unfortunately I cannot reproduce this on my personal MacBook Pro.

However it can be produced on munk, a mac mini M1 server in KL (@lgoettgens if you'd like to get SSH access, let me know)

PR #1186, opened 2025-06-02
Fix expect tests on macos

lgoettgens:
Maybe resolves https://github.com/oscar-system/GAP.jl/issues/1178, at least with this change I can no longer reproduce in on munk.

I have honestly no idea why this should change anything, but it seems it does.

Makes https://github.com/oscar-system/GAP.jl/pull/1185 obsolete and thus closes https://github.com/oscar-system/GAP.jl/pull/1185.


lgoettgens:
Three rounds of CI now succeeded on macos, so I am confident that this will resolve the issue at least midterm

PR #1190, opened 2025-06-04
Remove Julia.GAP hack

### LINKED ISSUE #1053: Replace `Julia.GAP` by a cleaner solution?

fingolfin:
Resolves #1053 

TODO:
- [ ] document that using `Julia.GAP` is not recommended, and why
  - ... because once we remove the hack that currently it ensure that it is always there
alternative, at least the first time it is called?


fingolfin:
Rebased after merging PR #1188 by @ThomasBreuer -- now this PR is very short as Thomas already removed most uses of `Julia.GAP` in his PR :-)

PR #1191, opened 2025-06-04
Clarify licensing terms, and some related cleanup

### LINKED ISSUE #621: copyright/license for the JuliaInterface manual
Currently the JuliaInterface manual does not contain statements about copyright and license.
Other GAP package manuals (for example AutoDoc) contain such statements.

What would be appropriate?
(Meanwhile the license is mentioned in `PackageInfo.g`, thus it would be possible for AutoDoc to create a statement automatically.)

fingolfin:
- **Fix license fields**
- **Add Copyright to JuliaInterface manual**
- **Clarify license & copyright in READMEs**
- **Add uniform copyright header too all source files**
- **Add Lars as a maintainer**

Resolves #621

Best reviewed on a per-commit basis


lgoettgens (commented):
A few places previously mentioned LGPL3 (and now only GPL3). IANAL and don't feel capable of judging this change in any way, but I still wanted to point it out. 

And there is some kind of inconsistency where we direct people to find a copy of the GPL3 (at some places it points to the gnu.org webpage, and at others to an address in Boston. Can we unify that in any way?

lgoettgens (approved):
Thanks, I don't see any more inconsistencies

PR #1196, opened 2025-06-16
remove the `IsJuliaWrapper` magic, add `#@local` to GAP testfiles

### LINKED ISSUE #1193: Design issue with `JuliaPointer` and `IsJuliaWrapper`
(The following is using OSCAR code, but mainly because I was lazy to convert it to pure GAP.jl code..)

One issue with using `JuliaPointer` and `IsJuliaWrapper` is that while they make it super convenient to call from GAP code into Julia code, the other way around actually get harder: It is now almost impossible to get hold of such a GAP wrapper object on the Julia side, in order to feed it into a GAP function call. 

This becomes a serious problem if one would like to be able to gradually rewrite some code from "GAP with a few calls to Julia" into "Julia with a few calls to GAP".

One trick solution to this is to wrap the value into a GAP list, but that's still difficult...
Suppose we have this on the GAP side:
```
gap> foo := a -> [a + One(QQBarField)];;
```
Then on Julia:
```
julia> e4 = GAP.Globals.E(4)
GAP: E(4)

julia> obj = GAP.Globals.foo(e4)
GAP: [ <Root 1.00000 + 1.00000*im of x^2 - 2x + 2> ]
```
So how to get the inner result? A first naive approach does not work (but arguably should):
```
julia> Vector{GapObj}(obj)
ERROR: Don't know how to convert value of type QQBarFieldElem to type GapObj
```
Given knowledge of the inner layout of GAP lists and a willingness to risk a crash, one can achieve the goal:
```
julia> u = unsafe_pointer_to_objref(unsafe_load(Ptr{Ptr{GapObj}}(GAP.ADDR_OBJ(obj)), 2))
GAP: <Root 1.00000 + 1.00000*im of x^2 - 2x + 2>
```
.... buuut surely this is not a good solution to suggest to users... But it works:
```
julia> GAP.Globals.foo(u)
GAP: [ <Root 2.00000 + 1.00000*im of x^2 - 4x + 5> ]
```

A better way out might be to have a symmetric system of automatic (re)wrapping of Julia objects when going to the GAP side, but that has its own dangers... hrm.

ThomasBreuer:
resolves #1193


pkg/JuliaInterface/src/convert.c
```diff
@@ -38,27 +38,6 @@ jl_value_t * julia_gap(Obj obj)
         return jl_false;
     }
     jl_value_t * result = (jl_value_t *)obj;
```
fingolfin, line 40:
```suggestion
    return (jl_value_t *)obj;
```

pkg/JuliaInterface/src/convert.c
```diff
@@ -38,27 +38,6 @@ jl_value_t * julia_gap(Obj obj)
...
-            result = julia_gap(obj);
-        }
-        else {
-            ErrorMayQuit("JuliaPointer must be a Julia object or an internal "
-                         "GAP object (not a %s)",
-                         (Int)TNAM_OBJ(obj), 0);
-        }
-    }
-    END_GAP_SYNC();
     return result;
```
fingolfin, line 41:
```suggestion
```

CHANGES.md
```diff
@@ -8,6 +8,12 @@
   access the `GAP` module reliably, use the GAP variable `GAP_jl` instead.
 - Add the GAP function `JuliaType`,
   for constructing parametrized Julia types in GAP
+- Remove the GAP filter `IsJuliaWrapper`.
```
fingolfin, line 11: I think this has the potential of breaking user code, so shouldn't it be marked as breaking in the release notes?

Also, not just the filter is removed (this a user could easily deal with), but also the special treatment for it on the kernel level...

```suggestion
- **Breaking:** Remove the GAP filter `IsJuliaWrapper` and its special treatment.
```

CHANGES.md
```diff
@@ -8,6 +8,12 @@
   access the `GAP` module reliably, use the GAP variable `GAP_jl` instead.
 - Add the GAP function `JuliaType`,
```
fingolfin, line 9: Actually probably best to keep all the "breaking" entries at the top, so move it above this line?

fingolfin (approved):

PR #1198, opened 2025-06-30
Bump curlinterface to v2.4.1

lgoettgens:
The only changes between 2.4.0 and 2.4.1 were changes in the CI setup and trivially fixing the tests. With this update, we can re-enable the distro tests for curlinterface.


fingolfin (approved):
Fine by me. Could also go right to 2.4.2 *shrug*

Issue #1142, opened 2025-02-05
Document updating the binary part of `GAP.jl`


lgoettgens:
e.g. in a `README.maintainer.md`.

This should cover how to update `GAP_jll`, `GAP_lib_jll`, `juliainterface` (including its jll), and all the `GAP_pkg_jll`.

This issue is motivated by https://github.com/oscar-system/Oscar.jl/issues/159.

lgoettgens:
A first draft has been added in https://github.com/oscar-system/GAP.jl/pull/1203. But in particular for the required changes after a gap update (e.g. updating the distro artifacts, rebuilding all changed GAP_pkg_*_jlls) there are still some open questions to me.
Thus I suggest that we just document what we do for the next gap update there.

lgoettgens:
This has been resolved by https://github.com/oscar-system/GAP.jl/pull/1203, https://github.com/oscar-system/GAP.jl/pull/1210, and https://github.com/oscar-system/GAP.jl/pull/1281.

PR #1203, opened 2025-07-11
Add first draft for `README.maintainer.md`

lgoettgens:
Progress towards https://github.com/oscar-system/GAP.jl/issues/1142.

This was inspired by the corresponding file in Singular.jl.

For updating GAP, I am not sure in which order we wanna rebuild pkg jlls and update the artifacts without by accident matching different versions of the same gap pkg. Thus, I left that part out for now. Maybe we can add that (or rather put the steps in an order) after the next gap release with the things we do then.

@fingolfin feel free to just push your ideas directly to this branch

### DIFF
--- a/README.maintainer.md
+++ b/README.maintainer.md
@@ -1,0 +1,96 @@
+# Directions for updating GAP.jl
+
+`GAP.jl` depends on the GAP kernel, the GAP library,
+and some glue code in the `juliainterface` GAP package,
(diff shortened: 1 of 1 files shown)


lgoettgens:
I would like to give this another read tonight or tomorrow, and maybe tweak some more things. After that, I am happy with postponing the rest to follow-ups

PR #1209, opened 2025-07-14
Bump julia compat to 1.10

lgoettgens:
cf https://github.com/oscar-system/Oscar.jl/issues/4200


lgoettgens:
Since there is a CI job in gap-system/gap that tests GAP.jl with julia 1.6, that job should be adapted before merging this PR here to keep the CI in gap green. I already prepared https://github.com/gap-system/gap/pull/6040 for that

PR #1210, opened 2025-07-14
README.maintainer.md: Rework section on build against new libjulia

lgoettgens:
As tried out with https://github.com/JuliaPackaging/Yggdrasil/pull/11656 and https://github.com/JuliaPackaging/Yggdrasil/pull/11659

### DIFF
--- a/README.maintainer.md
+++ b/README.maintainer.md
@@ -53,20 +53,19 @@ happening on a new julia release or on julia nightly.
 
 1. Wait for someone to update `libjulia_jll` to a new version.
 
-2. Update the GAP build recipe by bumping the patch part of the version,
-   and changing the `libjulia_jll` version number to the new one.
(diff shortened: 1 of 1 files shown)


lgoettgens:
CI results are almost as good as we could hope for: The julia 1.12-nightly testsuite mostly completes, thus showing that the rebuilds worked. It fails when trying to install Singular.jl (thus exercising https://github.com/oscar-system/GAP.jl/pull/1204; waiting for https://github.com/JuliaPackaging/Yggdrasil/pull/11660 which wasn't released when CI ran).

Only the test in groupoids seem weird: https://github.com/oscar-system/GAP.jl/actions/runs/16274612739/job/45950429175?pr=1210#step:8:33
Do you have any idea what the reason for that could be? Anyway, should have nothing to do with the changes here.

fingolfin (approved):

Issue #1118, opened 2025-01-07
Reminder: things to do if we switch to requiring Julia >= 1.10


fingolfin:
- [x] always use `jl_reinit_foreign_type` (introduced in Julia 1.9 IIRC) and drop a bunch of code from the C kernel, GAP_jll and GAP.jl itself https://github.com/oscar-system/GAP.jl/pull/1211, https://github.com/JuliaPackaging/Yggdrasil/pull/11697, https://github.com/gap-system/gap/pull/6043
- [x] add a package extension to GAP.jl which gets loaded when Nemo is also present, and move the code for conversion to/from Nemo types like `ZZRingElem`, `QQMatrix` etc. from Oscar.jl into that https://github.com/oscar-system/GAP.jl/pull/1212
- [x] drop `setglobal!` alternate code https://github.com/oscar-system/GAP.jl/pull/1208
- [x] check for `VERSION` checks that could be removed https://github.com/oscar-system/GAP.jl/pull/1208
- [x] Replace `Pidfile.jl` by `FileWatching.Pidfile` https://github.com/oscar-system/GAP.jl/pull/1207

PR #1212, opened 2025-07-18
Add Nemo package extension

lgoettgens:
That contains the Nemo<->Gap conversion methods that are currently defined in Oscar.


lgoettgens:
@fingolfin it would be great to hear if this is how you imagined it to be when writing the idea in https://github.com/oscar-system/GAP.jl/issues/1118.
I haven't yet made any effort to move tests over, but the matching downstream CI should already tell us if it works.

also cc @ThomasBreuer

Project.toml
```diff
@@ -98,6 +104,7 @@ Libdl = "1.10"
 MacroTools = "0.5.13"
 Markdown = "1.10"
 Ncurses_jll = "6.4.1"
+Nemo = "0.50.0, 0.51.0"
```
fingolfin, line 107: Drawback: with this one more package needs to be updated each time we make a breaking Nemo release :-(

ext/NemoExt/nemo_to_gap.jl
```diff
@@ -0,0 +1,35 @@
...
+GAP.@install GapObj(obj::QQFieldElem) = Wrappers.QUO(GapObj(numerator(obj)), GapObj(denominator(obj)))
+
+## `PosInf` to GAP infinity
+GAP.@install GapObj(obj::PosInf) = GAP.Globals.infinity
+
+## `ZZMatrix` to matrix of GAP integers
+GAP.@install GapObj(obj::ZZMatrix) = GapObj(Matrix(obj); recursive = true)
+
+## `QQMatrix` to matrix of GAP rationals or integers
+GAP.@install GapObj(obj::QQMatrix) = GapObj(Matrix(obj); recursive = true)
```
fingolfin, line 35: Note an objection, but:
```suggestion
## TODO/FIXME: rewrite to not first convert to `Matrix`
GAP.@install GapObj(obj::ZZMatrix) = GapObj(Matrix(obj); recursive = true)

## `QQMatrix` to matrix of GAP rationals or integers
## TODO/FIXME: rewrite to not first convert to `Matrix`
GAP.@install GapObj(obj::QQMatrix) = GapObj(Matrix(obj); recursive = true)
```

Background: If I understand this right, this first converts to Julia matrices and then converts those to GAP matrices. So there is quite some room for improvement... Actually our code converting Julia matrices to GAP seems to be in need of improvement:
```
julia> M = matrix(ZZ, [1 2 ; big(3)^100 4])
[                                               1   2]
[515377520732011331036461129765621272702107522001   4]

julia> typeof(M)
ZZMatrix

julia> M2 = Matrix(M)
2×2 Matrix{ZZRingElem}:
 1                                                 2
 515377520732011331036461129765621272702107522001  4

julia> @b GapObj($M)
705.719 ns (18 allocs: 592 bytes)

julia> @b GapObj($M2)
623.645 ns (18 allocs: 544 bytes)

julia> @b GAP.GapObj_internal($M, nothing, Val(false))
683.897 ns (18 allocs: 592 bytes)

julia> @b GAP.GapObj_internal($M2, nothing, Val(false))
516.892 ns (18 allocs: 544 bytes)
```
No matter what I do, there are 18 allocations.

fingolfin (approved):

Project.toml
fingolfin: sure, it wasn't really an objection anyway, just an observation

lgoettgens:
> So what needs to be done to change this from draft to ready, @lgoettgens ?

Someone clicking the button :)

Please review https://github.com/oscar-system/Oscar.jl/pull/5139as well before merging this. In particular please check, if there are more conversions in Oscar that should move here

fingolfin:
https://github.com/oscar-system/Oscar.jl/pull/5139 also looks good to me

lgoettgens:
Finally, CI is happy (apart from the matching job not having the name that github expects, so needs a force-merge).

PR #1213, opened 2025-07-21
Remove julia_to_gap again

fingolfin:
... and hopefully permanently this time (last time we brought
it back because it broke too much down stream code, but that
was back in December 2024).


lgoettgens:
Can you also prepare a PR for OSCAR that passes tests when run together with this PR (so maybe with the same branch name as here and then use downstream CI)?
I think having that before merging this PR would both ease the adaption of Oscar once we release GAP.jl 0.15, and it would verify that we don't break too much downstream code this time.

lgoettgens:
Hecke doesn't contain the string at all.

lgoettgens (approved):

CHANGES.md
```diff
@@ -3,6 +3,9 @@
 ## Version 0.15.0-DEV (released 2025-XX-XX)
 
 - Drop support for Julia 1.6 to 1.9; require Julia 1.10 or later.
+- Remove `julia_to_gap` (code which called it, say as `julia_to_gap(val)`,
+  can instead use `GapObj(val)`; code that provided custom methods
+  for it should instead use the `GAP.@install` macro)
```
lgoettgens, line 8:
```suggestion
- **Breaking:** Remove `julia_to_gap` (code which called it, say as
  `julia_to_gap(val)`, can instead use `GapObj(val)`; code that provided
  custom methods for it should instead use the `GAP.@install` macro).
```

PR #1214, opened 2025-07-23
Improve compatibility with JET

fingolfin:
Move top level @ccall into helper functions. Avoids the problem
described at <https://github.com/aviatesk/JET.jl/issues/734>.

Mind you, `using JET ; report_package("GAP")` still runs into an
error, but at least a different one.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -9,11 +9,15 @@
 ##  SPDX-License-Identifier: LGPL-3.0-or-later
 ##
 
-const GapObj = @ccall libgap.GAP_DeclareGapObj(:GapObj::Symbol, GAP::Module, Any::Any)::Any
+_declare_gap_obj() = @ccall libgap.GAP_DeclareGapObj(:GapObj::Symbol, GAP::Module, Any::Any)::Any
 
-const SmallBag = @ccall libgap.GAP_DeclareBag(:SmallBag::Symbol, GAP::Module, Any::Any, 0::Cint)::Any
+_declare_bag(sym::Symbol, large::Bool) = @ccall libgap.GAP_DeclareBag(sym::Symbol, GAP::Module, Any::Any, large::Cint)::Any
 
-const LargeBag = @ccall libgap.GAP_DeclareBag(:LargeBag::Symbol, GAP::Module, Any::Any, 1::Cint)::Any
+const GapObj = _declare_gap_obj()::Type
+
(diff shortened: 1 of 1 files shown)


src/types.jl
```diff
@@ -9,11 +9,15 @@
...
 
-const SmallBag = @ccall libgap.GAP_DeclareBag(:SmallBag::Symbol, GAP::Module, Any::Any, 0::Cint)::Any
+_declare_bag(sym::Symbol, large::Bool) = @ccall libgap.GAP_DeclareBag(sym::Symbol, GAP::Module, Any::Any, large::Cint)::Any
 
-const LargeBag = @ccall libgap.GAP_DeclareBag(:LargeBag::Symbol, GAP::Module, Any::Any, 1::Cint)::Any
+const GapObj = _declare_gap_obj()::Type
+
+const SmallBag = _declare_bag(:SmallBag::Symbol, false)
+
+const LargeBag = _declare_bag(:LargeBag::Symbol, true)
```
lgoettgens, line 20:
```suggestion
const SmallBag = _declare_bag(:SmallBag, false)

const LargeBag = _declare_bag(:LargeBag, true)
```
this is unnecessary and removing it makes it more readable IMO

src/GAP_pkg.jl
```diff
@@ -77,8 +64,10 @@ function gap_pkg_artifact_dir(pkgname)
 end
 
 function setup_overrides()
-    for pkg in gap_pkgs_with_overrides
-        jll = getproperty(GAP, Symbol("GAP_pkg_$(pkg)_jll"))
+    gap_pkg_jll_names = filter(startswith("GAP_pkg"), String.(names(GAP; imported=true)))
```
lgoettgens, line 67: This filter looks a bit brittle to me, but CI should tell us once we accidentally introduce other variables that match this pattern, right?

src/GAP_pkg.jl
```diff
@@ -77,8 +64,10 @@ function gap_pkg_artifact_dir(pkgname)
...
+    for pkg in gap_pkg_jll_names
+        jll = getproperty(GAP, Symbol(pkg))
```
lgoettgens, line 69:
```suggestion
        jll = getproperty(GAP, Symbol(pkg))
        @assert jll isa Module
```
At least with this we will notice future regression easily

lgoettgens (approved):

PR #1216, opened 2025-07-28
Enable more deprecation errors in CI

lgoettgens:
I am running into some deprecation error issues in https://github.com/oscar-system/GAP.jl/pull/1212, so I looked for a missing `--depwarn=error`, and this should be it.

This PR will also get updates for all the deprecation errors I will encounter.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -70,7 +70,7 @@ jobs:
         run: julia --color=yes -e 'using Pkg ; Pkg.add(["Singular", "Nemo"]) ; using Singular, Nemo'
       - name: "GAP tests"
         run: |
-          julia --project=. -e 'import GAP; GAP.create_gap_sh("/tmp")'
+          julia --project=. --depwarn=error -e 'import GAP; GAP.create_gap_sh("/tmp")'
           export GAP="/tmp/gap.sh -A --quitonbreak --norepl"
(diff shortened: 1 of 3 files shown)


ThomasBreuer (changes requested):
There are also two instances of the deprecated `Nemo.arb` in `pkg/JuliaExperimental/julia/realcyc.jl`, to be replaced by `Nemo.ArbFieldElem`.

lgoettgens:
> There are also two instances of the deprecated `Nemo.arb` in `pkg/JuliaExperimental/julia/realcyc.jl`, to be replaced by `Nemo.ArbFieldElem`.

Good catch, fixed in [`7b73211` (#1216)](https://github.com/oscar-system/GAP.jl/pull/1216/commits/7b73211b2ecdc01f4bd5fde21a1bb715f9d06145). I have no idea why this did not create a test failure, but that may be due to the unusual way it was used (as a type hint on the lhs of a `=`).

ThomasBreuer (approved):
I do not understand the failure (segmentation fault) in the tests of the GAP package YangBaxter; I have restarted this job.

(Concerning `Nemo.arb` vs. `Nemo.ArbFieldElem`, I got an error message when running the tests locally, but the error occurred at runtime, not already when the Julia code got loaded.)

PR #1217, opened 2025-07-29
Drop unnecessary compat on `Compat`

lgoettgens:
as all we used it for is available from julia base on all supported julia versions.


ThomasBreuer (approved):
Thanks.

Issue #736, opened 2021-10-16
Optimize `Vector{GapObj}` and some other similar cases


fingolfin:
Consider this little benchmark:
```
julia> using GAP, BenchmarkTools

julia> L = GAP.evalstr("List([3..6],SymmetricGroup)")
GAP: [ Sym( [ 1 .. 3 ] ), Sym( [ 1 .. 4 ] ), Sym( [ 1 .. 5 ] ), Sym( [ 1 .. 6 ] ) ]

julia> @btime typeof(collect(L))
  582.230 ns (13 allocations: 496 bytes)
Vector{Any} (alias for Array{Any, 1})

julia> @btime typeof([x for x in L])
  653.366 ns (6 allocations: 240 bytes)
Vector{GapObj} (alias for Array{GapObj, 1})

julia> @btime typeof(Vector{GapObj}(L))
  2.285 μs (11 allocations: 608 bytes)
Vector{GapObj} (alias for Array{GapObj, 1})
```

The performance of `Vector{GapObj}(L)` is disappointing. I think it is because of the `IdDict` we create and maintain, but for nought, at least in this case. Also note how `collect` is the fastest, but returns a suboptimal type.

In this particular case, one can do quite a bit better:
```
julia> function f(x::GapObj) v=Vector{GapObj}(undef,length(x)); for i in 1:length(x) v[i]=L[i] end ; return v end
f (generic function with 1 method)

julia> @btime typeof(f(L))
  261.013 ns (1 allocation: 112 bytes)
Vector{GapObj} (alias for Array{GapObj, 1})
```

Perhaps we can set up variants of our conversions which perform faster in "such" cases. Where it remains to be settled what exactly "such" cases are.... coming to mind are certainly:
- `T{S}` where `T` is a container like `Vector`, `Set`, ... and `S` either `GapObj` or `GAP.Obj` or `Int`.
- it could also be done for `T{BigInt}` or `T{fmpz}` but then it'd be a tradeoff in performance vs. memory usage, at least in principle. But it might actually better here, as not all "end users" may expect aliasing in these cases, and if someone starts to modify entries in the result in-place, they might be surprised... ?

Even if we can't solve this in general, I think that at least the `T{GapObj}` case is important enough to deserve special treatment here. I.e. special methods for `Vector{GapObj}(x::GapObj)` and perhaps also for `collect(x::GapObj)`

fingolfin:
[...]

PR #1222, opened 2025-08-15
Rewrite the GAP to Julia conversion

### LINKED ISSUE #1089: Rewrite `gap_to_julia`
... along the lines of PR #1029.

I actually started work on that in PR #777 but never found the energy to complete that.

I am tempted to say it'd be best to start over, taking inspiration from both listed PRs.

This would also be an opportunity to rectify issue #688

ThomasBreuer:
- Introduce an internal function `gap_to_julia_internal` for which the (perhaps recursive) methods get installed that do the work.
- Delegate from `gap_to_julia` to `gap_to_julia_internal`.
- Introduce a new object type for caching objects and their Julia types, in order to keep the relation between identical objects. Create these objects only if they are needed.
- Rename some internal types, in order to distinguish between the two conversion directions. We need different objects for bookkeeping in recursive situations.

Due to the changes, there are no "back delegations" anymore. The behaviour changed slightly:

- One test has a different result now, I think the new result is better.
- The method for converting GAP objects to Julia `Set`s was wrong w.r.t. the identity objects in recursive situations; a new test shows the difference.

The changes are breaking in the sense that new methods for GAP to Julia conversion shall now get installed for `gap_to_julia_internal` not for `gap_to_julia`.

addresses #1089

### DIFF
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -174,21 +174,20 @@ function gap_to_julia_internal(
     end
 
     recursive && recursion_dict !== nothing && haskey(recursion_dict, (obj, TT)) && return recursion_dict[(obj, TT)]
-    rec_dict = recursion_info_j(TT, obj, recursive, recursion_dict)
 
-    len_list = length(newobj)
-    new_array = Vector{T}(undef, len_list)
-    for i = 1:len_list
+    ret_val = Set{T}()
+
+    rec_dict = recursion_info_j(TT, obj, recursive, recursion_dict)
+    handle_recursion((obj, TT), ret_val, recursive, rec_dict)
+
+    for i = 1:length(newobj)
         current_obj = ElmList(newobj, i)
         if recursive && !isbitstype(typeof(current_obj))
-            new_array[i] =
-                gap_to_julia_internal(T, current_obj, rec_dict, Val(recursive))
+            push!(ret_val, gap_to_julia_internal(T, current_obj, rec_dict, Val(recursive)))
         else
-            new_array[i] = current_obj
+            push!(ret_val, current_obj)
         end
     end
-    ret_val = TT(new_array)
-    handle_recursion((obj, TT), ret_val, recursive, rec_dict)
     return ret_val
 end
 
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -90,10 +90,10 @@ function gap_to_julia_internal(
     recursive && recursion_dict !== nothing && haskey(recursion_dict, (obj, TT)) && return recursion_dict[(obj, TT)]
 
     len_list = length(obj)
-    new_array = TT(undef, len_list)
+    ret_val = TT(undef, len_list)::TT
 
     rec_dict = recursion_info_j(TT, obj, recursive, recursion_dict)
-    recursion_dict = handle_recursion((obj, TT), new_array, recursive, rec_dict)
+    recursion_dict = handle_recursion((obj, TT), ret_val, recursive, rec_dict)
 
     for i = 1:len_list
         if islist
@@ -104,14 +104,14 @@ function gap_to_julia_internal(
             current_obj = Wrappers.ELM_LIST(obj, i)
         end
         if recursive && !isbitstype(typeof(current_obj))
-            new_array[i] =
+            ret_val[i] =
                 gap_to_julia_internal(T, current_obj, recursion_dict, Val(recursive))
         else
-            new_array[i] = current_obj
+            ret_val[i] = current_obj
         end
     end
 
-    return new_array::TT
+    return ret_val::TT
 end
 
 ## Matrices or lists of lists
@@ -135,21 +135,21 @@ function gap_to_julia_internal(
     recursive && recursion_dict !== nothing && haskey(recursion_dict, (obj, TT)) && return recursion_dict[(obj, TT)]
 
     elm = Wrappers.ELM_MAT
-    new_array = TT(undef, nrows, ncols)
+    ret_val = TT(undef, nrows, ncols)::TT
 
     rec_dict = recursion_info_j(TT, obj, recursive, recursion_dict)
-    recursion_dict = handle_recursion((obj, TT), new_array, recursive, rec_dict)
+    recursion_dict = handle_recursion((obj, TT), ret_val, recursive, rec_dict)
 
     for i = 1:nrows, j = 1:ncols
         current_obj = elm(obj, i, j)
         if recursive && !isbitstype(typeof(current_obj))
-            new_array[i, j] =
+            ret_val[i, j] =
                 gap_to_julia_internal(T, current_obj, recursion_dict, Val(recursive))
         else
-            new_array[i, j] = current_obj
+            ret_val[i, j] = current_obj
         end
     end
-    return new_array::TT
+    return ret_val
 end
 
 ## Sets
@@ -234,23 +234,23 @@ function gap_to_julia_internal(
 
     recursive && recursion_dict !== nothing && haskey(recursion_dict, (obj, TT)) && return recursion_dict[(obj, TT)]
 
-    dict = TT()
+    ret_val = TT()
 
     rec_dict = recursion_info_j(TT, obj, recursive, recursion_dict)
-    recursion_dict = handle_recursion((obj, TT), dict, recursive, rec_dict)
+    recursion_dict = handle_recursion((obj, TT), ret_val, recursive, rec_dict)
 
     names = Wrappers.RecNames(obj)
     names_list = Vector{Symbol}(names)
     for key in names_list
       current_obj = getproperty(obj, key)
       if recursive && !isbitstype(typeof(current_obj))
-        dict[key] =
+        ret_val[key] =
           gap_to_julia_internal(T, current_obj, recursion_dict, Val(true))
       else
-        dict[key] = current_obj
+        ret_val[key] = current_obj
       end
     end
-    return dict
+    return ret_val
 end
 
 ## Generic method:
--- a/src/gap_to_julia.jl
+++ b/src/gap_to_julia.jl
@@ -9,6 +9,10 @@
 ##  SPDX-License-Identifier: LGPL-3.0-or-later
 ##
 
+#############################################################################
+##
+## some utilities
+
 ## Show a specific error on conversion failure.
 struct ConversionError <: Base.Exception
     obj::Any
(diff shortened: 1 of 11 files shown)


fingolfin (commented):
Before your PR:
```
julia> w = collect(1:50);

julia> @b Vector{Int}($w)
48.264 ns (2 allocs: 480 bytes)
```

After your PR:
```
julia> w = collect(1:50);

julia> @b Vector{Int}($w)
20.902 ns (2 allocs: 480 bytes)
```

So that's good already!

I'll look more later

src/gap_to_julia.jl
```diff
@@ -9,361 +9,424 @@
...
+        if recursive && !isbitstype(typeof(current_obj))
+            new_array[i] =
+                gap_to_julia_internal(T, current_obj, rec_dict, Val(recursive))
         else
             new_array[i] = current_obj
         end
     end
-    return Set{T}(new_array)
+    ret_val = TT(new_array)
+    handle_recursion((obj, TT), ret_val, recursive, rec_dict)
```
fingolfin, line 191: I have doubts about this: in our system, normally the new object should be inserted into the recursion dict *before* being filled, in order to handle circular structure. 

I think this could happen here as well, in principle (e.g. a `Set{Any}` could contain a `Vector{Any}` which then contains the initial set again, no?). 

I am motivated to change this so that we can fold `handle_recursion` into `recursion_info_j` (just like I folded it into `recursion_info_g`).

src/gap_to_julia.jl
```diff
@@ -9,361 +9,424 @@
...
+    len = length(parameters)
+    length(obj) == len ||
+        throw(ArgumentError("length of $obj does not match type $T"))
+    list = [
+        gap_to_julia_internal(parameters[i], obj[i], rec_dict, Val(recursive))
+        for i = 1:len
+    ]
+
+    ret_val = T(list)
+    recursion_dict = handle_recursion((obj, TT), ret_val, recursive, rec_dict)
```
fingolfin, line 222: Here in principle I have the same concern as for the `Set` case but at the same time it is just not possible to construct the Tuple *before* recursion, but it also shouldn't be necessary due to how identity works for tuples. I.e. if we construct the tuple multiple times, that's fine, only downside is that we also end up storing it multiple times.

Might be good to add some test cases exercising this "recursive data structure" situations ... So e.g.
```
julia> v = [1, "x"]
2-element Vector{Any}:
 1
  "x"

julia> t = (2, v)
(2, Any[1, "x"])

julia> v[2] = t
(2, Any[1, (#= circular reference @-2 =#)])

julia> v
2-element Vector{Any}:
 1
  (2, Any[#= circular reference @-2 =#])
```

...

src/gap_to_julia.jl
ThomasBreuer: There is a comment in the code:
```
## Assume that the argument `obj` of this function is not self-referential,
## for example we cannot really sort a self-referential GAP list.
## Then we need not worry about creating the result in the end,
## and then adding it to the dictionary that is needed for the recursion.
```
We can change our viewpoint and admit self-referential `obj`, by starting with an empty `Set` and then filling it up with `push!`.

ThomasBreuer:
That's strange:
I am sure there was a comment by @fingolfin on the conversion to `Set{T}` in `src/gap_to_julia.jl`,
and that I responded to this comment. How did the two comments disappear?

fingolfin:
The comments are still there, just marked as "resolved" and hence "collapsed"

lgoettgens:
> Unfortunately I am still seeing a performance regression in some cases. E.g. using the example from issue #736, in Julia 1.10.10:

I just tried to reproduce you benchmarks on my machine. With julia 1.10.10, I am getting similar regressions as you (for the same lines).
However, when using 1.11.6 for the benchmark, none of the cases worsens with this PR. Actually, a bunch of them (both for `Vector{GapObj}` and `Vector{Int}`) need less allocations and time with this PR (compared to master). (One has to note though that comparing 1.10.10 to 1.11.6 won't work that well as there are a lot of changes in julia that heavily influence the results)

fingolfin:
Some more timings on my mac, this time with Julia 1.12.0-rc1 (in light of what @lgoettgens reported):


master:
```julia-repl
julia> using GAP, BenchmarkTools

julia> L = GAP.evalstr("List([3..6],SymmetricGroup)")
GAP: [ Sym( [ 1 .. 3 ] ), Sym( [ 1 .. 4 ] ), Sym( [ 1 .. 5 ] ), Sym( [ 1 .. 6 ] ) ]

julia> @btime collect($L);
  1.913 μs (23 allocations: 880 bytes)

julia> @btime [x for x in $L];
  1.875 μs (24 allocations: 896 bytes)

julia> @btime Vector{GapObj}($L);
  178.540 ns (8 allocations: 480 bytes)

julia> v = GapObj([1,2,3,4]);

julia> @btime Vector{Int}($v);
  98.608 ns (8 allocations: 480 bytes)

julia> @btime Vector{Int}($v; recursive=false);
  108.539 ns (8 allocations: 480 bytes)

julia> function f(x::GapObj) v=Vector{Int}(undef,length(x)); for i in 1:length(x) v[i]=x[i] end ;  v end;

julia> @btime f($v);
  59.978 ns (2 allocations: 96 bytes)
```

This PR:
```julia-repl
julia> using GAP, BenchmarkTools

julia> L = GAP.evalstr("List([3..6],SymmetricGroup)")
GAP: [ Sym( [ 1 .. 3 ] ), Sym( [ 1 .. 4 ] ), Sym( [ 1 .. 5 ] ), Sym( [ 1 .. 6 ] ) ]

julia> @btime collect($L);
  1.833 μs (23 allocations: 880 bytes)

julia> @btime [x for x in $L];
  1.796 μs (24 allocations: 896 bytes)

julia> @btime Vector{GapObj}($L);
  112.928 ns (5 allocations: 448 bytes)

julia> v = GapObj([1,2,3,4]);

julia> @btime Vector{Int}($v);
  137.534 ns (5 allocations: 448 bytes)

julia> @btime Vector{Int}($v; recursive=false);
  95.608 ns (2 allocations: 96 bytes)

julia> function f(x::GapObj) v=Vector{Int}(undef,length(x)); for i in 1:length(x) v[i]=x[i] end ;  v end;

julia> @btime f($v);
  100.854 ns (2 allocations: 96 bytes)
```


So: `Vector{GapObj}` got faster but `Vector{Int}` slower

fingolfin:
But if I look at this at a larger scale, the allocations absolutely do occur, so that doesn't look like a small compiler fluke, does it?
```
julia> v = GapObj(collect(1:10000));

julia> @b sum(v)
940.542 μs (58948 allocs: 1.357 MiB)

julia> @b collect(v)
834.083 μs (48989 allocs: 1.524 MiB)
```


And I think I now have an idea what might be going on: the problem is *boxing* of `Int64` values.

Indeed:
```julia-repl
julia> @b v[511]
39.485 ns

julia> @b v[512]
41.923 ns (1 allocs: 16 bytes)

julia> ptr = Ptr{Nothing}(UInt64(0xfa1))
Ptr{Nothing} @0x0000000000000fa1

julia> @code_warntype GAP._GAP_TO_JULIA(ptr)
MethodInstance for GAP._GAP_TO_JULIA(::Ptr{Nothing})
  from _GAP_TO_JULIA(ptr::Ptr{Nothing}) @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:21
Arguments
  #self#::Core.Const(GAP._GAP_TO_JULIA)
  ptr::Ptr{Nothing}
Locals
  as_int::Int64
Body::Any
...

julia> @code_native GAP._GAP_TO_JULIA(ptr)
...
; │┌ @ int.jl:540 within `>>` @ int.jl:533
	asr	x0, x20, #2
; │└
Lloh0:
	adrp	x8, _ijl_box_int64@GOTPAGE
Lloh1:
	ldr	x8, [x8, _ijl_box_int64@GOTPAGEOFF]
	blr	x8
	b	LBB0_9
...
```

The problem is that the return type of `_GAP_TO_JULIA` is `Any` (it has to be, because we also unwrap (unbox) any GapObj which is a `T_JULIA` wrapper.

We don't see it for "tiny" integers (below 1024 or so I think) because the Julia kernel caches boxes for those as an optimization (see `#define NBOX_C 1024` and below in `src/datatype.c` in the Julia kernel).

So the good news is this unrelated to this PR, the bad news is that it impedes all kinds of optimizations. E.g. despite having this:
```
@wrap Length(x::Any)::GapInt
```
we can still end up with that wrapper allocating if the returned length exceeds the cached bound. So:
```
julia> @b GAP.Wrappers.Length($v)
30.307 ns (1 allocs: 16 bytes)

julia> @b length($v)
30.009 ns (1 allocs: 16 bytes)
```

Basically, the type assertion `::GapInt` is too late. If we could somehow convey that information to `GAP._GAP_TO_JULIA` it could perhaps avoid the boxing with some help by the union splitting code... Question is how to achieve that. Maybe:
- for `ValueGlobalVariable` add an optional type argument 
- likewise for `ElmList`
- ... and that leaves the GAP function call code:
```
# specialize non-kwargs versions, which increases performance
(f::GapObj)() = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f) : slow_call_gap_func_nokw(f, ()))
(f::GapObj)(a1) = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f, a1) : slow_call_gap_func_nokw(f, (a1,)))
(f::GapObj)(a1, a2) = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f, a1, a2) : slow_call_gap_func_nokw(f, (a1, a2)))
(f::GapObj)(a1, a2, a3) = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f, a1, a2, a3) : slow_call_gap_func_nokw(f, (a1, a2, a3)))
(f::GapObj)(a1, a2, a3, a4) = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f, a1, a2, a3, a4) : slow_call_gap_func_nokw(f, (a1, a2, a3, a4)))
(f::GapObj)(a1, a2, a3, a4, a5) = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f, a1, a2, a3, a4, a5) : slow_call_gap_func_nokw(f, (a1, a2, a3, a4, a5)))
(f::GapObj)(a1, a2, a3, a4, a5, a6) = _GAP_TO_JULIA(is_func(f) ? _call_gap_func(f, a1, a2, a3, a4, a5, a6) : slow_call_gap_func_nokw(f, (a1, a2, a3, a4, a5, a6)))
```

Unlike the first two, this is user facing, so we can't just modify it.

But we could have versions that are specialized and take a return type hint, and which are then at least by the wrappers generated by `@wrap` ...


anyway this is way beyond scope of this PR. I'll turn this into a separate issue later.

fingolfin:
So I redid my `Vector{Int}` measurement, where it looked as if it got slower with this PR. New results:

`master` + Julia 1.10.10
```julia-repl
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  210.998 ns (7 allocations: 496 bytes)
```

`master` + Julia 1.11.6
```julia-repl
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  171.167 ns (8 allocations: 480 bytes)
```

`master` + Julia 1.12.0-rc1
```julia-repl
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  82.345 ns (8 allocations: 480 bytes)
```

This PR + Julia 1.10.10
```julia-repl
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  524.521 ns (8 allocations: 624 bytes)
```

This PR + Julia 1.11.6
```julia-repl
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  150.633 ns (8 allocations: 480 bytes)
```

This PR + Julia 1.12.0-rc1
```julia-repl
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  72.842 ns (5 allocations: 448 bytes)
```

So... it's complicated... Fastest is this PR and Julia 1.12, slowest ist also this PR but with Julia 1.10. Ugh.

But this branch is also faster than master when used in Julia 1.11, so I'd say: let's merge this, and I'll open a bunch of issues so we can improve more later.

fingolfin (approved):

Issue #1233, opened 2025-08-25
Conversion from `GapObj` to e.g. `Vector{Int}` should not enable recursion


fingolfin:
Compare this in Julia 1.10.10:
```
julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v);
  213.810 ns (7 allocations: 496 bytes)

julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v; recursive=true);
  214.352 ns (7 allocations: 496 bytes)

julia> v = GapObj([1,2,3,4]); @btime Vector{Int}($v; recursive=false);
  179.537 ns (3 allocations: 432 bytes)
```

The difference in allocations and performance comes from allocating (or not) an `IdDict`.

The numbers are different in different Julia versions, and also with PR #1222 they change, but the qualitative result is the same: the default is `recursive=true` and it is slower and has allocation overhead.

PR #1238, opened 2025-09-02
next step of improving the `gap_to_julia` conversion

ThomasBreuer:
(This is a follow-up of #1222.)

- documentation:

  - Extend the documentation of "Explicit GAP-to-Julia and Julia-to-GAP Conversions": Julia types control the conversions, meaning of recursion, etc.

  - Add/extend docstrings to some internal functions that are used for conversions.

  - Do not document a `recursion_dict` argument of `gap_to_julia` and `GapObj`.

- code:

  - Add `_needs_tracking_gap_to_julia`, use it to switch off recursion where possible, in order to avoid the creation of dictionaries for tracking object identity.

  - Add dedicated `gap_to_julia_internal` methods for the prescribed types `Vector`, `Matrix`, `Set`, `Dict`.

  - Support the conversion to tuples with `Vararg` parameters. (This was simply missing up to now.)

  - Change the condition for the shortcuts where subobjects can be kept without calls of `gap_to_julia_internal`: For non-bitstypes, admit this shortcut only if the subobject has already the prescribed type; up to now, Julia's `convert` was called if the type did not fit, and threw an exception if a conversion was not possible. In practice, the change means that more non-recursive conversions are possible.

- breaking change (code and documentation):
  - Change the default of the GAP-to-Julia conversion to be non-recursive; up to now, the default for this direction was recursive conversion.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,12 @@
   and `gap_to_julia(T, obj; recursive = false)` was used to request
   non-recursive conversion.
 
+- **Breaking:** Do not support the optional `GapCacheDict` argument for
+  `GapObj` anymore, because recursive conversion to GAP is handled by
+  `GAP.GapObj_internal` and does not involve cals to `GapObj`.
+  We had always stated in the documentation that users should not enter
+  this argument because it gets created automatically in recursive conversions.
+
 ## Version 0.15.3 (released 2025-09-24)
 
 - Drop dependency on `Pkg`.
--- a/src/julia_to_gap.jl
+++ b/src/julia_to_gap.jl
@@ -78,9 +78,6 @@ The following `GapObj` conversions are supported by GAP.jl.
 """
 GapObj(x; recursive::Bool = false) = GapObj_internal(x, nothing, BoolVal(recursive))
 
-GapObj(x, cache::GapCacheDict; recursive::Bool = false) = GapObj_internal(x, cache, BoolVal(recursive))
-#TODO: omit this method?
-
 # The calls to `GAP.@install` install methods for `GAP.GapObj_internal`
 # so we must make sure it is declared before
 """
--- a/test/conversion.jl
+++ b/test/conversion.jl
@@ -539,8 +539,7 @@ end
     @test GapObj(1:3:10) == r
     @test_throws GAP.ConversionError GapObj(1:2^62)
 
-    r = GapObj(1:2:11, IdDict())
-    @test r == GapObj(1:2:11)
+    r == GapObj(1:2:11)
     @test String(GAP.Globals.TNAM_OBJ(r)) == "list (range,ssort)"
     r = GAP.Obj(1:10)
     @test String(GAP.Globals.TNAM_OBJ(r)) == "list (range,ssort)"


ThomasBreuer:
Compared with the timings from #736, we get fewer allocations.
```
julia> L = GAP.evalstr("List([3..6],SymmetricGroup)")
GAP: [ Sym( [ 1 .. 3 ] ), Sym( [ 1 .. 4 ] ), Sym( [ 1 .. 5 ] ), Sym( [ 1 .. 6 ] ) ]

julia> @btime typeof(collect($L))
  824.286 ns (15 allocations: 560 bytes)
Vector{GapObj} (alias for Array{GapObj, 1})

julia> @btime typeof([x for x in $L])
  819.961 ns (16 allocations: 576 bytes)
Vector{GapObj} (alias for Array{GapObj, 1})

julia> @btime typeof(Vector{GapObj}($L))
  208.511 ns (1 allocation: 80 bytes)
Vector{GapObj} (alias for Array{GapObj, 1})
```

ThomasBreuer:
Also issue #1233 can be closed when this pull request gets merged, recursion gets switched off in the conversion to vectors of bitstype objects.

docs/src/conversion.md
```diff
@@ -48,6 +48,105 @@ The exceptions are as follows.
...
+
+   - For Julia-to-GAP conversion, the default is non-recursive conversion.
+
+   - For GAP-to-Julia conversion, the default is recursive conversion.
+
+   - For Julia-to-GAP conversion, recursion stops at subobjects of type
+     `GapObj`.
+
+   - For GAP-to-Julia conversion, recursion stops at subobjects that do
+     not have the type `GapObj`.
```
fingolfin, line 84: Maybe I am misunderstanding something, or maybe I am too picky, but: What about converting a GAP list like `[1,2,3]` to `Vector{Int16}` ? Technically the entries of the GAP list are not `GapObj`, they are `Int`... Yet we do "recurse" to convert those integers (good!), as we see here:
```gap-repl
julia> l = GapObj([1,2,3])
GAP: [ 1, 2, 3 ]

julia> Vector{Int16}(l)
3-element Vector{Int16}:
 1
 2
 3

julia> Vector{Float32}(l)
3-element Vector{Float32}:
 1.0
 2.0
 3.0
```

docs/src/conversion.md
```diff
@@ -48,6 +48,105 @@ The exceptions are as follows.
...
+   - GAP lists and records can have subobjects,
+     the same holds for various Julia objects such as vectors, matrices,
+     tuples, and dictionaries.
+     One may or may not want to convert the subobjects recursively,
+     this is controlled by the `recursive` keyword argument of the functions
+     `GAP.gap_to_julia` and `GapObj`, which can be set to `true` or `false`.
+
+   - For Julia-to-GAP conversion, the default is non-recursive conversion.
+
+   - For GAP-to-Julia conversion, the default is recursive conversion.
```
fingolfin, line 78: Why is that so, actually?

I am tempted to think this is a result of "nested" conversions like `Vector{Vector{Int}}(x)` requiring it to work:
```
julia> l = GapObj([1,2,3])
GAP: [ 1, 2, 3 ]

julia> l2 = GapObj([l,l])
GAP: [ [ 1, 2, 3 ], [ 1, 2, 3 ] ]

julia> Vector{Vector{Int}}(l2)
2-element Vector{Vector{Int64}}:
 [1, 2, 3]
 [1, 2, 3]

julia> Vector{Vector{Int}}(l2; recursive=false)
ERROR: MethodError: Cannot `convert` an object of type GapObj to an object of type Vector{Int64}
```

With this PR, the error is gone (YAY!) and so `recursive` really becomes `track_identity` or so... Indeed:
```
julia> ll1 = Vector{Vector{Int}}(l2)
2-element Vector{Vector{Int64}}:
 [1, 2, 3]
 [1, 2, 3]

julia> ll2 = Vector{Vector{Int}}(l2; recursive=false)
2-element Vector{Vector{Int64}}:
 [1, 2, 3]
 [1, 2, 3]

julia> ll1[1] === ll1[2]
true

julia> ll2[1] === ll2[2]
false
```


So perhaps we switch the default here now (well, or rather in the next breaking release) ?

docs/src/conversion.md
fingolfin: Maybe referring to `GAP.Obj` instead would do the job?

docs/src/conversion.md
```diff
@@ -48,6 +48,105 @@ The exceptions are as follows.
...
+
+     - It can happen that the results of recursive and non-recursive
+       conversion are equal, but they differ w.r.t. the identity of subobjects.
+       For example, the two entries of the GAP list
+       `GAP.evalstr("[ [ 1, 2 ], ~[1] ]")` are identical,
+       the same holds for the two entries of the vector obtained by
+       recursive conversion of this list to an object of type
+       `Vector{Vector{Int}}`;
+       however, the two entries of the vector obtained by
+       recursive conversion of this list to an object of type
```
fingolfin, line 119:
```suggestion
       non-recursive conversion of this list to an object of type
```
??

src/gap_to_julia.jl
```diff
@@ -17,12 +17,21 @@
 """
     gap_to_julia_internal(::Type{T}, x::Any, rec_dict::JuliaCacheDict, ::Val{recursive}) where {T, recursive}
 
-returns an object of type `T` that corresponds to the GAP object `x`.
-
-The function `gap_to_julia` may call `gap_to_julia_internal`,
+Return an object of type `T` that corresponds to the GAP object `x`.
+
+The value of `recursive` can be `GAP.BoolVal(true)` or `GAP.BoolVal(false)`,
```
fingolfin, line 22: `GAP.BoolVal` is an internal helper that only exists as an optimization. I'd not mention it much in the text...
```suggestion
The value of `recursive` can be `Val(true)` or `Val(false)`,
```

fingolfin (commented):
Some comments. I have not yet looked at it all

docs/src/conversion.md
ThomasBreuer: A long time ago, we had decided that GAP-to-Julia conversion is by default recursive, and Julia-to-GAP conversion is by default non-recursive, because that would fit to the expected use cases. This has been documented already in other places.

This decision is not related to the behaviour w.r.t. prescribed nested types.

My point of view is that we can state two kinds of conditions:
First we have a requested Julia type of the conversion result, which can force several layers of recursion.
Once this type is satisfied, there may be further layers down in the object, and then the `recursive` flag decides whether conversion goes on on these layers or not.

In the early times of GAP.jl, we did not think about prescribing nested types, nowadays this is the usual situation:
If one knows what kind of GAP object one wants to convert, one usually knows which Julia type one wants for the conversion result, in particular one wants to avoid a second modification in Julia that would turn the result into one with a better type.
In this sense, the `recursive` flag in the GAP-to-Julia conversion has become less important due to the fact that prescribed nested types such as `Vector{Vector{Int}}` often determine the result completely.
Thus I think that changing the default would not have big consequences.

docs/src/conversion.md
```diff
@@ -48,6 +48,105 @@ The exceptions are as follows.
 
 ## Explicit GAP-to-Julia and Julia-to-GAP Conversions
 
+The following rules hold for explicit conversions.
+
+1. Julia types control the conversions.
+
+   - For conversions from Julia to GAP,
+     there is at most one possibility on the GAP side,
+     and the type of the given Julia object decides which code is used.
```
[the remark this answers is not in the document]
ThomasBreuer: O.k.
(Same for the other direction.)

docs/src/conversion.md
ThomasBreuer: Yes, we can do that.

On the other hand, I would say that the given example does not fit:
The prescribed type `Vector{Int16}` forces a conversion of the list entries to `Int16` independent of the decision about recursion.

fingolfin:
I've rebased this now and cleaned up the commit history a bit to help me review. I'll try to post a few more things later and hope that I can talk through some of it with @ThomasBreuer on Monday to get it merged.

PR #1244, opened 2025-09-08
Update to GAP 4.15.0

lgoettgens:
The steps to actually do the update once the GAP release is there is the following:

1. Build GAP_jll using https://github.com/JuliaPackaging/Yggdrasil/pull/11987
2. Build GAP_lib_jll using https://github.com/JuliaPackaging/Yggdrasil/pull/11988
3. Merge the changes to `common.jl` and `update.jl` from https://github.com/JuliaPackaging/Yggdrasil/pull/12032 with `[skip build] [skip ci]`.
4. Split all changes to packages (except `juliainterface`) off of https://github.com/JuliaPackaging/Yggdrasil/pull/12032 into several small PRs and merge them.
5. Build GAP_pkg_juliainterface_jll. (Two options: 1. Merge https://github.com/oscar-system/GAP.jl/pull/1239 and update https://github.com/JuliaPackaging/Yggdrasil/pull/12032 to use the new merge commit. 2. Build using a branch commit from https://github.com/oscar-system/GAP.jl/pull/1239. Option 1 leaves the master branch temporarily broken, while option 2 builds from a temporary branch. I am not sure which of the two is preferable.)
6. Merge this PR.

### DIFF
--- a/.github/workflows/CI-distro.yml
+++ b/.github/workflows/CI-distro.yml
@@ -75,6 +75,7 @@ jobs:
           - gap-package: 'polymaking'                 # `polymake command not found. Please set POLYMAKE_COMMAND by hand`
           - gap-package: 'profiling'                  # segfaults during testing
           - gap-package: 'ringsforhomalg'             # `Error, found no GAP executable in PATH`
+          - gap-package: 'smallclassnr'               # tests pass, but the job fails. Needs to be investigated further
           - gap-package: 'toricvarieties'             # random segfaults during testing
           - gap-package: 'xgap'                       # no jll
 


lgoettgens:
In the following, I list all errors that occurred while running the GAP.jl and Oscar.jl tests with locally built binaries (all deployed at https://github.com/lgoettgens/NameOfTheJll.jl), with julia 1.11.6 on munk:

1. **compilation on macos:** On macos, the on-the-fly compilation of juliainterface fails. See https://github.com/oscar-system/GAP.jl/pull/1239#issuecomment-3269792746 for the error log
2. **`is_finite` known too early:** https://github.com/oscar-system/Oscar.jl/blob/b99b05560554386d1104fb6372859d609b741696/test/Groups/quotients.jl#L178 fails. But if I understand correctly, the actual thing that is tested there is that after calling `quo` the `has_is_finite` is set. So this is just a case where gap learns too many things about an object too early, and the testcase there needs to be adapted to something gap does not know things about.
3. **logging output???:** Some output of the form ` julia(36724) MallocStackLogging: can't turn off malloc stack logging because it was not enabled.`  (with increasing numbers in the beginning) in various doctests. One example with several of these outputs is https://github.com/oscar-system/Oscar.jl/blob/0a759cdefa61d38c9c1b17f8e1edf990043e5362/src/AlgebraicGeometry/ToricVarieties/ToricMorphisms/attributes.jl#L180. This didn't happen in the first half of doctests, but in the second half occurred in almost every one.
4. **primgrp 4.0.0:** https://github.com/oscar-system/Oscar.jl/blob/f6cf559f1d1a757060ff03b71355e1c567ba8c07/src/Groups/libraries/primitivegroups.jl#L33 fails. I found https://github.com/gap-packages/primgrp/pull/52 which added support for primitive groups up to order 8191, but only using zenodo (and so is a similar case as https://github.com/oscar-system/Oscar.jl/issues/2850). We should decide if we want to answer `has_primitive_groups(n)` with `true` if we don't actually have it but it needs to be downloaded from zenodo. Related failure in https://github.com/oscar-system/Oscar.jl/blob/f6cf559f1d1a757060ff03b71355e1c567ba8c07/src/Groups/libraries/primitivegroups.jl#L138
5. **Changed order in character tables:** https://github.com/oscar-system/Oscar.jl/blob/cc27a3dd3604b667c16389a47ebbf30d2ea931cd/src/Groups/group_characters.jl#L2868 fails (it now returns `[1,2,4,3]`). Many failures in https://github.com/oscar-system/Oscar.jl/blob/cc27a3dd3604b667c16389a47ebbf30d2ea931cd/test/Groups/group_characters.jl#L7-L741, but I think all are just a permutation of rows/cols.
6.

lgoettgens:
> @lgoettgens Thanks. I will adjust the tests from 2. and 5. above.

The list is still growing while my tests run. The main point here is that we (I'll also ping max once the list is final) can decide which of the failures just need a test adaption later in Oscar, and which ones need to be addressed in GAP (before the release at best).

ThomasBreuer:
Concerning the changed ordering of character tables, we had the same problem in GAP tests. The best solution is to change the tests such that they do not depend on a particular ordering, hence these changes are actually independent of switching to the new GAP version; the new tests will look less suggestive (and more ugly).

ThomasBreuer:
Concerning 6. (the changed output in the `test/book/specialized/boehm-breuer-git-fans/explG25_5.jlcon` example),
I cannot reproduce this locally.

(I am running Julia 1.10.9 with the master branch of GAP (4.16dev). For that, I have used the instructions from `GAP.jl/etc/README.md`, section "Using GAP.jl with a different version of GAP than what `GAP_jll` provides". In a fresh Oscar session, the examples from the book chapter are entered, and I do not get the different `rays(c)` output shown above.)

lgoettgens:
It looks like there is an issue with the cryst sources. I'll investigate later if this is a problem of the artifact setup or their server

lgoettgens:
CI seems to be happy now. I added one exception for one of the two new distro packages to the distro CI. Its tests pass but the CI job still fails. My current guess is that this packages uses yet another way to exit gap after running its tests that we currently don't accommodate for. I'll look into enabling that job again once I am back in office next week

ThomasBreuer:
> CI seems to be happy now. I added one exception for one of the two new distro packages to the distro CI. Its tests pass but the CI job still fails. My current guess is that this packages uses yet another way to exit gap after running its tests that we currently don't accommodate for. I'll look into enabling that job again once I am back in office next week

Most packages call `FORCE_QUIT_GAP(1)` in the end of their `testall.g` files.
In `smallclassnr/tst/testall.g`, we have `FORCE_QUIT_GAP( pass )` where `pass` is `true` or `false`.
According to the documentation, both integers and booleans are allowed as argument of `FORCE_QUIT_GAP`.
(And one should use the documented `ForceQuitGap` instead of `FORCE_QUIT_GAP`.)

Does the boolean return value cause problems for the CI job?
Then also the tests of the package `twistedconjugacy` should show the same problem.

PR #1246, opened 2025-09-19
fix `GAP.Packages.versioninfo`

ThomasBreuer:
It may happen that a GAP package is loaded but its directory does not exist.

For example, our test that installs/loads/removes a package creates such a GAP session.
Calling `GAP.Packages.versioninfo` afterwards ran into an error up to now.

### DIFF
--- a/test/packages.jl
+++ b/test/packages.jl
@@ -26,6 +26,7 @@
     @test GAP.Packages.install("https://github.com/gap-packages/RegisterPackageTNUMDemo/releases/download/v0.4/RegisterPackageTNUMDemo-0.4.tar.gz", interactive = false)
     @test GAP.Packages.load("RegisterPackageTNUMDemo")
     @test GAP.Packages.remove("RegisterPackageTNUMDemo", interactive = false)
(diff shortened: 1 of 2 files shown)


test/packages.jl
```diff
@@ -26,6 +26,7 @@
     @test GAP.Packages.install("https://github.com/gap-packages/RegisterPackageTNUMDemo/releases/download/v0.4/RegisterPackageTNUMDemo-0.4.tar.gz", interactive = false)
     @test GAP.Packages.load("RegisterPackageTNUMDemo")
     @test GAP.Packages.remove("RegisterPackageTNUMDemo", interactive = false)
+    GAP.Packages.versioninfo(IOBuffer(); full = true)
```
lgoettgens, line 29:
```suggestion
    GAP.Packages.versioninfo(IOBuffer(); full = true) # test that this does not error, see #1246
```

lgoettgens (approved):
Good catch, thanks!

fingolfin (approved):

Issue #1249, opened 2025-09-24
Race conditions when compiling juliainterface with multiple tasks


lgoettgens:
```
  1 dependency had output during precompilation:
┌ Oscar
│  [ Info: Compiling JuliaInterface ...
└  
     Testing Running tests...
[ Info: Compiling JuliaInterface ...

Adding worker processes
[ Info: /home/oscarci-tester/ssd-data/ssd-runner-02/_work/GAP.jl/GAP.jl/oscar-dev/Oscar/test/runtests.jl -- fixed SEED 42
      From worker 6:	[ Info: Compiling JuliaInterface ...
      From worker 5:	[ Info: Compiling JuliaInterface ...
      From worker 7:	[ Info: Compiling JuliaInterface ...
      From worker 4:	[ Info: Compiling JuliaInterface ...
      From worker 3:	[ Info: Compiling JuliaInterface ...
      From worker 2:	[ Info: Compiling JuliaInterface ...
      From worker 7:	/home/oscarci-tester/ssd-data/ssd-runner-02/_work/_tool/julia/1.10.10/x64/bin/julia: symbol lookup error: /home/oscarci-tester/ssd-data/ssd-runner-02/_work/GAP.jl/GAP.jl/pkg/JuliaInterface/bin/x86_64-pc-linux-gnu-julia1.10-64-kv9/JuliaInterface.so: undefined symbol: gap_julia
Worker 7 terminated.
Unhandled Task ERROR: EOFError: read end of file
Stacktrace:
```

So, yeah, it tries to compile `JuliaInterface` multiple times in parallel. A simple fix would be to use `Pidfile.mkpidlock` in `build_JuliaInterface` (resp. wrap the call to `build_JuliaInterface` in it).

(However that would still end up re-compiling `JuliaInterface` multiple times. To protect against *that*, more work is needed -- but perhaps it is not necessary given that this should not affect release versions of GAP.jl, so it's just for our internal use?

_Originally posted by @fingolfin in https://github.com/oscar-system/GAP.jl/issues/1248#issuecomment-3326000386_

PR #1250, opened 2025-09-24
Add a pidlock for `build_JuliaInterface()`

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/1249.

This is more meant as a bandaid. Once this is merged, I'll open a new issue on the point of the redundant rebuilds as mentioned in #1249:
> (However that would still end up re-compiling `JuliaInterface` multiple times. To protect against *that*, more work is needed -- but perhaps it is not necessary given that this should not affect release versions of GAP.jl, so it's just for our internal use?

### DIFF
--- a/pkg/JuliaInterface/src/calls.c
+++ b/pkg/JuliaInterface/src/calls.c
@@ -230,6 +230,7 @@ Obj WrapJuliaFunc(jl_value_t * function)
     return func;
 }
 
+
 //
 //
 //


lgoettgens:
I know that this now unnecessarily touches the juliainterface c code, but since I want to merge https://github.com/oscar-system/GAP.jl/pull/1248 immediately afterwards anyway, I would just leave it in (instead of restarting CI once more).

PR #1257, opened 2025-09-30
In gap_to_julia_internal for Tuple, T === TT

fingolfin:
... so merge them into one; we use TT as that's what
the other related methods do.


ThomasBreuer (approved):
Good. Thanks.

Issue #940, opened 2023-09-18
Add a convenient wrapper for `GAP.Globals.Read(GapObj(joinpath(@__DIR__, "..", "gap", "exec.g")))` etc


fingolfin:
Julia code that wishes to read specific GAP files akin to Julia's `include` command currently has to use code like this:
```julia
GAP.Globals.Read(GapObj(joinpath(@__DIR__, "..", "gap", "exec.g")))
```
That's a bit unwieldy. Perhaps we should add a macro that makes this nice, like
```julia
GAP.@include("../gap/exec.g")
```
(similar to how we'd do `include("../gap/exec.jl")` if it was a Julia file instead).

The reason why I think it needs to be a macro is that `include` is special in so far as that Julia creates a new method for it in each new `module`, which has a "pointer" to that module, and so can e.g. compute paths relative to that module's source location. Our code doesn't have that ability, unless it uses a macro.

PR #1265, opened 2025-10-07
Add `GAP.@include` macro

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/940.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -53,6 +53,18 @@ const GAP_VERSION = VersionNumber(sysinfo["GAP_VERSION"])
 
 include("types.jl")
 
+"""
+    GAP.@include(path)
+
+Read and execute the GAP code in the file at `path`,
+which is interpreted as a relative path relative to the file where this macro is used.
+This is similar to julia's built-in `include` function,
+but for GAP code instead of Julia code.
+"""
+macro include(path)
+    return :(Wrappers.Read(GapObj(normpath(@__DIR__, $path))))
+end
+
 const last_error = Ref{String}("")
 
 const disable_error_handler = Ref{Bool}(false)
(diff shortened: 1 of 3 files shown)


ThomasBreuer (approved):
Looks good.

(Concerning the docstring:
Do we have the rule that the header line of the docstring of a not exported variable must contain the name of the module to which the variable belongs? If yes then the docstring of `@wrap` should get the header line `GAP.@wrap`.)

lgoettgens:
> (Concerning the docstring:
> Do we have the rule that the header line of the docstring of a not exported variable must contain the name of the module to which the variable belongs? If yes then the docstring of `@wrap` should get the header line `GAP.@wrap`.)

We don't have a rule, but I personally find it confusing if it has no qualification there but one needs it when trying to actually use the function.

src/GAP.jl
```diff
@@ -53,6 +53,18 @@ const GAP_VERSION = VersionNumber(sysinfo["GAP_VERSION"])
 
 include("types.jl")
 
+"""
+    GAP.@include(path)
+
+Read and execute the GAP code in the file at `path`,
+which is interpreted as a relative path relative to the file where this macro is used.
+This is similar to julia's built-in `include` function,
+but for GAP code instead of Julia code.
```
fingolfin, line 62: I was concerned whether `GAP.@include("/absolute/path/file.G")` works (if not we should warn users). But I think it will:
```
julia> normpath(@__DIR__, "file.jl")
"/Users/mhorn/Projekte/OSCAR/GAP.jl/file.jl"

julia> normpath(@__DIR__, "/tmp/file.jl")
"/tmp/file.jl"
```
Very nice!

fingolfin (approved):
nice

PR #1271, opened 2025-10-16
Fix expect test on julia nightly

lgoettgens:
This should fix the expect tests on nightly. See https://github.com/oscar-system/GAP.jl/pull/1269#issuecomment-3406548715 for an explanation.


fingolfin:
Does `julia --color=no` also turn off the colors? That might be easier?

PR #1276, opened 2025-10-22
Run downstream tests with gap 4.15.1

lgoettgens:


lgoettgens:
I am doing this on tutulla instead, once https://github.com/oscar-system/Oscar.jl/pull/5486 is merged

lgoettgens:
See https://github.com/oscar-system/Oscar.jl/pull/5374#issuecomment-3435602440 for the findings.

PR #1278, opened 2025-10-23
Version 0.16.0

lgoettgens:
or do you have anything else we should wait for @fingolfin @ThomasBreuer?


ThomasBreuer (approved):
If the test failures for some packages are  not a problem then I do not object.

PR #1286, opened 2025-11-23
Tweak some docstrings

fingolfin:
Partially inspired by https://github.com/tecosaur/CheckDoc.jl feedback.

### DIFF
--- a/src/GAP.jl
+++ b/src/GAP.jl
@@ -56,10 +56,11 @@ include("types.jl")
 """
     GAP.@include(path)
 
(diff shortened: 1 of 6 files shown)


lgoettgens:
> Argh wait, now it works on this PR after restarting it one more time??? That's a weird coincidence (but I'll take it)

My best guess is that this was some weird interaction of the cloudflare issues last week, together with some caching in github actions.
If you encounter this again, please let me know and I'll have a look then

PR #1287, opened 2025-11-23
Optimize conversion Symbol -> GapObj

fingolfin:
Avoid intermediate conversion to String, and thus one allocation.

Before:

    julia> @b GAP.Globals.GAPInfo.Version
    151.180 ns (3 allocs: 80 bytes)

    julia> @b GapObj(:Version)
    56.061 ns (3 allocs: 80 bytes)

After:

    julia> @b GAP.Globals.GAPInfo.Version
    144.766 ns (2 allocs: 48 bytes)

    julia> @b GapObj(:Version)
    27.758 ns (2 allocs: 48 bytes)

### DIFF
--- a/src/adapter.jl
+++ b/src/adapter.jl
@@ -213,7 +213,8 @@ function setbangindex!(x::GapObj, v::Any, i::Int64)
 end
 
 # records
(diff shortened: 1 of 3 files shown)


lgoettgens (approved):
Nice!

PR #1310, opened 2026-01-06
Simplify polymake handling

lgoettgens:
using a new BinaryWrappers.jl version that is capable of wrapping Perl scripts, see https://github.com/oscar-system/BinaryWrappers.jl/pull/7.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -61,7 +61,7 @@ NemoExt = "Nemo"
 [compat]
 AbstractAlgebra = "0.41.11, 0.42.1, 0.43, 0.44, 0.45, 0.46, 0.47"
 Artifacts = "1.10"
-BinaryWrappers = "0.1.4"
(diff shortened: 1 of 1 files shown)


benlorenz (approved):
Looks good, did you check whether the polymaking gap distro tests also work on macos?

lgoettgens:
> > > Looks good, did you check whether the polymaking gap distro tests also work on macos?
> > 
> > 
> > Running `GAP.Packages.test("polymaking")` with this branch on munk succeeds. (Only small downside is that there is a `perl` in the PATH)
> 
> Does this perl binary cause problems? Maybe we could add the path to the correct perl version to the PATH in the wrapper script?

I don't think that is needed. I just wanted to mention it here, as we would not notice if the system perl instead of the Perl_jll perl would be used. But I honestly see no way how that could happen anyway

PR #1312, opened 2026-01-07
Add script for running JET with options

fingolfin:
Specifically, don't report anythin related to our unholy
is_loaded_directly hack. We know it's evil, no need to see
JET get twisted up about it.

Similar scripts could / should be add to our other packages,
with possible more things to ignore added for each, as needed.

The `AnyFrameMethod` could also perhaps be submitted to JET,
dunno if there'd be interest.

### DIFF
--- a/etc/jet.jl
+++ b/etc/jet.jl
@@ -1,0 +1,23 @@
+using Revise
+using JET
+
+struct AnyFrameMethod <: ReportMatcher
+    m::Union{Function,Method,Symbol}
+end
+
+function JET.match_report(matcher::AnyFrameMethod, @nospecialize(report::JET.InferenceErrorReport))
+    # check all VirtualFrames in the VirtualStackTrace for a match to the specified method
+    m = matcher.m
+    if m isa Symbol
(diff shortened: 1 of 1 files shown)


lgoettgens (approved):
If it helps you, please add it. I haven't looked at JET reports in a while.

lgoettgens:
Side-remark: When looking at a JET report, I just got two potential issues related to `is_loaded_directly`. One of them will be fixed by https://github.com/Nemocas/AbstractAlgebra.jl/pull/2277, the other one is something genuine in julia Base (about `show(io::IOContext{IOBuffer}, frame::Base.StackTraces.StackFrame)`). Therefore, I would be interested if there are any more popping up for you (and with which combination of versions of julia and packages).

lgoettgens:
`AnyFrameMethod` has been added to JET.jl in https://github.com/aviatesk/JET.jl/pull/798, which was released last week as part of https://github.com/aviatesk/JET.jl/releases/tag/v0.11.4. Therefore, I think we can close this PR.

PR #1313, opened 2026-01-07
Add with_info_level helper

fingolfin:
This resolves some JET warnings. But I think it's useful on its own.

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -280,7 +280,7 @@ function install(spec::String, version::String = "";
     Pidfile.mkpidlock("$pkgdir.lock"; stale_age=300) do
       with_info_level(Globals.InfoPackageManager, quiet ? 0 : debug ? 3 : nothing) do
         if version == ""
-          return Globals.InstallPackage(GapObj(spec), interactive; keepDirectory=debug))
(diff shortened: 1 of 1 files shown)


src/packages.jl
```diff
@@ -370,13 +371,7 @@ function remove(spec::String; interactive::Bool = true, quiet::Bool = false,
...
 
-    if quiet || debug
-      oldlevel = Wrappers.InfoLevel(Globals.InfoPackageManager)
-      Wrappers.SetInfoLevel(Globals.InfoPackageManager, quiet ? 0 : 3)
-      res = Globals.RemovePackage(GapObj(spec), interactive; debug)
-      Wrappers.SetInfoLevel(Globals.InfoPackageManager, oldlevel)
-      return res
-    else
+    with_info_level(Globals.InfoPackageManager, quiet ? 0 : debug ? 3 : nothing) do
       return Globals.RemovePackage(GapObj(spec), interactive)
```
lgoettgens:
```suggestion
      return Globals.RemovePackage(GapObj(spec), interactive; debug)
```
same here with line 376

ThomasBreuer (approved):
Yes, looks good.
Thanks.

PR #1331, opened 2026-01-30
CI: track code coverage in 'expect' test

fingolfin:


fingolfin:
Ugh seems that coverage tracking messes with the expect tests. But why? *sigh*

PR #1332, opened 2026-01-30
Remove GAP.Display and StringDisplayObj

fingolfin:
Nothing is calling them and they are undocumented.

Right now `GAP.Display` also doesn't allow specifying custom `io::IO`. So it is not clear what we gain over just calling `GAP.Globals.Display` ?

Alternatively we could keep them but make them discoverable, e.g. by adding documentation, or replacing `GAP.Display` with a method for a standard function(perhaps `show` with a special mimetype, or as a method to Julia's `display`, or by some other means).

@ThomasBreuer any thoughts on this one?

### DIFF
--- a/pkg/JuliaInterface/gap/utils.gi
+++ b/pkg/JuliaInterface/gap/utils.gi
@@ -30,17 +30,6 @@ BindGlobal( "SetBangPosition",
     obj![ pos ]:= val;
   end );
 
-BindGlobal( "StringDisplayObj",
-function(obj)
-  local str, out;
-  str := "";
-  out := OutputTextString(str, false);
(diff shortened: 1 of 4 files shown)


ThomasBreuer:
`GAP.Display` is not interesting.
`StringDisplayObj` is to some extent what one would expect from GAP's `DisplayString`, but GAP.jl is not the right place for it.
(For historical reasons, GAP's `DisplayString` is not supported well.)

What would be the right Julia syntax corresponding to GAP's `Display`, that is, one wants to show a formatted, human-readable screen output for the given GAP object?

fingolfin:
I agreed with @ThomasBreuer in a video call that we can indeed remove this.

Issue #1330, opened 2026-01-29
Improve code coverage for `GAP.Packages`


fingolfin:
We don't have any tests at all for
- `GAP.Packages.build`
- `GAP.Packages.build_recursive `
- `GAP.Packages.test`
- `GAP.Packages.update`

That's bad because even a trivial mistake in those right at the start won't be discovered by CI.

For `GAP.Packages.install` and `GAP.Packages.load` we have some test but a substantial part of those functions is not covered at all.

@ThomasBreuer perhaps you can look into that? It doesn't have to be perfect, but at least some minimal tests for those would be nice.

lgoettgens:
Somehow related: https://github.com/oscar-system/GAP.jl/issues/574 and https://github.com/oscar-system/GAP.jl/pull/584

PR #1334, opened 2026-01-30
Add some tests for `GAP.Packages` functions

ThomasBreuer:
addresses #1330

The nontrivial bit is to call `GAP.Packages.build` in a way that the function does something.
The package in question must claim to be not available, thus one cannot simple take a loaded package or a package that has been installed with `GAP.Packages.install`. The idea is to modify the information in the info records temporarily such that GAP thinks the package is not available.
(This must be done package by package, not once before a loop, because PackageManager calls `InitializePackagesInfoRecords`, which overwrites the changes in the info records.)

### DIFF
--- a/test/packages.jl
+++ b/test/packages.jl
@@ -71,14 +71,19 @@
       # build the package and its dependencies (may do nothing
       # if they already are built)
       @test GAP.Packages.build_recursive(pkg[:name])
-      # Make sure that GAP stores package information.
+      # `GAP.Packages.build` shall try to build the package.
+      # For that, the package must be regarded as
+      # - not yet loaded (stored in `GAP.Globals.GAPInfo.PackagesLoaded`
+      # - and not available (according to the `AvailabilityTest` function
+      #   in the `GAP.Globals.PackageInfo` records for the package).
+      # First we load the package, then we need not distinguish later
+      # whether the package was loaded or not.
       GAP.Packages.with_info_level(GAP.Globals.InfoPackageLoading, 4) do
         @test GAP.Packages.load(pkg[:name]; quiet=false)
       end
       # Manipulate GAP's global information such that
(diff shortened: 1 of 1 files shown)


test/packages.jl
```diff
@@ -49,4 +49,48 @@
     # - a nonexisting path
     path = path * "xxx"
     @test ! GAP.Packages.load(path)
+
+    # Run package tests.
+    @test ! GAP.Packages.test("no_such_package")
```
fingolfin: Actually, I'd expect this to throw an exception (the input arguments were invalid). The values `true` / `false` should be reserved for "package was loaded, teste were run and passed / failed"

test/packages.jl
```diff
@@ -49,4 +49,48 @@
...
+    @test GAP.Packages.update("fga", interactive = false)
+    @test GAP.Packages.remove("fga", interactive = false)
+
+    # Test building a package.
+    # For that, we choose two packages with kernel extensions,
+    # such that one is a needed package of the other.
+    pkgs = map(name -> Dict{Symbol, Any}(:name => name), ["orb", "genss"])
+    for pkg in pkgs
+      # Make sure that GAP stores package information.
+      @test GAP.Packages.load(pkg[:name])
```
fingolfin: Let's get some more verbosity to see if that gives a hint as to what is wrong...
```suggestion
     with_info_level(Globals.InfoPackageLoading, 4) do
      @test GAP.Packages.load(pkg[:name]; quiet=false)
     end
```

Actually might be a good idea to add a `debug` kwarg to `GAP.Packages.load` which activates such a `with_info_level` in it?

fingolfin (approved):
Seems to pass now, so fine from my POV. However, I edited this PR, so @ThomasBreuer should check if the current version makes sense to him as well.

PR #1350, opened 2026-02-11
Avoid some boxing

lgoettgens:


src/packages.jl
```diff
@@ -497,26 +497,26 @@ function test(name::String)
     end
   end
 
-  error_occurred = false
-  called_QuitGap = false
+  error_occurred = Ref(false)
+  called_QuitGap = Ref(false)
```
fingolfin:
```suggestion
  error_occurred::Bool = false
  called_QuitGap::Bool = false
```
is not sufficient to fix this?

src/packages.jl
```diff
@@ -529,7 +529,7 @@ function test(name::String)
   # Due to the hack above, we run into an error in TestPackage that is usually unreachable.
   # In the case of a `QuitGap` call, we thus don't check for `result == true`.
   # Note: `result` may be `fail`, so it is not always booleany.
```
fingolfin:
```suggestion
```

fingolfin (commented):
Can we perhaps just get rid of `result`?

lgoettgens:
> Can we perhaps just get rid of `result`?

There are various gap packages now failing, so I would suggest to revert https://github.com/oscar-system/GAP.jl/pull/1350/commits/fcacb1b96f01d92ac493cf713723858a918a2f82
I don't quite remember which pattern of gap package tests needed this, but the failing ace, browse, edim etc. Seem to be witnesses that the `result` handling was broken in that commit

src/packages.jl
```diff
@@ -497,26 +497,27 @@ function test(name::String)
...
+  error_occurred = Ref(false)
+  called_QuitGap = Ref(false)
   function fake_QuitGap(code)
-    called_QuitGap && return # only do something on the first call
-    called_QuitGap = true
+    called_QuitGap[] && return # only do something on the first call
+    called_QuitGap[] = true
     if (code isa Bool && !code) || (code isa Int && code % 256 != 0)
-      error_occurred = true
+      error_occurred[] = true
```
fingolfin: What we'd really want is to abort here (e.g. by throwing an exception). But I am afraid our GAP-Julia error bridging is still too brittle for that. But I hope this can be improved in the future.

src/packages.jl
```diff
@@ -497,26 +497,27 @@ function test(name::String)
...
-  result = false
   try
     with_gap_var(:ERROR_OUTPUT, Globals._JULIAINTERFACE_ORIGINAL_ERROR_OUTPUT) do
       with_gap_var(:QuitGap, fake_QuitGap) do
         with_gap_var(:QUIT_GAP, fake_QuitGap) do
           with_gap_var(:ForceQuitGap, fake_QuitGap) do
             with_gap_var(:FORCE_QUIT_GAP, fake_QuitGap) do
               result = Globals.TestPackage(GapObj(name))
+              # Note: `result` may be `fail`, so it is not always booleany.
+              return !error_occurred && (called_QuitGap || result == true)
```
lgoettgens:
```suggestion
              return !error_occurred[] && (called_QuitGap[] || result == true)
```

lgoettgens:
Ace is still unhappy. Thus I would suggest to revert all your changes to Packages.test (as it worked before that and just was a mechanical change), and we can continue playing around with that function in a new PR

PR #1361, opened 2026-03-13
Greatly improve stack trace shown on GAP errors

### LINKED ISSUE #1168: `GAP.evalstr` hiding some GAP errors
```julia
julia> GAP.evalstr("2^100/0")
ERROR: Error thrown by GAP: 
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] evalstr(cmd::String)
   @ GAP ~/code/julia/GAP.jl/src/ccalls.jl:149
 [3] top-level scope
   @ REPL[10]:1
```
This suggests that there should be some GAP error message after the `:`.

When doing the same thing in slightly different ways (without using `evalstr`), the GAP error message is shown:
```julia
julia> GAP.GapObj(BigInt(2)^100) / 0
ERROR: Error thrown by GAP: Error, Rational operations: <divisor> must not be zero
not in any function at *defin*:0
[...]

julia> GAP.prompt()
gap> 2^100/0;
Error, Rational operations: <divisor> must not be zero
not in any function at *stdin*:1
type 'quit;' to quit to outer loop
```
maybe related to https://github.com/oscar-system/GAP.jl/issues/896

fingolfin:
In the GAP error handler, we now also capture the Julia stacktrace, and replace the GAP stacktrace rendering (via GAP's `Where` function) by a custom stacktrace rendering.

This prints both a nice GAP stacktrace (where GAP stack frames are captured in a custom `GAPError` type), but also a regular Julia backtrace leading up to the point we entered GAP.

Also modify `evalstr` and how it captures errors to align with the new feature, so GAP errors and warnings no longer clobber each other during reporting.

On macOS, this requires GAP_jll 400.1500.101 which was re-compiled using GCC (the previous builds where using clang on macos, which lead to serious problems for Julia's stack unwinding code).

One situation that is not (yet?) handled right is complex chaining between Julia and GAP (think Julia calls GAP calls Julia calls ...); we only print one GAP and one Julia stacktrace.

---

Draft because I am not yet happy with the code the AI produced. But I figured I should put this out there for further feedback and experiments.

Here is a mini example. Before:
```julia-repl
julia> gap_func(n) = GAP.Globals.SymmetricGroup(n)
gap_func (generic function with 1 method)

julia> gap_func(-3)
ERROR: Error thrown by GAP: Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 3rd choice method found for `SymmetricGroupCons' on 2 arguments at /Users/mhorn/.julia/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/lib/methsel2.g:250 called from
SymmetricGroupCons( IsPermGroup, arg[1] ) at /Users/mhorn/.julia/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/grp/basic.gd:734 called from
<function "SymmetricGroup">( <arguments> )
 called from read-eval loop at *defin*:0

Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:44
 [2] ThrowObserver(depth::Int32)
   @ GAP ~/Projekte/OSCAR/GAP.jl/src/GAP.jl:99
 [3] _call_gap_func
   @ ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:357 [inlined]
 [4] (::GapObj)(a1::Int64)
   @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:334
 [5] gap_func(n::Int64)
   @ Main ./REPL[3]:1
 [6] top-level scope
   @ REPL[4]:1
```

Note how this gives us no clue at all where the call into GAP happened. Of course in this contrieved example it is obvious, but this can also happen deep in the innards of OSCAR.

After this patch is applied, things are much nicer:
```julia-repl
julia> gap_func(n) = GAP.Globals.SymmetricGroup(n)
gap_func (generic function with 1 method)

julia> gap_func(-3)
ERROR: Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
GAP stacktrace:
 [1] HANDLE_METHOD_NOT_FOUND @ ~/.julia/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/lib/methsel2.g:250
 [2] SymmetricGroup @ ~/.julia/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/grp/basic.gd:734
Julia stacktrace:
 [1] gap_func(n::Int64)
   @ Main ./REPL[2]:1
 [2] top-level scope
   @ REPL[3]:1
```


fingolfin:
One of the errors is:
```
[ Info: Doctest: running doctests.
┌ Error: doctest failure in ../src/globals.jl:23-46
│ 
│ ```jldoctest
│ julia> GAP.Globals.Size    # a global GAP function
│ GAP: <Attribute "Size">
│ 
│ julia> GAP.Globals.size    # there is no GAP variable with this name
│ ERROR: GAP variable size not bound
│ [...]
│ 
│ julia> hasproperty( GAP.Globals, :size )
│ false
│ 
│ julia> GAP.Globals.size = 17;
│ 
│ julia> hasproperty( GAP.Globals, :size )
│ true
│ 
│ julia> GAP.Globals.size
│ 17
│ 
│ julia> GAP.Globals.UnbindGlobal(g"size")
│ 
│ julia> GAP.Globals.Julia   # Julia objects can be values of GAP variables
│ Main
│ 
│ ```
│ 
│ Subexpression:
│ 
│ GAP.Globals.UnbindGlobal(g"size")
│ 
│ Evaluated output:
│ 
│ Syntax error: ; expected in stream:1
│ Error, Rational operations: <divisor> must not be zero
│ not in any function at stream:1
│ 
│ Expected output:
│ 
│ 
│ 
│   diff =
│    Syntax error: ; expected in stream:1
│    Error, Rational operations: <divisor> must not be zero
│    not in any function at stream:1
└ @ Documenter ~/work/GAP.jl/GAP.jl/src/globals.jl:23
```

So that looks like a real bug in the code.

fingolfin:
I've rebased this and added a commit made by Codex to remove use of various internal Julia APIs that we shouldn't be using (and that in fact are gone/changed in Julia nightly).

The mysterious error above is still there. Worse: in different Julia versions, different but similar failures crop up; and I can't repro it locally -- making me wonder if this is a memory corruption issue or so. Hm.

I'll try to investigate as much as my kid let's me today (no Kindergarten)

fingolfin:
Ah it works in Julia 1.12 but with 1.10 and 1.11 I can also reproduce it locally

lgoettgens:
What is the state of this? I am happy to give this a review, but it's marked as a draft. Is there something that you want/need to adapt to with the new gap 4.16, i.e. with https://github.com/gap-system/gap/pull/6263 being available?

fingolfin:
It was mainly a draft because some CI was still failing; plus I did not have time to review everything *myself* (large parts are written by the AI, after all), and I did not want to let others suffer through this without warning.

Regarding adaptation to GAp 4.16.0: yeah, there is probably some potential for simplification. But this could also be done in a follow-up (and it might even make sense to add additional helpers to GAP to allow further simplification).

If you are willing to have a closer look, that'd be appreciated, but be warned, there may be slop in here :-/

.github/workflows/gap.yml
```diff
@@ -71,17 +71,30 @@ jobs:
...
           mv GAPROOT /tmp/GAPROOT
           cd /tmp/GAPROOT
           ./autogen.sh
-          ./configure --enable-Werror
+          ./configure
           make -j`nproc`
       - name: "Override bundled GAP"
         run: |
-          julia --project etc/setup_override_dir.jl /tmp/GAPROOT /tmp/gap_jll_override --enable-Werror
+          julia --project etc/setup_override_dir.jl /tmp/GAPROOT /tmp/gap_jll_override
```
lgoettgens, line 95: seems unrelated?

src/ccalls.jl
```diff
@@ -100,8 +100,7 @@ GAP: [  ]
 ```
 """
 function evalstr_ex(cmd::String)
-    res = @ccall libgap.GAP_EvalString(cmd::Cstring)::GapObj
-    return res
+    return @ccall libgap.GAP_EvalString(cmd::Cstring)::GapObj
```
lgoettgens: change seems unrelated

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+        label = String(raw_frame[1])
+        file = raw_frame[2] == Globals.fail ? nothing : String(raw_frame[2])
+        line = raw_frame[3] == Globals.fail ? nothing : Int(raw_frame[3])
+        push!(frames, GAPStackFrame(label, file, line))
+    end
+    return frames
+end
+
+function clear_gap_error_frames()
+    hasproperty(Globals, :_JULIAINTERFACE_CLEAR_ERROR_STACK) || return
```
lgoettgens:
```suggestion
```
this should not happen (similar for the other similar places)

gap/err.g
```diff
@@ -9,13 +9,89 @@
...
+
+    # In standalone mode / gap.sh, GAP.jl deliberately disables all custom
+    # Julia-side error handling and lets GAP behave natively.
+    if not Julia.GAP.is_error_handler_disabled() then
+        MakeReadWriteGlobal("_JULIAINTERFACE_ERROR_STACK");
+        _JULIAINTERFACE_ERROR_STACK := _JULIAINTERFACE_CAPTURE_STACK(options.context);
+        MakeReadOnlyGlobal("_JULIAINTERFACE_ERROR_STACK");
+    fi;
+
+    orig := VALUE_GLOBAL("_ORIG_ErrorInner");
```
lgoettgens: I don't like that this relies on the internal of `ReplaceBinding` on how the original value is named.

gap/err.g
```diff
@@ -9,13 +9,89 @@
...
+#
+# TODO: Revisit this for future GAP releases. If GAP eventually exposes a
+# stable callback-time accessor for the original error context, this ErrorInner
+# override may become unnecessary.
+ReplaceBinding("ErrorInner", function(options, earlyMessage)
+    local orig;
+
+    # In standalone mode / gap.sh, GAP.jl deliberately disables all custom
+    # Julia-side error handling and lets GAP behave natively.
+    if not Julia.GAP.is_error_handler_disabled() then
```
lgoettgens:
```suggestion
    if not GAP_jl.is_error_handler_disabled() then
```

src/ccalls.jl
```diff
@@ -147,19 +146,28 @@ i.e., it can be accessed as `GAP.Globals.x` after the call of `GAP.evalstr`,
...
-    copy_gap_error_to_julia()
+    res = evalstr_ex(cmd * ";")
 
-    msg = get_and_clear_last_error()
+    snapshot = take_gap_error_snapshot()
     if any(x::GapObj->x[1] === false, res)
-      error("Error thrown by GAP: $msg")
-    elseif !isempty(msg)
+      snapshot === nothing && copy_gap_error_to_julia()
+      snapshot === nothing && (snapshot = take_gap_error_snapshot())
```
lgoettgens: this looks overly complicated. In particular, I do not understand why we call `take_gap_error_snapshot` a second time here

src/ccalls.jl
```diff
@@ -147,19 +146,28 @@ i.e., it can be accessed as `GAP.Globals.x` after the call of `GAP.evalstr`,
...
+      throw_gap_error(snapshot)
+    end
+
+    # If there is a warning string on the GAP side, copy it into a snapshot.
+    # We do this only if no error snapshot has been captured already, to avoid
+    # clobbering the callback-based data used above.
+    snapshot === nothing && copy_gap_error_to_julia()
+    snapshot === nothing && (snapshot = take_gap_error_snapshot())
```
lgoettgens: same here

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+        print(io, ": ", err.message)
+    end
+
+    if !isempty(err.gap_frames)
+        print(io, "\nGAP stacktrace:")
+        for (i, frame) in pairs(err.gap_frames)
+            print(io, "\n [", i, "] ")
+            printstyled(io, frame.function_label; bold=true)
+            if frame.file !== nothing
+                print(io, " @ ", Base.contractuser(frame.file))
```
lgoettgens:
```suggestion
                print(io, "\n @ ", Base.contractuser(frame.file))
```
to match the two-line style from julia stacktraces

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+                if frame.line !== nothing
+                    print(io, ":", frame.line)
+                end
+            elseif frame.line !== nothing
+                print(io, " @ line ", frame.line)
```
lgoettgens:
```suggestion
                print(io, "\n @ line ", frame.line)
```
to match the two-line style from julia stacktraces

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+# stack traces we actually see around GAP errors on supported platforms:
+# - ccalls.jl / GAP.jl are GAP.jl's own dispatch and initialization plumbing
+# - .dylib / .so / .dll frames are native library frames (libjulia, libgap,
+#   JuliaInterface, etc.)
+# - .c / .h frames are native source locations that may appear when unwinding
+#   through GAP or libjulia internals
+# - file == ":-1" and line <= 0 are synthetic / locationless frames that do not
+#   point at useful source code
+# This filter is intentionally conservative: if a frame already has a clear
+# Julia source location outside GAP.jl internals, keep it.
```
lgoettgens: I don't quite agree with the AI that this in "conservative". I would much prefer a small-ish list of things here. But that would certainly be something that can be changed later on, and that I need to play around with in a working system to build an opinion.

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+    file == ":-1" && return true
+    frame.line <= 0 && return true
+    return false
+end
+
+# Keep the top-level Julia frame, but drop everything above it. Those outer
+# frames are typically REPL / include / test harness machinery and make the
+# rendered backtrace noisier without helping to locate the user call site.
+function trim_julia_backtrace_at_toplevel(trace::Vector{Tuple{Base.StackTraces.StackFrame,Int}})
+    for (i, (frame, _)) in pairs(trace)
```
lgoettgens:
```suggestion
    for (i, (frame, _)) in enumerate(trace)
```
the same due to `trace` being a proper `Vector`, but `resize!` accepts a size and not an index.

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+    return flag
+end
+
+# GAP's raw error text often contains prefixes such as "Error, " and extra
+# continuation lines. GAPError.message keeps only the first user-facing summary
+# line, while GAPError.raw_text preserves the full text for display fallback.
+function gap_error_message(raw_text::String)
+    isempty(raw_text) && return ""
+
+    first_line = split(chomp(raw_text), '\n'; limit=2)[1]
```
lgoettgens: whatever this does 😂 
In particular, the `chomp` confuses me

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+    if isempty(raw_text) && isempty(frames)
+        return nothing
+    end
+
+    return GAPError(
+        gap_error_message(raw_text),
+        frames,
+        raw_text,
+        capture_current_julia_backtrace(),
+    )
```
lgoettgens, line 268: why doesn't this function call `clear_gap_error` in the end?

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+    last_error_snapshot[] = GAPError(
+        gap_error_message(raw_text),
+        frames,
+        raw_text,
+        capture_current_julia_backtrace(),
+    )
+    clear_gap_error()
+end
+
+function capture_gap_error_snapshot_from_runtime()
```
lgoettgens: 95% of this function is identical to the one above. Would it be possible to merge them?

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+    # Snapshots are single-use. Once a GAP error has been consumed, later calls
+    # must not see it again.
+    last_error_snapshot[] = nothing
+    return snapshot
+end
+
+function throw_gap_error(snapshot::Union{Nothing,GAPError})
+    if snapshot === nothing
+        snapshot = capture_gap_error_snapshot_from_runtime()
+        snapshot === nothing || clear_gap_error()
```
lgoettgens:
```suggestion
```
as noted elsewhere, the clear should be in `capture_gap_error_snapshot_from_runtime` instead

src/errors.jl
```diff
@@ -9,23 +9,287 @@
...
+    snapshot = last_error_snapshot[]
+    # Snapshots are single-use. Once a GAP error has been consumed, later calls
+    # must not see it again.
+    last_error_snapshot[] = nothing
+    return snapshot
+end
+
+function throw_gap_error(snapshot::Union{Nothing,GAPError})
+    if snapshot === nothing
+        snapshot = capture_gap_error_snapshot_from_runtime()
```
lgoettgens: I would prefer if not `throw_gap_error` would need to find the error it wants to throw, but instead its callsite would do that.

src/GAP.jl
```diff
@@ -254,7 +255,9 @@ function __init__()
     # Start GAP.
     initialize(cmdline_options)
 
-    if !isdefined(Main, :__GAP_ARGS__)
+    if isdefined(Main, :__GAP_ARGS__)
+        set_error_handler_disabled(true)
```
lgoettgens: why does `__init__` need both this and the change in line 77?

src/prompt.jl
```diff
@@ -48,12 +46,16 @@ function prompt()
...
 
     # restore GAP.jl error handler
-    disable_error_handler[] = false
+    set_error_handler_disabled(false)
     replace_global!(:ERROR_OUTPUT, Globals._JULIAINTERFACE_ERROR_OUTPUT)
 end
 
 # helper function for `gap.sh` scripts created by create_gap_sh()
 function run_session()
+    original_error_output = ValueGlobalVariable("_JULIAINTERFACE_ORIGINAL_ERROR_OUTPUT")
```
lgoettgens: why is this needed? I thought that when running via gap.sh the error handling is left completely to GAP?

test/basics.jl
```diff
@@ -32,7 +32,7 @@ using IOCapture
     @test x.a == 1
 
     xx = GapObj([1, 2, 3])
-    @test_throws ErrorException xx[4]
+    @test_throws GAP.GAPError xx[4]
```
lgoettgens: I think `GAPError` should be exported, so this is just `GAPError` here.

test/errors.jl
```diff
@@ -0,0 +1,248 @@
...
+##  This file is part of GAP.jl, a bidirectional interface between Julia and
+##  the GAP computer algebra system.
+##
+##  Copyright of GAP.jl and its parts belongs to its developers.
+##  Please refer to its README.md file for details.
+##
+##  SPDX-License-Identifier: LGPL-3.0-or-later
+##
+
+using IOCapture
```
lgoettgens, line 12: this package is not mentioned in the Project.toml...

test/errors.jl
```diff
@@ -0,0 +1,248 @@
...
+    try
+        f()
+        return nothing
+    catch err
+        return err
+    end
+end
+
+gap_error_type() = isdefined(GAP, :GAPError) ? getfield(GAP, :GAPError) : Exception
+gap_frame_type() = isdefined(GAP, :GAPStackFrame) ? getfield(GAP, :GAPStackFrame) : Any
```
lgoettgens: these two seem useless to me. The `isdefined` should always be true, and then the body can be manually inlined.

test/errors.jl
```diff
@@ -0,0 +1,248 @@
...
+
+message_of(err) = hasproperty(err, :message) ? getproperty(err, :message) : ""
+gap_frames_of(err) = hasproperty(err, :gap_frames) ? getproperty(err, :gap_frames) : Any[]
+
+frame_label(frame) = getproperty(frame, :function_label)
+frame_file(frame) = getproperty(frame, :file)
+frame_line(frame) = getproperty(frame, :line)
```
lgoettgens: Is there a reason to use `getproperty` over `.`? I thought the latter optimizes better.

src/errors.jl
lgoettgens: Instead of just calling `contractuser` for simplifying paths, what about using the same code as in `GAP.Packages.versioninfo` (https://github.com/oscar-system/GAP.jl/blob/fbc9a6234ad447d5c934527c5f1fecab5226c740/src/packages.jl#L570) to also simplify artifact paths.

lgoettgens (commented):
A bunch of comments. Some are suggestions for improvements, some are my missing understanding of how things work (or just being confused by stupid AI things).

.github/workflows/gap.yml
fingolfin: I can't compile GAP locally without that. The relevant commit message in this PR is:
```
no Werror to avoid complaints about readline headers
```
So on the one hand it is unrelated, but on the other, it is necessary for me to use. We can factor it out into a separate PR, but that's extra work of course

src/errors.jl
fingolfin: IIRC it *does* happen during startup. But then of course there should at least be a comment explaining just that! I'll look into it

gap/err.g
fingolfin: Neither do I. Do you have a suggestion for an alternative?

lgoettgens: Do it explicitly (like in https://github.com/oscar-system/GAP.jl/pull/1085/changes#diff-fd5ae88aa7e6b473f49bdde6cc0ad4d4190f7e92f8994d29d1218e70a2c11fe5) instead of calling `ReplaceBinding`. Or alternatively, add a function `ReplaceBindingOrigName(name)` that does
```
orig_name := "_ORIG_";
APPEND_LIST_INTR(orig_name, name);
return orig_name
```
and then use that both in `ReplaceBinding` and here (i.e., adding an API guarantee to `ReplaceBinding`).

src/errors.jl
lgoettgens: Hmm, would it then maybe be possible to have a global variable `in_startup::Bool` or something instead that disables all of these functions? That could make it a bit easier to understand, and more consistent, as all the functions would have the same guard.

test/errors.jl
```diff
@@ -0,0 +1,248 @@
...
+        @test !occursin("throw_gap_error", shown)
+        @test !occursin("call_gap_func_nokw", shown)
+        @test occursin("top-level scope", shown)
+        top_level = findfirst("top-level scope", shown)
+        @test top_level !== nothing
+        @test !occursin("eval(m::Module", shown[first(top_level):end])
+    end
+
+    @testset "disabling custom error handling clears stale state" begin
+        @test GAP.evalstr("Julia.GAP.is_error_handler_disabled()") === false
```
lgoettgens:
```suggestion
        @test GAP.evalstr("GAP_jl.is_error_handler_disabled()") === false
```

furthermore, why does this do the gap-julia-loop, and not just

```suggestion
        @test GAP.is_error_handler_disabled() === false
```
?

test/errors.jl
```diff
@@ -0,0 +1,248 @@
...
+
+        err = capture_exception() do
+            GAP.Globals.gapjl_traceback_outer()
+        end
+        @test err isa gap_error_type()
+        @test !isempty(gap_frames_of(err))
+
+        GAP.set_error_handler_disabled(true)
+        try
+            @test GAP.evalstr("Julia.GAP.is_error_handler_disabled()") === true
```
lgoettgens: same

test/errors.jl
```diff
@@ -0,0 +1,248 @@
...
+            @test GAP.take_gap_error_snapshot() === nothing
+            @test isempty(GAP.capture_gap_error_frames())
+            @test length(GAP.Globals._JULIAINTERFACE_ERROR_STACK) == 0
+            @test isempty(String(GAP.Globals._JULIAINTERFACE_ERROR_BUFFER))
+        finally
+            GAP.set_error_handler_disabled(false)
+        end
+
+        @test GAP.evalstr("Julia.GAP.is_error_handler_disabled()") === false
```
lgoettgens: same

src/errors.jl
fingolfin: I am not sure if another global variable will *simplify* anything, as we then need extra logic to set it. But we could add a helper function, say `is_startup_done() = hasproperty(Globals, :_JULIAINTERFACE_CLEAR_ERROR_STACK)`, with a comment; and then call that in various places.

lgoettgens: I like that

test/errors.jl
fingolfin: But it is in `test/Project.toml` ?

lgoettgens: Indeed... must have missed it :sweat_smile:

src/errors.jl
fingolfin: I'd prefer not to do this: when I want to look at the code, I need the full path; in fact in my terminal emulator, I can just cmd-click on the path and it'll open in my editor. This still works after `contractuser`; but if we replace the SHA in the path by a fixed string, it won't; and having to figure out the SHA to restore it is a PITA.

src/prompt.jl
fingolfin: it isn't needed, the AI was confused because the file is named prompt.jl, but after I explained the situation it conceded 😂

src/errors.jl
fingolfin: yeah that is misleading. changed the comment

src/errors.jl
fingolfin: Added a comment to explain this, please check if it clears up things

src/errors.jl
fingolfin: I've simplified this and some of the callers now

src/errors.jl
fingolfin: I've refactored the code and `capture_gap_error_snapshot_from_runtime` is in fact gone now

src/errors.jl
lgoettgens: No, it does not. Removing the chomp should do the same thing. If I just take the first line, it does not matter if I drop any *trailing* newlines

.github/workflows/gap.yml
```diff
@@ -71,17 +71,30 @@ jobs:
...
         run: |
-          brew install autoconf zlib # gmp pkg-config
+          # - autoconf, gmp, zlib are needed by GAP's build system
+          # - coreutils provides nproc
+          # - force use of Homebrew gcc to build GAP for proper stacktrace support
+          #   Homebrew installs versioned binaries such as gcc-15 and g++-15,
+          #   while later scripts expect unversioned gcc and g++ in PATH.
+          brew install autoconf gmp zlib gcc coreutils
+
+          # ensure gcc and g++ are avaialble under those names (on macOS,
```
lgoettgens:
```suggestion
          # ensure gcc and g++ are available under those names (on macOS,
```

src/errors.jl
```diff
@@ -9,23 +9,299 @@
...
+        gap_error_message(raw_text),
+        frames,
+        raw_text,
+        capture_current_julia_backtrace(),
+    )
+end
+
 function copy_gap_error_to_julia()
     global disable_error_handler
     if disable_error_handler[]
```
lgoettgens:
```suggestion
    if is_error_handler_disabled()
```

src/errors.jl
```diff
@@ -34,13 +310,14 @@ function ThrowObserver(depth::Cint)
         return
```
lgoettgens:
```suggestion
    if is_error_handler_disabled()
         return nothing
```

src/errors.jl
lgoettgens:
```suggestion
    first_line = first(split(raw_text, '\n'; limit=2))
```

.github/workflows/gap.yml
```diff
@@ -71,17 +71,30 @@ jobs:
...
+          # these names are aliases for clang / clang++ out of the box)
+          shimdir="${HOME}/bin"
+          mkdir -p "${shimdir}"
+          ln -sf "$(brew --prefix)/bin/gcc-15" "${shimdir}/gcc"
+          ln -sf "$(brew --prefix)/bin/g++-15" "${shimdir}/g++"
```
lgoettgens: my AI agent complains that these lines will break one the CI runner updates its internal gcc version, and suggest something along the lines of `ls "$(brew --prefix)/bin"/gcc-*`. I am not sure if this is actually necessary here, but I wanted to mention it.

test/errors.jl
```diff
@@ -0,0 +1,238 @@
...
+            gapjl_traceback_inner();
+        end;;
+        """)
+
+        err = capture_exception() do
+            GAP.Globals.gapjl_traceback_outer()
+        end
+
+        @test err !== nothing
+        @test err isa GAPError
```
lgoettgens:
```suggestion
        @test err isa GAPError
```
this is redundant (but no strong preference to remove it)

also applies to more places further down

test/errors.jl
```diff
@@ -0,0 +1,238 @@
...
+    try
+        f()
+        return nothing
+    catch err
+        return err
+    end
+end
+
+message_of(err) = hasproperty(err, :message) ? err.message : ""
+gap_frames_of(err) = hasproperty(err, :gap_frames) ? err.gap_frames : Any[]
```
lgoettgens: All places that construct GAPErrors use the constructor with three or four arguments, so the `hasproperty` is not really needed. Furthermore, it will error anyway, either with an UndefinedPropertyError or due to the fallback value being useless.

This also applies to all additional `hasproperty` tests further down

test/errors.jl
```diff
@@ -0,0 +1,238 @@
...
+        @test err !== nothing
+        @test err isa GAPError
+        @test hasproperty(err, :message)
+        @test occursin("boom", message_of(err))
+        @test hasproperty(err, :gap_frames)
+
+        frames = gap_frames_of(err)
+        @test frames isa Vector
+        @test !isempty(frames)
+        @test all(frame -> frame isa GAP.GAPStackFrame, frames)
```
lgoettgens:
```suggestion
        @test frames isa Vector{GAP.GAPStackFrame}
        @test !isempty(frames)
```

test/errors.jl
```diff
@@ -0,0 +1,238 @@
...
+
+        labels = [frame.function_label for frame in frames]
+        @test "gapjl_traceback_inner" in labels
+        @test "gapjl_traceback_outer" in labels
+        @test any(frame -> frame.file !== nothing, frames)
+        @test any(frame -> frame.line !== nothing, frames)
```
lgoettgens: What is the reason to test for the existence of *at least one* frame with file-line-info?

test/errors.jl
```diff
@@ -0,0 +1,238 @@
...
+        @test err isa GAPError
+
+        shown = try
+            throw(err)
+        catch thrown
+            sprint(showerror, thrown, catch_backtrace())
+        end
+        @test occursin("Julia stacktrace", shown)
+        @test occursin("julia_gap_traceback_inner", shown)
+        @test occursin("julia_gap_traceback_outer", shown)
```
lgoettgens: These tests don't test the order of the frames in the printed stack. If you have any good ideas on how to do that, that would be great. (I don't have a suggestion right now)

lgoettgens (commented):
some more comments, but now mostly minor ones. Overall, this looks now quite good IMO

.github/workflows/gap.yml
fingolfin: Your AI agent has a point, and mine raised that as well back when I asked it to write this code; but the problem is that figuring out the right number by itself is a rather complicated piece of code. I figured that if this starts failing, then the fix is rather obvious and easy; while debugging the complex code needed to "guess" the version seemed more problematic (given that'd we'd likely be doing it infrequently, every other year or so). I felt that was a bad tradeoff

A compromise would be to install e.g. the `gcc@14` package and use `gcc-14`; that would then be "safe" for a longer time, but of course get us a bit older compiler (which shouldn't be a problem in practice, though?) Sadly homebrew doesn't offer `gcc@15` right now (once they have gcc 16, their `gcc` package will provide that, and there will be a new `gcc@15` package... ugh

lgoettgens: just as a remark: the yggdrasil build uses gcc v7

fingolfin: I switched to GCC 14

test/errors.jl
fingolfin: I've made this strict clearer, I hope.

test/errors.jl
fingolfin: done

fingolfin:
I've tweaked it a bit so that calls in `ccalls.jl` in `GAP.jl` are visible again, as I found that helpful in debugging - but perhaps just because I also work on GAP.jl and so it is not a "confusing detail' for me to see, e.g., `(::GapObj)(a1::Int64)` pop up as a function name.

So the example from the initial PR description now looks like this (with `_call_gap_func` and `(::GapObj)(a1::Int64)` now "visible"
```
julia> gap_func(n) = GAP.Globals.SymmetricGroup(n);

julia> gap_func(-3)
ERROR: Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
GAP stacktrace:
 [1] HANDLE_METHOD_NOT_FOUND
     @ ~/.julia/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/lib/methsel2.g:250
 [2] SymmetricGroup
     @ ~/.julia/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/grp/basic.gd:734
Stacktrace:
 [1] _call_gap_func
   @ ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:361 [inlined]
 [2] (::GapObj)(a1::Int64)
   @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:338
 [3] gap_func(n::Int64)
   @ Main ./REPL[32]:1
 [4] top-level scope
   @ REPL[33]:1
```

lgoettgens:
Why did you remove the "Julia" from the second "Stacktrace:" header? I would either leave it there as the other one is "GAP stacktrace", or merge both stacktraces (then you don't need any headers anymore, but numbering should change as well)

fingolfin:
I removed it because the "Stacktrace:" is always printed by Julia, I can't hide it. So were printing
```
Julia Stacktrace:
Stacktrace:
```
Which looked weird.

For the same reason I can't "merge" the two stacktraces

fingolfin:
Any further remarks on this, @lgoettgens ? It'd be kinda cool if we could get this in together with GAP 4.16.0 (OTOH... maybe it would also be stupid -- two major changes at once... hm)

src/errors.jl
```diff
@@ -9,38 +9,295 @@
...
+has_gap_error_state() = hasproperty(Globals, :_JULIAINTERFACE_CLEAR_ERROR_STACK)
+
+function Base.showerror(io::IO, err::GAPError)
+    show_gap_error(io, err)
+end
+
+function Base.showerror(io::IO, err::GAPError, bt; backtrace=true)
+    show_gap_error(io, err)
+    if backtrace && !isempty(err.julia_stacktrace)
+        Base.show_backtrace(io, Any[err.julia_stacktrace...])
```
lgoettgens, line 66:
```suggestion
        Base.show_backtrace(io, Any[err.julia_stacktrace...]; prefix = "Julia ")
```
this should then only print
```
Julia Stacktrace:
[...]
```

lgoettgens:
> Any further remarks on this, @lgoettgens ? It'd be kinda cool if we could get this in together with GAP 4.16.0 (OTOH... maybe it would also be stupid -- two major changes at once... hm)

I more minor comment (above). I think I am happy with this in the current state. Putting it into the same release as GAP 4.16.0 seems fine to me. But I would prefer to first merge the GAP update into the GAP.jl master branch and then test this PR once again (both on CI and manually) to verify that everything still works as expected.

src/errors.jl
fingolfin: That option was only added in Julia 1.13

lgoettgens:
I just fixed the conflicts and will play around with this together with GAP 4.16.0 tomorrow

fingolfin:
The Julia nightly failures are genuine. It looks for `[4] julia_gap_traceback_outer` in the error (which admittedly is very tight, making this a bit laxer is sensible).

But the error message we get differs quite a bit from what we see in older Julia versions (so in that sense I am glad the test is so strict):
```
Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
GAP stacktrace:
 [1] HANDLE_METHOD_NOT_FOUND
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/lib/methsel2.g:258
 [2] SymmetricGroup
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/grp/basic.gd:740
Stacktrace:
 [1] _call_gap_func(func::GapObj, a1::Int64)
   @ GAP ~/work/GAP.jl/GAP.jl/src/ccalls.jl:361 [inlined]
 [2] (::GapObj)(a1::Int64)
   @ GAP ~/work/GAP.jl/GAP.jl/src/ccalls.jl:338
 [3] (::var\"#julia_gap_traceback_inner#julia_gap_traceback_inner##0\")()
   @ Main ~/work/GAP.jl/GAP.jl/test/errors.jl:161
 [4] (::var\"#julia_gap_traceback_outer#julia_gap_traceback_outer##0\"{var\"#julia_gap_traceback_inner#julia_gap_traceback_inner##0\"})()
   @ Main ~/work/GAP.jl/GAP.jl/test/errors.jl:162 [inlined]
 [5] (::var\"#43#44\"{var\"#julia_gap_traceback_outer#julia_gap_traceback_outer##0\"{var\"#julia_gap_traceback_inner#julia_gap_traceback_inner##0\"}})()
   @ Main ~/work/GAP.jl/GAP.jl/test/errors.jl:165 [inlined]
 [6] capture_exception(f::var\"#43#44\"{var\"#julia_gap_traceback_outer#julia_gap_traceback_outer##0\"{var\"#julia_gap_traceback_inner#julia_gap_traceback_inner##0\"}})
   @ Main ~/work/GAP.jl/GAP.jl/test/errors.jl:16
 [7] top-level scope
   @ ~/work/GAP.jl/GAP.jl/test/errors.jl:24")
```

For reference, this is what I get in Julia 1.12:
```
Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
GAP stacktrace:
 [1] HANDLE_METHOD_NOT_FOUND
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/lib/methsel2.g:258
 [2] SymmetricGroup
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/grp/basic.gd:740
Stacktrace:
 [1] _call_gap_func
   @ ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:361 [inlined]
 [2] (::GapObj)(a1::Int64)
   @ GAP ~/Projekte/OSCAR/GAP.jl/src/ccalls.jl:338
 [3] (::var"#julia_gap_traceback_inner#julia_gap_traceback_inner##0")()
   @ Main ~/Projekte/OSCAR/GAP.jl/test/errors.jl:161
 [4] julia_gap_traceback_outer
   @ ~/Projekte/OSCAR/GAP.jl/test/errors.jl:162 [inlined]
 [5] #25
   @ ~/Projekte/OSCAR/GAP.jl/test/errors.jl:165 [inlined]
 [6] capture_exception(f::var"#25#26"{var"#julia_gap_traceback_outer#julia_gap_traceback_outer##0"{var"#julia_gap_traceback_inner#julia_gap_traceback_inner##0"}})
   @ Main ~/Projekte/OSCAR/GAP.jl/test/errors.jl:16
 [7] top-level scope
   @ ~/Projekte/OSCAR/GAP.jl/test/errors.jl:24
```

lgoettgens:
The main application of this still works as before with GAP 4.16.0.

While playing around with this, I also tried a more mutually recursive call. I know that this PR explicitly *not* attempts to also make stacktraces work for these cases. Is it possible to instead of the below half-ish stacktrace we print a sentence that stacktraces for mutually recursive GAP-Julia calls are not supported yet?
```julia-repl
julia> func1(n) = GAP.Globals.SymmetricGroup(n);

julia> func2(n) = GAP.evalstr("Julia.func1($n)");

julia> func2(-3)
ERROR: Error thrown by GAP: Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
Error, Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
GAP stacktrace:
 [1] HANDLE_METHOD_NOT_FOUND
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/lib/methsel2.g:258
 [2] SymmetricGroup
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/grp/basic.gd:740
not in any function at *defin*:0
Stacktrace:
 [1] evalstr(cmd::String)
   @ GAP ~/code/julia/GAP.jl/src/ccalls.jl:158
```

fingolfin:
> Is it possible to instead of the below half-ish stacktrace we print a sentence that stacktraces for mutually recursive GAP-Julia calls are not supported yet?

Maybe -- but I am not sure whether we can even detect that situation reliably? Or do you want that message to be *always* printed?

I will (let Codex) have a quick look to see what is possible, but if it is hard, I suggest we first merge this, and iterate later; it has been sitting here long enough...

Anyway, for now I fixed the merge conflict (and revised the CHANGES.md entry)

lgoettgens:
> > Is it possible to instead of the below half-ish stacktrace we print a sentence that stacktraces for mutually recursive GAP-Julia calls are not supported yet?
> 
> Maybe -- but I am not sure whether we can even detect that situation reliably? Or do you want that message to be _always_ printed?
>
> I will (let Codex) have a quick look to see what is possible, but if it is hard, I suggest we first merge this, and iterate later; it has been sitting here long enough...

No, just if it happened. Thanks for looking into it. I agree that this is already much better than the current version, and thus we should get it in asap

fingolfin:
I've asked Codex to look into your comment, and just pushed the resulting commit. Now I get this:
```julia-repl
julia> func1(n) = GAP.Globals.SymmetricGroup(n);

julia> func2(n) = GAP.evalstr("Julia.func1($n)");

julia> func2(-3)
ERROR: Error thrown by GAP: Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
Julia/GAP stacktraces for recursive GAP -> Julia -> GAP error paths are not supported yet.
Error, Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
GAP stacktrace:
 [1] HANDLE_METHOD_NOT_FOUND
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/lib/methsel2.g:258
 [2] SymmetricGroup
     @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/grp/basic.gd:740
not in any function at *defin*:0
```

lgoettgens:
> I've asked Codex to look into your comment, and just pushed the resulting commit. Now I get this:
> 
> ```julia-console
> julia> func1(n) = GAP.Globals.SymmetricGroup(n);
> 
> julia> func2(n) = GAP.evalstr("Julia.func1($n)");
> 
> julia> func2(-3)
> ERROR: Error thrown by GAP: Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
> Julia/GAP stacktraces for recursive GAP -> Julia -> GAP error paths are not supported yet.
> Error, Error thrown by GAP: no method found! For debugging hints type ?Recovery from NoMethodFound
> GAP stacktrace:
>  [1] HANDLE_METHOD_NOT_FOUND
>      @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/lib/methsel2.g:258
>  [2] SymmetricGroup
>      @ ~/.julia/artifacts/2095dbb0f458ae555c3bae2c73802851612d20be/share/gap/grp/basic.gd:740
> not in any function at *defin*:0
> ```

Thanks! This looks like what I thought of. Unfortunately, CI does not like it.

lgoettgens (approved):

lgoettgens:
It looks like the nightly CI errors have something to do with stack traces. Could you please try to fix them?

Issue #1371, opened 2026-04-07
Loading GAP.jl with GAP installed and precompiled in read-only system depot fails


fingolfin:
We have installed Oscar.jl and GAP.jl via the advanced install instructions at <https://www.oscar-system.org/install/generic/> for use in our computer labs.

This entails installing these packages into a special Julia depot that is owned by a system user. Everything is precompiled in there, too.

Later, when ordinary users are trying to run Oscar, the idea is that this special depot is put into Julia's depot search below the user's primary depot. Meaning that anything new is added into the user's home depot, but any packages, artifacts etc. in the shared special depot can be accessed.

Unfortunately this fails in a strange way when loading GAP.jl (and hence when loading Oscar). I have not yet fully analyzed this, but I suspect this is a mixture of issues. I'll first provide the error I am seeing and then will share some theories about this:

```
julia> using GAP
Error, GAP is not bound in Julia at /users/test/Oscar/depot/packages/GAP/DpWCq/pkg/JuliaInterface/gap/JuliaInterface.gi:48 called from
Scratch := julia.GAP.eval( ValueGlobal( "JuliaEvalString" )( ":(import Scratch; Scratch)" ) )
 ; at /users/test/Oscar/depot/artifacts/eac0e07d9389a9a61d6e38f612b9394d26e87cd5/atlasrep-2.1.9/gap/userpref.g:118 called from
record.default(
 ) at /users/test/Oscar/depot/artifacts/21532479477ddf5e931137aa39cb1e00a8edb15c/share/gap/lib/userpref.g:285 called from
<function "DeclareUserPreference">( <arguments> )
 called from read-eval loop at /users/test/Oscar/depot/artifacts/eac0e07d9389a9a61d6e38f612b9394d26e87cd5/atlasrep-2.1.9/gap/userpref.g:166
ERROR: InitError: GAP variable _JULIAINTERFACE_ERROR_BUFFER not bound
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] getproperty(::GAP.GlobalsType, name::Symbol)
    @ GAP /users/test/Oscar/depot/packages/GAP/DpWCq/src/globals.jl:53
  [3] copy_gap_error_to_julia()
    @ GAP /users/test/Oscar/depot/packages/GAP/DpWCq/src/GAP.jl:76
  [4] initialize(argv::Vector{String})
[...]

PR #1372, opened 2026-04-10
Update `AtlasRep` to 2.1.10

lgoettgens:
towards https://github.com/oscar-system/GAP.jl/issues/1371


lgoettgens:
Bump @fingolfin

ThomasBreuer:
The failure of the tests for the walrus package is not related to the new version of AtlasRep.

The failing test takes a random graph `RandomDigraph( 100, 0.05 )` and calls `DigraphDijkstraST` on it. With the given settings, I get the same failure in at least one out of 150 cases, due to the fact that `DigraphDijkstraST` may return `infinity` instead of the expected list.

fingolfin (approved):

PR #1377, opened 2026-05-04
Conversions for `fpMatrix`/`FpMatrix`

mjrodgers:
Basic (preliminary) conversion methods for matrices over finite fields; here we cover `fpMatrix` and `FpMatrix`.

Originally https://github.com/oscar-system/Oscar.jl/pull/5945

### DIFF
--- a/ext/NemoExt/nemo_to_gap.jl
+++ b/ext/NemoExt/nemo_to_gap.jl
@@ -78,3 +78,9 @@ function GAP.GapObj_internal(obj::Union{ZZMatrix,QQMatrix}, ::GapCacheDict, ::Va
     end
     return ret_val
 end
+
+## matrix of elements of a finite field (of prime order)
+GAP.@install function GapObj(obj::Union{fpMatrix, FpMatrix})
+    e = GAP.Globals.Z(GapObj(characteristic(base_ring(obj))))
+    return GapObj(lift(obj)) * e
+end
--- a/test/NemoExt/nemo_to_gap.jl
+++ b/test/NemoExt/nemo_to_gap.jl
@@ -135,3 +135,15 @@ import AbstractAlgebra
     @test GAP.Obj(x) == val
   end
 end
+
+@testset "fpMatrix" begin
+    F = Native.GF(5)
+    m = matrix(F, 4, 4, 1:16)
+    @test GAP.Obj(m) == GAP.evalstr("[ [ Z(5), Z(5)^2, Z(5)^0, Z(5)^3 ], [ 0*Z(5), Z(5), Z(5)^2, Z(5)^0 ], [ Z(5)^3, 0*Z(5), Z(5), Z(5)^2 ], [ Z(5)^0, Z(5)^3, 0*Z(5), Z(5) ] ]")
+end
+
+@testset "FpMatrix" begin
+    F = Native.GF(ZZ(5))
+    m = matrix(F, 4, 4, 1:16)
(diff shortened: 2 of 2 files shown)


ext/NemoExt/nemo_to_gap.jl
```diff
@@ -78,3 +78,9 @@ function GAP.GapObj_internal(obj::Union{ZZMatrix,QQMatrix}, ::GapCacheDict, ::Va
     end
     return ret_val
 end
+
+## matrix of elements of a finite field (of prime order)
+GAP.@install function GapObj(obj::Union{fpMatrix, FpMatrix})
+    e = GAP.Globals.Z(GapObj(characteristic(base_ring(obj))))
+    return GapObj(lift(obj)) * e
+end
```
fingolfin, line 86: Looking at this I was thinking about suggesting using some of the existing field element conversion code... then realized most of that is still in OSCAR. Now I wonder if there is more code in OSCAR's `src/GAP/` that we should migrate to GAP.jl...

mjrodgers: I don't think we have much there, at least in `oscar_to_gap.jl`; I only see conversions for `AbsSimpleNumFieldElem` and `QQAbFieldElem` as far as field element conversions. Am I overlooking them somewhere?

fingolfin: But `AbsSimpleNumFieldElem` is from Nemo, isn't it? (To be fair, for *some* of the code involving it, Hecke is needed, e.g. when `Hecke.is_cyclotomic_type` is accessed. However, the stuff in `src/GAP/oscar_to_gap.jl` seem ready for a transfer here?)

`QQAbFieldElem` is of course from OSCAR and so we have to leave everything involving it there.

There is also code dealing with `zzModRing`, `ZZModRing`, `FqFieldElem`, etc.; some of it might be transferable?

Pinging also @ThomasBreuer

fingolfin: (That said, we can certainly first merge this PR and do more work on additional PR...)

mjrodgers: Ah okay, I have no problem moving some of these things to Nemo as part of this pull request. What is the process for this? (I know we need separate pull requests in Nemo and Oscar, just wondering if anything special needs to be done to synchronize them)

lgoettgens: I would prefer if this were split into a new PR, as moving code upstream (from Oscar to GAP->NemoExt) will probably need a breaking GAP release.



> (I know we need separate pull requests in Nemo and Oscar, just wondering if anything special needs to be done to synchronize them)

If you name the branches exactly the same in both repositories, there will be another CI job that runs tests under the assumption that both branches in both repositories would already be merged (i.e. to test both of them together).

fingolfin: Let's discuss it tomorrow afternoon?

(If @ThomasBreuer or @lgoettgens have comments, or would like to be involved, let us know)

fingolfin (approved):
Seems fine to me, thanks @mjrodgers 

@ThomasBreuer @lgoettgens shall we merge it?

ext/NemoExt/nemo_to_gap.jl
```diff
@@ -78,3 +78,9 @@ function GAP.GapObj_internal(obj::Union{ZZMatrix,QQMatrix}, ::GapCacheDict, ::Va
     end
     return ret_val
 end
+
+## matrix of elements of a finite field (of prime order)
+GAP.@install function GapObj(obj::Union{fpMatrix, FpMatrix})
+    e = GAP.Globals.Z(GapObj(characteristic(base_ring(obj))))
+    return GapObj(lift(obj)) * e
```
fingolfin, line 85: This multiplication is the "right thing" to do. And in the next GAP version, this will be faster, thanks to https://github.com/gap-system/gap/pull/6398

lgoettgens (approved):

ThomasBreuer (approved):

PR #1379, opened 2026-05-05
Update GAP to 4.16.0

lgoettgens:
Closes https://github.com/oscar-system/GAP.jl/pull/1370.

This loads locally built binaries of the gap 4.16.0-beta1 (using recipes from https://github.com/JuliaPackaging/Yggdrasil/pull/13635, https://github.com/JuliaPackaging/Yggdrasil/pull/13636, and https://github.com/JuliaPackaging/Yggdrasil/compare/master...lgoettgens:Yggdrasil:lg/GAP_pkg (to be split into multiple PRs once the gap release is out)).

CI should inform us about any regressions in GAP.jl and OSCAR (at least once the CI runners are up again).

cc @fingolfin


lgoettgens:
The "CI with GAP" jobs are expected to fail, as the `[sources]` section in the `Project.toml` is just used for the project itself, but not for dependencies. And in the "CI with GAP" jobs, for technical reasons, we use a layerd environment approach.

Edit: this now includes https://github.com/oscar-system/GAP.jl/pull/1380 which changes that

lgoettgens:
CI will fail until https://github.com/JuliaPackaging/Yggdrasil/pull/13665 is merged (but this has some complications)

lgoettgens:
This looks quite good. In particular, the OSCAR downstream tests succeed :tada: 

Some comments on the failures:
- Everything with julia 1.10 does not understand the `[sources]` feature that I use to load unreleased jlls.
- The two CI-with-GAP jobs with macOS have some `LOAD_DYN` issue (see https://github.com/oscar-system/GAP.jl/actions/runs/25510277396/job/74866893252?pr=1379#step:9:1206 for the complete stacktrace). I already tried porting over the patches from Yggdrasil as they are fiddling around with linker options, but that did not help. @fingolfin any idea what this could be about?

fingolfin:
no I a puzzled by this right now -- it happened *before* the GAP kernel major version was  incremented. Weird.

.github/workflows/GAP_patches/stable-4.16/sysinfo.patch
fingolfin, line 1: We should move to stable-4.16 I guess?

Project.toml
```diff
@@ -64,37 +64,37 @@ Artifacts = "1.10"
...
+GAP_pkg_deepthought_jll = "~101.900.1"
+GAP_pkg_digraphs_jll = "~101.140000.0"
+GAP_pkg_edim_jll = "~101.30800.1"
+GAP_pkg_ferret_jll = "~101.1600.0"
+GAP_pkg_float_jll = "~101.1000.0"
+GAP_pkg_fplsa_jll = "~101.20800.0"
+GAP_pkg_gauss_jll = "~202500.120100.0"
+GAP_pkg_guava_jll = "~301.210000.0"
+GAP_pkg_io_jll = "~401.90300.1"
+GAP_pkg_json_jll = "~201.20400.0"
```
fingolfin:
```suggestion
GAP_pkg_json_jll = "~201.40000.0"
```

Project.toml
```diff
@@ -64,37 +64,37 @@ Artifacts = "1.10"
...
+GAP_pkg_digraphs_jll = "~101.140000.0"
+GAP_pkg_edim_jll = "~101.30800.1"
+GAP_pkg_ferret_jll = "~101.1600.0"
+GAP_pkg_float_jll = "~101.1000.0"
+GAP_pkg_fplsa_jll = "~101.20800.0"
+GAP_pkg_gauss_jll = "~202500.120100.0"
+GAP_pkg_guava_jll = "~301.210000.0"
+GAP_pkg_io_jll = "~401.90300.1"
+GAP_pkg_json_jll = "~201.40000.0"
 GAP_pkg_juliainterface_jll = "=0.160000.1"
```
lgoettgens:
```suggestion
GAP_pkg_juliainterface_jll = "=0.170000.0"
```

lgoettgens:
@fingolfin do you have an idea about the upstream CI macos failures?

fingolfin:
@lgoettgens not yet. I'll look into it.

lgoettgens:
The failures of the `utils` package also seem deterministic, and thus worth looking at.

ThomasBreuer:
> The failures of the `utils` package also seem deterministic, and thus worth looking at.

I can reproduce the error message locally in some situation, I will investigate this further.

ThomasBreuer:
Concerning the problem with `GAP.Packages.test("utils")`:

The current version 0.95 of `utils` wants to run `makedoc.g` as a part of the package tests, in order to generate/update some test files that contain the examples from the package manual.
This means to build the documentation of the package in the current directory (using the `autodoc` package), but this fails because the directory is not writable in the situation that the package is a Julia artifact.

When an up-to-date package manual has been built then the test files in question are available, hence running `makedoc.g` is not necessary for the package tests.
However, this is not the case in the CI tests of the `utils` package itself.

A clean solution would be to test the manual examples via the GAPDoc functions `ExtractExamples`/`RunExamples`
directly, without creating/updating test files for them.
Shall we change `utils` accordingly, and then request the new  `utils` version in `Artifacts.toml`?

lgoettgens:
> A clean solution would be to test the manual examples via the GAPDoc functions `ExtractExamples`/`RunExamples`
> directly, without creating/updating test files for them.
> Shall we change `utils` accordingly, and then request the new `utils` version in `Artifacts.toml`?

This sounds reasonable to me. Thanks for looking into it!

cdwensley:
The PR of @ThomasBreuer has been merged and version 0.96 of utils released.

fingolfin:
I've imported the new `utils` version, and applied a patch which hopefully will fix the macOS issue. It was created together with Codex; I supplied the guess for the fix, it did the implementation and local testing. Let's see if it works in CI. 

I have no idea why it worked before, though. Huh.

Codex chose to put the new helpers into a new file `etc/override_utils.jl`, but we could also inline that.

fingolfin (approved):
From my POV this is good to go, thanks!

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Changes in GAP.jl
 
+## Version 0.17.0-DEV (released YYYY-MM-DD)
+
+- **Breaking:** Update to GAP 4.16.0.
```
lgoettgens:
```suggestion
- **Breaking:** Update to GAP 4.16.0
- Update the GAP package distribution to 4.16.0
- Update the `utils` GAP package to 0.96
- ```

lgoettgens:
> I've imported the new `utils` version, and applied a patch which hopefully will fix the macOS issue. It was created together with Codex; I supplied the guess for the fix, it did the implementation and local testing. Let's see if it works in CI.
> 
> I have no idea why it worked before, though. Huh.
> 
> Codex chose to put the new helpers into a new file `etc/override_utils.jl`, but we could also inline that.

If I understand the code correctly, it just calls `install_name_tool -id $(prefix)/lib/libgap.[1-9]+.dylib @rpath/libgap.[1-9]+.dylib` (as pseudocode). The code is IMO way to complicated for this simple thing. I will try to inline and simplify it.

lgoettgens:
> > I've imported the new `utils` version, and applied a patch which hopefully will fix the macOS issue. It was created together with Codex; I supplied the guess for the fix, it did the implementation and local testing. Let's see if it works in CI.
> > I have no idea why it worked before, though. Huh.
> > Codex chose to put the new helpers into a new file `etc/override_utils.jl`, but we could also inline that.
> 
> If I understand the code correctly, it just calls `install_name_tool -id $(prefix)/lib/libgap.[1-9]+.dylib @rpath/libgap.[1-9]+.dylib` (as pseudocode). The code is IMO way to complicated for this simple thing. I will try to inline and simplify it.

Done. I also incorporated an old patch from https://github.com/oscar-system/GAP.jl/commit/48cbbabf126273d54ccaf27703703414edfb7737#diff-57c0ee68eebbb63675048be14480929ea2cfecb112937c867630e800aebeee04, that renames the dylib in case that the kernel version between GAP_jll.jl and upstream gap differ. This should thus long-term fix the downstream CI in the gap repo (as discussed in on slack).
@fingolfin if you are still happy with the state of this branch, please merge.

fingolfin (approved):

PR #1380, opened 2026-05-05
Make override setup work with `[sources]` in `Project.toml`

lgoettgens:
by re-using the package environment instead of adding one on top. This changes the way the override setup needs to be called, see the changes in `README.maintainer.md` for details.

### DIFF
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -81,8 +81,8 @@ jobs:
           make -j`nproc`
       - name: "Override bundled GAP"
         run: |
-          julia --proj=override etc/setup_override_dir.jl /tmp/GAPROOT /tmp/gap_jll_override --enable-Werror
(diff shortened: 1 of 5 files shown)


fingolfin (approved):
I forgot why I wanted the separate project. If this works better for you right now, I am all for it. And if it causes me or others problems in the future, in the worst case we can revert *and then document the problem*. But chances are there is none (anymore( *shrug*

lgoettgens:
> I forgot why I wanted the separate project. If this works better for you right now, I am all for it. And if it causes me or others problems in the future, in the worst case we can revert _and then document the problem_. But chances are there is none (anymore( _shrug_

Before we changed to the JLLPrefixes approach, we had to install a bunch of things (like GMP_jll) manually. And this would have bloated the main environment. (at least that's my best guess)

Issue #1378, opened 2026-05-04
For CI tests change GAP user preference `DownloadMaxTime` to some reasonable value (e.g. 30 seconds?)


fingolfin:
And then the same should be done for Oscar.jl

See https://github.com/gap-system/gap/issues/5985 for background.

PR #1381, opened 2026-05-05
set timeout for downloads in `GAP.Packages.test`

ThomasBreuer:
We want to set a timeout for GAP's `Download` function (from the utils package) in CI tests.

The CI tests for GAP.jl itself and for JuliaInterface do not call GAP code that wants to download files, thus the relevant CI tests are the ones that call `GAP.Packages.test` and the ones that run the Oscar tests.

The function `GAP.Packages.test` now sets and resets a timeout.
The Oscar tests are run by calling `Pkg.test("Oscar")`, this will be handled by setting the timeouts in Oscar.
(I did not find a CI test that runs parts of the GAP test suite. Such a test would also need a timeout.)

addresses #1378

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -510,6 +510,11 @@ function test(name::String)
     return
   end
 
+  # Set a timeout of 30 seconds for web downloads.
+  # This affects functions that use the AtlasRep package if it is available.
+  orig_timeout = Globals.UserPreference(GapObj("utils"), GapObj("DownloadMaxTime"))
+  Globals.SetUserPreference(GapObj("utils"), GapObj("DownloadMaxTime"), 30)
+
   disable_error_handler[] = true
   result = false
   try
@@ -534,6 +539,9 @@ function test(name::String)
     called_QuitGap || throw(ArgumentError("GAP's TestPackage failed"))
     result = true
   end
+
+  Globals.SetUserPreference(GapObj("utils"), GapObj("DownloadMaxTime"), orig_timeout)
+
   return !error_occurred && result
 end
 


src/packages.jl
```diff
@@ -510,6 +510,11 @@ function test(name::String)
     return
   end
 
+  # Set a timeout of 30 seconds for web downloads.
+  # This affects functions that use the AtlasRep package if it is available.
+  orig_timeout = Globals.UserPreference(GapObj("utils"), GapObj("DownloadMaxTime"))
+  Globals.SetUserPreference(GapObj("utils"), GapObj("DownloadMaxTime"), 30)
```
fingolfin, line 516: So this will then also affect every user calling `GAP.Packages.test` Do we really want that?

Alternative approach (for discussion): when setting up CI, we could do something like this:
```
mkdir -p ~/.gap
echo >> ~/.gap/gap.ini  # ensure gap.ini ends with newline
echo 'SetUserPreference( "utils", "DownloadMaxTime", 300 );' >> ~/.gap/gap.ini
```


This could perhaps go into the start of `etc/ci_test.sh`, or into one or several of the `.github/workflow/*.yml` files

lgoettgens: I agree that this is not the right place, and see the following options:
1. what @fingolfin suggests above
2. unconditionally set a timeout in the GAP.jl `__init__` for all sessions

ThomasBreuer: > So this will then also affect every user calling `GAP.Packages.test` Do we really want that?

I would say yes.
The condition for a test run should not depend on how the tests are run, directly or indirectly, via some github workflow or by starting them interactively.
(Analogously, GAP's function `Test` changes some settings when it gets called.)

lgoettgens: But why *just* for the tests. Doesn't every user of atlasrep benefit of a timeout with fallback servers?

ThomasBreuer: It was stated like that in #1378.

My understanding was that in a test situation, a timeout will have the effect that the test does not hang but fails; then it can be restarted and will hopefully pass.

Outside a test situation, it is not clear what one expects. In the end, it is a *user* preference; are we allowed to simply change it by GAP.jl or Oscar.jl?

If we want to take the "pragmatic" approach to set the timeout whenever GAP.jl starts GAP (then Oscar needs not worry about this), perhaps we should at least check whether the timeout is already set to a positive integer, and in this case keep this value. (This will not happen in CI tests.)

PR #1382, opened 2026-05-08
unconditionally set a global `Download` timeout

### LINKED ISSUE #1378: For CI tests change GAP user preference `DownloadMaxTime` to some reasonable value (e.g. 30 seconds?)

ThomasBreuer:
an alternative to and thus closes #1381, as proposed by @lgoettgens:

Always set a `Download` timeout of 30 seconds, except if the value in the user's `gap.ini` defines already a finite timeout value.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changes in GAP.jl
 
+## Version 0.16.7-DEV (released 2026-XX-XX)
+
+- Set a timeout of 30 seconds for GAP's `Download` function,
+  except if the user's `gap.ini` file defines already a finite timeout value.
+
 ## Version 0.16.6 (released 2026-04-30)
 
 - Declare compatibility with AbstractAlgebra 0.49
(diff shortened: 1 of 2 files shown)


fingolfin (approved):
Just discussed this with @ThomasBreuer at length. Everything is difficult 😂 

But this seems OK to me. In an ideal world, we'd also give a message to the user informing them if there was a download timeout, and if so, how to adjust the timeout. But that can wait for the future.

lgoettgens (approved):
LGTM! Should this be mentioned in the changelog then?

ThomasBreuer:
> Should this be mentioned in the changelog then?

I will update the pull request accordingly.

fingolfin:
The CI failures in `CI with GAP` agains GAP master fails on macOS due to the GAP kernel version there having changed. As such, it is not a problem of this PR.

PR #1386, opened 2026-05-21
Make the juliainterface manual available

### LINKED ISSUE #624: building the documentation for JuliaInterface
@fingolfin wrote in the discussion of pull request #617:

I just realized that we never really build the documentation for `JuliaInterface`, do we? We probably should (HTML+TXT only, no PDF, so that user's don't need LaTeX). But how?? We have to be careful to do it only after GAP actually runs, so we can't do it during precompilation as part of `regenerate_gaproot()` (I think).

Next best I can think of is to *delete* any generated documentation files in `regenerate_gaproot()`, and then regenerate them in `__init__` if we notice they are missing (this way, we always regenerate them if needed).

Another solution would be to be slightly nasty and commit the relevant files (*.html, *.txt, *.js, *.css, ...) into the repository... 

Or if JuliaInterface was built into a JLL (see issue #619), then that JLL could contain the generated documentation, too.

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/624.

~~This requires https://github.com/JuliaPackaging/Yggdrasil/pull/11356 to be available.~~

We look for the files in a particular relative path w.r.t. the `JuliaInterface.so` file. If the so is loaded from the GAP_pkg_juliainterface_jll artifacts, this will succeed as of https://github.com/JuliaPackaging/Yggdrasil/pull/11356. For a locally compiled `JuliaInterface.so`, this will not work at the moment (to be improved in a future PR).


ThomasBreuer (approved):
Looks good, thanks.

One more possible problem is that there may be a JuliaInterface manual in the package path because one has read `makedoc.g` into the GAP session. This happens for example for developers when they work on the documentation. In this case, now two help books for JuliaInterface get added to the GAP session, and at least one of them is outdated. This is also something for a later pull request.

ThomasBreuer:
As far as I understand the test failures, no version of `GAP_pkg_juliainterface_jll` fits the requirements.
This pull request sets the version to `=0.160000.1`, but this is apparently not yet available.

lgoettgens:
> As far as I understand the test failures, no version of `GAP_pkg_juliainterface_jll` fits the requirements. This pull request sets the version to `=0.160000.1`, but this is apparently not yet available.

It was merged in Yggdrasil this morning, and was registered in https://github.com/JuliaRegistries/General/pull/156087. So it should be available now (modulo caching)

src/GAP.jl
```diff
@@ -261,6 +261,11 @@ function __init__()
       Globals.SetUserPreference(GapObj("utils"), GapObj("DownloadMaxTime"), 30)
     end
 
+    # Load the JuliaInterface manual (from a relative path of the juliainterface.so file)
+    juliaInterface_pkginfo = only(GAP.Globals.PackageInfo(GapObj("juliainterface")))
+    juliaInterface_pkgdoc = only(juliaInterface_pkginfo.PackageDoc)
+    GAP.Globals.HELP_ADD_BOOK(juliaInterface_pkgdoc.BookName, juliaInterface_pkgdoc.LongTitle, GAP.Globals.Directory(GapObj(abspath(dirname(JuliaInterface_path), "..", "..", "share", "doc", "JuliaInterface", "doc"))))
```
fingolfin: Very long line, and annoying to debug. How about splitting it, e.g.
```suggestion
    ji_docpath = abspath(dirname(JuliaInterface_path), "..", "..", "share", "doc", "JuliaInterface", "doc")
    ji_docpath_gap = GAP.Globals.Directory(GapObj(ji_docpath))
    GAP.Globals.HELP_ADD_BOOK(juliaInterface_pkgdoc.BookName, juliaInterface_pkgdoc.LongTitle, ji_docpath_gap)
```

fingolfin (approved):

PR #1387, opened 2026-06-03
Improve CI coverage tracking for JuliaInterface

fingolfin:
Generate gap.sh with coverage when requested.

Let CI reuse one instrumented JuliaInterface build explicitly.

---

I've asked Codex to combine my earlier PRs trying to improve coverage, and make it work. Let's see...

WARNING: I have not yet reviewed this closely myself, though it did already undergo some revisions. Still, AI slop my await you, so don't waste your time reviewing this at this point unless you really want to :-). My goal is to first see if get a coverage increase; only then is it worth to invest more time.

Closes #1339. Closes #1343. Closes #1344.

### DIFF
--- a/etc/ci_test.sh
+++ b/etc/ci_test.sh
@@ -5,17 +5,26 @@ set -x
 
 AnyFailures=No
 
+JuliaInterfaceBuildDir="${PWD}/pkg/JuliaInterface/tmp"
+
+export CFLAGS="${CFLAGS:+${CFLAGS} }--coverage"
+export LDFLAGS="${LDFLAGS:+${LDFLAGS} }--coverage"
+export FORCE_JULIAINTERFACE_COMPILATION="${JuliaInterfaceBuildDir}"
+
 mkdir -p coverage
 #
 cd pkg/JuliaInterface
 pwd
 # Force recompilation of JuliaInterface with coverage instrumentation.
 # Use a fixed build directory so that gcov can find .gcno/.gcda files.
-CFLAGS="--coverage" LDFLAGS="--coverage" FORCE_JULIAINTERFACE_COMPILATION=tmp ${GAP} --nointeract
+${GAP} --nointeract
+JuliaInterfaceSo=$(find "${JuliaInterfaceBuildDir}/bin" -name JuliaInterface.so -type f | head -n 1)
+test -n "${JuliaInterfaceSo}"
+export GAP_JL_JULIAINTERFACE_SO="${JuliaInterfaceSo}"
+unset FORCE_JULIAINTERFACE_COMPILATION
+
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaInterface.coverage -r tst/testall.g || AnyFailures=Yes
-gcov -o tmp/gen/src/ src/*.c*
-rm -rf tmp # Delete the coverage instrumentation in JuliaInterface again
 cd ../..
 
 #
@@ -23,6 +32,11 @@ cd pkg/JuliaExperimental
 pwd
 ${GAP} makedoc.g
 ${GAP} --cover ../../coverage/JuliaExperimental.coverage -r tst/testall.g || AnyFailures=Yes
+cd ../..
+
+cd pkg/JuliaInterface
+gcov -o "${JuliaInterfaceBuildDir}/gen/src/" src/*.c*
+rm -rf "${JuliaInterfaceBuildDir}" # Delete the coverage instrumentation in JuliaInterface again
 cd ../..
 
 if [ ${AnyFailures} = Yes ]
--- a/etc/ci_test.sh
+++ b/etc/ci_test.sh
@@ -20,8 +20,18 @@ then
     JuliaInterfaceSo="${GAP_JL_JULIAINTERFACE_SO}"
     test -f "${JuliaInterfaceSo}"
     JuliaInterfaceBuildDir=$(cd "$(dirname "${JuliaInterfaceSo}")/../.." && pwd)
-    unset FORCE_JULIAINTERFACE_COMPILATION
-else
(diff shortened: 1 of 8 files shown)


fingolfin:
It works!

fingolfin:
Code looks fine to me.

The CI failures is probably due to the mismatch between GAp kernel versions in GAP_jll and the gap repo. So should be re-run after we merge the 4.16.0 changes.

src/setup.jl
```diff
@@ -270,8 +278,13 @@ Given a directory path, create three files in that directory:
 - two TOML files, `Manifest.toml` and `Project.toml`, which are required by
   the script to function (they record the precise versions of GAP.jl and other
   Julia packages involved)
+
+If `code_coverage` is true, the generated script starts Julia with
```
lgoettgens:
```suggestion
If `code_coverage` is `true`, the generated script starts Julia with
```

etc/ci_test.sh
```diff
@@ -5,17 +5,26 @@ set -x
...
+
 mkdir -p coverage
 #
 cd pkg/JuliaInterface
 pwd
 # Force recompilation of JuliaInterface with coverage instrumentation.
 # Use a fixed build directory so that gcov can find .gcno/.gcda files.
-CFLAGS="--coverage" LDFLAGS="--coverage" FORCE_JULIAINTERFACE_COMPILATION=tmp ${GAP} --nointeract
+${GAP} --nointeract
+JuliaInterfaceSo=$(find "${JuliaInterfaceBuildDir}/bin" -name JuliaInterface.so -type f | head -n 1)
```
lgoettgens: Isn't this equivalent to
```suggestion
JuliaInterfaceSo="${JuliaInterfaceBuildDir}/bin/JuliaInterface.so"
```
?

etc/ci_test.sh
```diff
@@ -5,17 +5,26 @@ set -x
...
+test -n "${JuliaInterfaceSo}"
```
lgoettgens: Does this line like this make the CI job fail? Or does it just proceed silently?

src/setup.jl
```diff
@@ -300,6 +313,10 @@ function create_gap_sh(dstdir::String, dstname::String="gap.sh"; use_active_proj
         @info "Generating gap.sh ..."
     end
 
+    julia_cmd = filter(arg -> !startswith(arg, "--code-coverage"), String.(Base.julia_cmd().exec))
+    if code_coverage
+        push!(julia_cmd, "--code-coverage")
```
lgoettgens: I am not completely happy with this, as the `--code-coverage` can have more than two values.
Suggestion for semantics: the kwarg `code_coverage` can be either a `String` or a `Nothing`, where `nothing` stands over taking the setting of the current julia session.
There would then be no filter at all. In case the field is a `String`, we just `push!(julia_cmd, "--code-coverage=$(code_coverage)")`. This works, as in the presence of multiple conflicting cli flags, julia will always take the *last* one.

etc/ci_test.sh
fingolfin: I might be too tired and getting something wrong, but no, I don't think so; it should be `bin/GAPARCH/JuliaInterface.so`

etc/ci_test.sh
fingolfin: This script is running with `-e`, so any top level command exiting with an error aborts it. So if `$JuliaInterfaceSo` is empty (because no match was found), then this aborts the script.

src/setup.jl
fingolfin: fine by me

etc/ci_test.sh
lgoettgens: yeah, you're right...

src/setup.jl
fingolfin: done via Codex (have not yet reviewed it myself, will do that next)

fingolfin: OK, it seems good enough that it shouldn't be a total waste of your time to review it again, @lgoettgens ;-)

lgoettgens (approved):

lgoettgens:
Could you add a changelog entry, as this adds a kwarg to `create_gap_sh`?

lgoettgens:
CI seems to be unhappy :(

Is there a particular reason why you moved the coverage generation from the main CI jobs to the tree hash one?

fingolfin:
> CI seems to be unhappy :(

Looking into it now.

> Is there a particular reason why you moved the coverage generation from the main CI jobs to the tree hash one?

To get full coverage for JuliaInterface.so, we must compile it with `--coverage`; that means we can't use the version from the JLL. And the `treehash` test already does exactly that, so it is a more natural fit than trying to shoehorn this into the `CI.yml` tests. For those, there is the added complication that of course also must have at least some tests that *do* use the JLL -- one of the intermediate versions of this PR completely removed all tests of that situation, which was bad, so I asked Codex to redo it.

lgoettgens:
> > Is there a particular reason why you moved the coverage generation from the main CI jobs to the tree hash one?
> 
> To get full coverage for JuliaInterface.so, we must compile it with `--coverage`; that means we can't use the version from the JLL. And the `treehash` test already does exactly that, so it is a more natural fit than trying to shoehorn this into the `CI.yml` tests. For those, there is the added complication that of course also must have at least some tests that _do_ use the JLL -- one of the intermediate versions of this PR completely removed all tests of that situation, which was bad, so I asked Codex to redo it.

Sounds reasonable. Thanks for the explanation

fingolfin:
Tests were passing just now, but I've just pushed one more cleanup tweak (overall I instructed the AI to cleanup the `etc/ci_test.sh` script quite a bit).

Assuming the tests pass again, this is ready now from my POV.

fingolfin:
There is still an issue in so far as `call_gap_func` in `pkg/JuliaInterface/src/calls.c` is not getting any coverage reported. But I can analyze that in a follow-up PR; this one already does quite enough, IMHO.

PR #1388, opened 2026-06-05
Try to make REPL tests more reliable

lgoettgens:
even if this does not work with julia nightly yet, it is still an improvement


fingolfin (approved):
seems plausible, thanks

PR #1389, opened 2026-06-09
Remove unreleased 0.16.7 from changelog

lgoettgens:
Alternatively, we could create a release-0.16 branch right before merging the GAP 4.16.0 bump and do the 0.16.7 release from there. But I don't think that we get that much from that, so I wouldn't do that.


fingolfin:
This morning I was inclined to agree, but in light of PR #1390 by @thofma I guess we should quickly release 0.16.7 after all..

PR #1385, opened 2026-05-19
Migrate `AbsSimpleNumFieldElem` conversion code from Oscar

mjrodgers:
We have some things from `oscar_to_gap` that really belong here in `nemo_to_gap`, attempting to move them and see what could go wrong.

PR #1392, opened 2026-06-11
Run downstream tests against Oscar.jl

lgoettgens:
This is just a matching branch for https://github.com/oscar-system/Oscar.jl/pull/6054 to be able to run the downstream tests against Oscar.jl.
cc @fingolfin

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -74,7 +74,7 @@ jobs:
         env:
           JULIA_DEBUG: "GAP"
         run: |
-          julia --project=. --depwarn=error -e 'import GAP; GAP.create_gap_sh("/tmp")'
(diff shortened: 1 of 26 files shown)


lgoettgens:
Yay, CI is green :tada:
I will keep this branch open to be able to re-run tests once things like https://github.com/oscar-system/GAP.jl/pull/1385 and https://github.com/oscar-system/GAP.jl/pull/1361 are merged

lgoettgens:
Ok, we (or rather I) were stupid. The OscarCI jobs in the branch have all been suceeding in < 1 minute, which hints at a Pkg resolution failure. And indeed, I created a Hecke branch and locally edited its Project.toml, but I never pushed the commit doing that.

I thus just restarted CI here another time to now actually let it run... Let's hope for the best (cc @fingolfin )

fingolfin:
Oof! Well I also missed it, glad you noticed it

PR #1393, opened 2026-06-16
Increase code coverage a bit

fingolfin:


lgoettgens:
This PR shows a problem with doing JuliaInterface coverage in the treehash job: Since you touch a file of JuliaInterface, the treehash job fails (as it should, to remind us of a jll rebuild). But since that happens as the first step of that job, everything regarding coverage is not even executed.

In my opinion it would make sense to split the current treehash job into two:
1. just compare the treehashes and fail if they differ
2. executing the tests with a forced juliainterface rebuild (with coverage instrumentation)

If we would do that, the former job would still fail for this PR, but the latter would show us that the coverage indeed increased for the c code.

fingolfin:
@lgoettgens agreed.

Of course separate from that, we need to decide if we want to merge this before 0.17.0 (and then make a new JuliaInterface release), or wait...

fingolfin:
Actually, I could also split the C changes from the GAP/Julia changes (which can be merged right away)

lgoettgens (approved):

PR #1396, opened 2026-06-17
Split the treehash CI job into two

lgoettgens:
As suggested in https://github.com/oscar-system/GAP.jl/pull/1393#issuecomment-4727337149

The new "Test with JuliaInterface rebuild" job should also be marked as "required" in the github UI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -125,7 +125,7 @@ jobs:
       - name: "Set up Julia"
         uses: julia-actions/setup-julia@v3
         with:
-          version: ${{ matrix.julia-version }}
+          version: 1
         id: setup-julia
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -91,6 +91,7 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
+
```
fingolfin, line 94: Any reason for this newline? but not important
```suggestion
```

fingolfin (approved):

.github/workflows/CI.yml
lgoettgens: just for readability. I found it easier to distinguish the different jobs with more than one newline in-between.

PR #1398, opened 2026-06-17
NewJuliaObj: turn check into assert

fingolfin:
Calling this with a GAPObj would just be invalid, and our current
code cannot do this, ever. We also don't guard against e.g. null
pointers or other invalid pointer.

As a side effect, this also increases code coverage.

### DIFF
--- a/pkg/JuliaInterface/src/JuliaInterface.c
+++ b/pkg/JuliaInterface/src/JuliaInterface.c
@@ -117,8 +117,7 @@ static Obj JuliaObjectTypeFunc(Obj o)
 
 Obj NewJuliaObj(jl_value_t * v)
 {
-    if (is_gapobj(v))
-        return (Obj)v;
+    GAP_ASSERT(!is_gapobj(v));
     JL_GC_PUSH1(&v);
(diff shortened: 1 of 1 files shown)


lgoettgens (approved):

lgoettgens:
Actually, this PR *decreases* code coverage, but that is due to a technicality in the juliainterface treehash checking code (which now goes into the rebuild branch instead of the branch using the bundled jll), see https://app.codecov.io/gh/oscar-system/GAP.jl/pull/1398/indirect-changes#a4159829f69eb5e4656baeffc0eda4eb-L258.

Is this something we can somehow avoid?

fingolfin:
I don't think we can do much about it: if our C sources have deviated from the JLL, then we just don't enter the code that is only run if the checksums match. But I think that's fine?

What I can do is: add a test that forces a checksum mismatch (say, by appending a whitespace to a C file), to ensure that code path is always exercised. Doesn't seem terribly important, though

lgoettgens:
Yeah, just keep it as is. I just wanted to write down my observation somewhere. We just have to remember that this can (and will) occur from time to time.

PR #1399, opened 2026-06-17
Relax nightly backtrace assertion

fingolfin:
Stop assuming exact Julia frame numbers or pretty-printed local
function names in the traceback test. Julia nightly currently
renders those local helpers as closure frames, so the test should
only check that the inner and outer helper names both appear and
in the right order.


lgoettgens (approved):

lgoettgens:
The nightly failures now are just expect timeouts. Activating auto-merge...

PR #1400, opened 2026-06-18
Version 0.17.0

lgoettgens:
I think all the PRs that we wanted to wait for are now merged. I do not really have an opinion if we merge https://github.com/oscar-system/GAP.jl/pull/1398 and do the jll rebuild cycle before or after this release.

Downstream tests in https://github.com/oscar-system/GAP.jl/pull/1392 were happy yesterday. I just updated the branch there to latest master to really make sure that we don't break OSCAR. Once they are green again (and I *only* care there about downstream CI, in particular not about any julia nightly stuff), this PR is good to go from my POV.

Closes https://github.com/oscar-system/GAP.jl/pull/1392.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Changes in GAP.jl
 
-## Version 0.17.0-DEV (released YYYY-MM-DD)
+## Version 0.17.0 (released 2026-06-18)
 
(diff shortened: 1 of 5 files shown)


fingolfin (approved):
Looks good to me, thanks!

PR #1401, opened 2026-06-18
Isolate JuliaInterface C coverage

fingolfin:
Only track JuliaInterface C coverage on Linux. This avoids merging compiler-dependent executable-line mappings, which lead to surprising and misleading "uncovered code" reports.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -110,9 +110,28 @@ jobs:
 
 
   test-with-juliainterface-rebuild:
-    name: "Test with JuliaInterface rebuild"
-    runs-on: ubuntu-latest
+    name: JuliaInterface rebuild - Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # Collect C coverage with Julia LTS on Linux. Restricting C coverage
+          # to GCC on Linux avoids merging incompatible executable-line maps.
+          - julia-version: lts
+            os: ubuntu-latest
+
+          # Also test the rebuild with the latest stable Julia on Linux.
+          - julia-version: '1'
+            os: ubuntu-latest
+
+          # Cover platform-specific Julia setup code on macOS as well. Do not
+          # collect C coverage here: Clang on macOS and GCC on Linux disagree
+          # about which C source lines are executable, and merging their reports
+          # makes Codecov mark covered continuation lines as uncovered.
+          - julia-version: '1'
+            os: macOS-latest
     steps:
       - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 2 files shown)


etc/ci_test.sh
```diff
@@ -83,25 +92,27 @@ ${GAP} --cover ../../coverage/JuliaInterface.coverage -r tst/testall.g || AnyFai
...
+# remain available for later workflow steps.
+if [ "${JULIAINTERFACE_COVERAGE:-No}" = Yes ]
 then
-    rm -rf "${JuliaInterfaceBuildDir}" # Delete the coverage instrumentation in JuliaInterface again
+    cd pkg/JuliaInterface
+    gcov -o "${JuliaInterfaceBuildDir}/gen/src/" src/*.c*
+
+    if [ "${RemoveJuliaInterfaceBuildDir}" = Yes ]
+    then
+        rm -rf "${JuliaInterfaceBuildDir}"
```
lgoettgens:
```suggestion
        rm -rf "${JuliaInterfaceBuildDir}" # Delete the coverage instrumentation in JuliaInterface again
```
this comment would be worth to keep IMO

lgoettgens (dismissed):

lgoettgens:
This produces more coverage changes in some files than I expected. In particular, the coverage loss in `src/julia-config.jl` confuses me a bit.

fingolfin:
Yeah I also find that confusing. I've applied your good change suggestion, and will investigate further tomorrow

lgoettgens:
This has conflicts now

fingolfin:
Conflicts resolved, commits squashed, comments updated

etc/ci_test.sh
```diff
@@ -86,25 +96,29 @@ ${GAP} --cover ../../coverage/JuliaInterface.coverage -r tst/testall.g || AnyFai
...
 
-# Convert the C coverage counters from the instrumented JuliaInterface build
-# into .gcov files for Codecov. This must happen before deleting the build dir.
-cd pkg/JuliaInterface
-gcov -o "${JuliaInterfaceBuildDir}/gen/src/" src/*.c*
-
-# Avoid leaving this script's temporary coverage build behind. When the build
-# was supplied by the caller, keep it available for later workflow steps.
-if [ "${RemoveJuliaInterfaceBuildDir}" = Yes ]
+# Only the Linux LTS rebuild job emits JuliaInterface C coverage. Convert its
```
lgoettgens: This fact is stated multiple times in this file. But looking closely at it, the file should only talk about the `JULIAINTERFACE_COVERAGE` env var, or "rebuild jobs with coverage". Keeping it in the current prose is prone for future inconsistencies if the CI matrix is touched but this file isn't

.github/workflows/CI.yml
```diff
@@ -125,40 +144,48 @@ jobs:
...
-          CFLAGS: "--coverage"
-          LDFLAGS: "--coverage"
+          CFLAGS: ${{ runner.os == 'Linux' && matrix.julia-version == 'lts' && '--coverage' || '' }}
+          LDFLAGS: ${{ runner.os == 'Linux' && matrix.julia-version == 'lts' && '--coverage' || '' }}
           JULIAINTERFACE_BUILD_DIR: "${{ github.workspace }}/pkg/JuliaInterface/tmp"
         run: |
           mkdir -p pkg/JuliaInterface/src/force-treehash-mismatch
           touch pkg/JuliaInterface/src/force-treehash-mismatch/nonempty
-          julia --project=. --color=yes -e 'import GAP; write("juliainterface.path", GAP.JuliaInterface_path)'
+          julia --project=. --color=yes --code-coverage=user -e 'import GAP; write("juliainterface.path", GAP.JuliaInterface_path)'
```
lgoettgens, line 167: Why include the flag here unconditionally and not just in the one matrix case?

.github/workflows/CI.yml
```diff
@@ -125,40 +144,48 @@ jobs:
...
       - name: "Cache artifacts"
         uses: julia-actions/cache@v3
       - name: "Build package"
         uses: julia-actions/julia-buildpkg@v1
-      - name: "Build JuliaInterface for C coverage"
+      - name: "Build JuliaInterface with coverage"
         env:
-          CFLAGS: "--coverage"
-          LDFLAGS: "--coverage"
+          CFLAGS: ${{ runner.os == 'Linux' && matrix.julia-version == 'lts' && '--coverage' || '' }}
```
lgoettgens: Wouldn't it be easier to understand if the "with-coverage?" was a matrix dimension, cause then you wouldn't need to repeat the condition over and over again

.github/workflows/CI.yml
fingolfin: Added a comment above to explain

.github/workflows/CI.yml
fingolfin: indeed

etc/ci_test.sh
fingolfin: good point

lgoettgens (approved):

lgoettgens:
The CI job has changed its name, so we need to adapt the "required CI jobs" for this to become mergeable. (I'll do that later today if I don't forget and nobody beats me to it)

PR #1404, opened 2026-06-26
Honor GAP package artifact overrides

fingolfin:
Use gap_pkg_artifact_dir when building GAP package roots so
UUID/name artifact overrides are respected during startup.

Allow the helper to return an override directory directly when it
already contains PackageInfo.g, and cover that behavior with a
regression test.

Together, this allows usage of an Overrides.toml looking like this:

    [c863536a-3901-11e9-33e7-d5cd0df7b904]
    GAP_pkg_alnuth = "/some/alternate/path"

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 ## Version 0.17.2 (released 2026-06-21)
 
 - Update the `alnuth` GAP package to 4.0.2
+- Honor `Overrides.toml` UUID/name overrides for `GAP_pkg_*` artifacts
+  when locating GAP package directories
 
 ## Version 0.17.1 (released 2026-06-21)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,13 @@
 # Changes in GAP.jl
 
+## Version 0.17.3-DEV (released 2026-MM-DD)
+
+- Honor `Overrides.toml` UUID/name overrides for `GAP_pkg_*` artifacts
+  when locating GAP package directories
+
 ## Version 0.17.2 (released 2026-06-21)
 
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -3,6 +3,8 @@
 ## Version 0.17.2 (released 2026-06-21)
 
 - Update the `alnuth` GAP package to 4.0.2
+- Honor `Overrides.toml` UUID/name overrides for `GAP_pkg_*` artifacts
+  when locating GAP package directories
```
lgoettgens, line 6:
```suggestion

## Version 0.17.3-DEV (released 2026-MM-DD)

- Honor `Overrides.toml` UUID/name overrides for `GAP_pkg_*` artifacts
  when locating GAP package directories

## Version 0.17.2 (released 2026-06-21)

- Update the `alnuth` GAP package to 4.0.2
```

test/setup.jl
```diff
@@ -43,3 +43,51 @@ end
...
+    override_dir = joinpath(tmpdir, "override", "alnuth")
+    mkpath(override_dir)
+    write(joinpath(override_dir, "PackageInfo.g"), "")
+    write(
+      artifacts_toml,
+      """
+      [GAP_pkg_alnuth]
+      git-tree-sha1 = "809593e819b916279aa515bc8644a9c1e5ab2a96"
+      """,
+    )
```
lgoettgens: What is this Artifacts.toml actually used for in this test?

fingolfin: It wasn't, and is removed now. Thanks for catching this.

lgoettgens (approved):

Issue #1407, opened 2026-07-29
download errors with some packages (e.g. Browse, GAPDoc)


benlorenz:
From time to time the Oscar tests are failing because some gap packages are failing to download:

```
  Failed to install some artifacts:
  
  Unable to automatically download/install artifact 'GAP_pkg_gapdoc' from sources listed in '/home/runner/.julia/packages/GAP/leFHp/Artifacts.toml'.
  Sources attempted:
  - https://pkg.julialang.org/artifact/b48bc43492416cba88eb3066634c32e1d08005b3
      Error: RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/b48bc43492416cba88eb3066634c32e1d08005b3
  - https://www.math.rwth-aachen.de/~Frank.Luebeck/GAPDoc/GAPDoc-1.6.10.tar.bz2
      Error: RequestError: Connection timed out after 30010 milliseconds while requesting https://www.math.rwth-aachen.de/~Frank.Luebeck/GAPDoc/GAPDoc-1.6.10.tar.bz2
  - https://files.gap-system.org/pkg/GAPDoc-1.6.10.tar.bz2
      Error: RequestError: HTTP/2 404 while requesting https://files.gap-system.org/pkg/GAPDoc-1.6.10.tar.bz2
  
  Unable to automatically download/install artifact 'GAP_pkg_browse' from sources listed in '/home/runner/.julia/packages/GAP/leFHp/Artifacts.toml'.
  Sources attempted:
  - https://pkg.julialang.org/artifact/eadf6b3d3655ba9bd89084bbe9300095c2c0b1c8
      Error: RequestError: HTTP/2 404 while requesting https://pkg.julialang.org/artifact/eadf6b3d3655ba9bd89084bbe9300095c2c0b1c8
  - https://www.math.rwth-aachen.de/~Browse/Browse-1.8.22.tar.bz2
      Error: RequestError: Connection timed out after 30006 milliseconds while requesting https://www.math.rwth-aachen.de/~Browse/Browse-1.8.22.tar.bz2
  - https://files.gap-system.org/pkg/Browse-1.8.22.tar.bz2
      Error: RequestError: HTTP/2 404 while requesting https://files.gap-system.org/pkg/Browse-1.8.22.tar.bz2
  Error: Process completed with exit code 1.
```

Multiple issues:

[...]

PR #1408, opened 2026-07-30
Improve `etc/update_artifacts.jl`

lgoettgens:
Some steps towards https://github.com/oscar-system/GAP.jl/issues/1407.

Notable changes:

- Properly detect all syntactical variants of `ArchiveFormats`, that GAP allows
- If a package has both .tar.gz and .tar.bz2, prefer .tar.gz independent of the order in `ArchiveFormats`. See https://github.com/JuliaPackaging/PkgServer.jl/issues/238 for the reason
- If a package has multiple formats or locations, verify that all provide the same tree-hash.
- Only put locations in the Artifacts.toml that actually exist.

This PR does not change anything yet for running GAP.jl. One would need to do a follow-up PR to this one where one executes the script.

### DIFF
--- a/etc/update_artifacts.jl
+++ b/etc/update_artifacts.jl
@@ -17,7 +17,7 @@
 #     julia --project=etc etc/update_artifacts.jl local/path/meta.json
 #
 
-using Downloads: download
+using Downloads: download, RequestError
 import Pkg
 using Pkg.Artifacts
 import Pkg.PlatformEngines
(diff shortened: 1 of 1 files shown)


ThomasBreuer (approved):
Thanks.
It works.

When *all* archives of a package are missing, then the script runs into an error because `git_tree_sha1s` is empty. Perhaps a warning should be printed instead, and the package in question should be regarded as unavailable.

lgoettgens:
With https://github.com/gap-system/PackageDistro/pull/1464, this is no longer needed

lgoettgens:
Obsolete due to https://github.com/gap-system/PackageDistro/pull/1464

PR #1409, opened 2026-08-03
Update gap packages to versions with `.tar.gz` archive

lgoettgens:
Resolves https://github.com/oscar-system/GAP.jl/issues/1407. Closes https://github.com/oscar-system/GAP.jl/pull/1410.

- [x] browse: `.tar.gz` available since https://github.com/gap-system/PackageDistro/pull/1461, `.tar.gz` preferred archive since https://github.com/gap-system/PackageDistro/pull/1464
- [x] crisp: `.tar.gz` available since https://github.com/gap-system/PackageDistro/pull/1469. Also include https://github.com/gap-system/PackageDistro/pull/1483 to work around https://github.com/bh11/crisp/issues/13
- [x] edim: `.tar.gz` preferred archive since https://github.com/gap-system/PackageDistro/pull/1464
- [x] gapdoc: `.tar.gz` preferred archive since https://github.com/gap-system/PackageDistro/pull/1464
- [x] irredsol: `.tar.gz` available since https://github.com/gap-system/PackageDistro/pull/1470

cc @benlorenz @ThomasBreuer @fingolfin

PR #1410, opened 2026-08-04
Fix URL of crisp

aaruni96:
After the release of 1.4.9, "latest" no longer works for trying to download 1.4.8

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -243,7 +243,7 @@ git-tree-sha1 = "85ac41f54aedd09fb370d977dd6ee3309684056d"
 
     [[GAP_pkg_crisp.download]]
     sha256 = "3360e17c1611a2ab1654d982aade5566a1c048c795675e321cbf15d06248d569"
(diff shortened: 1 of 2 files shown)


ThomasBreuer:
This change will fix the problem for the moment.
However, the file `Artifacts.toml` will be updated by `etc/update_artifacts.jl`, thus the real problem is that the source data of this script must be corrected.

(See also bh11/crisp/pull/14)

lgoettgens:
This is included in https://github.com/oscar-system/GAP.jl/pull/1409 which we can hopefully merge soonish

fingolfin (approved):
yes also fine by me. yes the next import will break it, but hopefully by then there will be a fixed crips release. but this buys us some time

lgoettgens (approved):
Let's get this merged and released immediately to fix CI in various places. https://github.com/oscar-system/GAP.jl/pull/1409 may then wait a bit longer for an irredsol release.

PR #1415, opened 2026-08-13
handle `resume` option in the Julia based `Download` method

ThomasBreuer:
Once gap-packages/utils#105 becomes available, the `GAP.Globals.Download` method that is based on the Julia package Downloads.jl must decline requests where `resume:= true` is set.

(Downloads.jl  is based on libcurl, which supports resuming, but the function `Downloads.download` apparently does not support this.)

### DIFF
--- a/src/packages.jl
+++ b/src/packages.jl
@@ -52,6 +52,11 @@ function init_packagemanager()
          :name => "via Julia's Downloads.download",
          :isAvailable => Globals.ReturnTrue,
          :download => function(url, opt)
+            if hasproperty(opt, :resume)
+              return GapObj(Dict{Symbol, Any}(:success => false,
+                                              :error => GapObj("no support for resuming")),
+                            recursive=true)
+            end
             try
               timeout = hasproperty(opt, :maxTime) ? Real(opt.maxTime) : Inf
               if hasproperty(opt, :target)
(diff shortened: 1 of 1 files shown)


lgoettgens:
Is this safe to merge before that `utils` version dripped down to GAP.jl? (I would of course wait until https://github.com/gap-packages/utils/pull/105 isa actually merged)

ThomasBreuer:
I think this is safe.

The change has an effect (failure of the download) only if the user adds on purpose `resume:= true` to the optional second argument of `GAP.Globals.Download`, which is not supported before gap-packages/utils/pull/105 becomes public.

src/packages.jl
```diff
@@ -52,6 +52,11 @@ function init_packagemanager()
          :name => "via Julia's Downloads.download",
          :isAvailable => Globals.ReturnTrue,
          :download => function(url, opt)
+            if hasproperty(opt, :resume)
```
fingolfin, line 55: But `resume` could also be set to `false`. So perhaps better do this:
```suggestion
            if hasproperty(opt, :resume) && opt.resume == true
```

fingolfin (approved):

fingolfin (approved):

PR #1416, opened 2026-08-24
Add script to update GAP*_jll compat entries

fingolfin:
Bumping the ~30 GAP*_jll bounds in Project.toml after a GAP release was a manual step with no tooling. The new `etc/update_jll_versions.jl` reads the release's `package-infos.json`, then pins each JLL to the newest registered build of exactly that upstream package version.

Taking the registry's newest version instead would be wrong: a `GAP_pkg_X_jll` must be built from the same version of package X that `Artifacts.toml` ships, because `setup_overrides` grafts the JLL's binaries onto the artifact's source tree.

Prepared with AI assistance (Claude Opus 5).

### DIFF
--- a/README.maintainer.md
+++ b/README.maintainer.md
@@ -176,18 +176,20 @@ all of the JLL packages that depend on GAP.
    `GAP_pkg_*_jll` packages.
 
 7. In `GAP.jl`, create a PR with the following changes:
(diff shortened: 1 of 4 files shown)


lgoettgens (approved):
Looks good (at seems to work given that you used it for https://github.com/oscar-system/GAP.jl/pull/1417)

Issue #1224, opened 2025-08-20
GC related crash


fingolfin:
Just encountered this while playing with @ThomasBreuer PR #1222

```
julia> v = [1, "x"]
2-element Vector{Any}:
 1
  "x"

julia> t = (2, v)
(2, Any[1, "x"])

julia> v[2] = t
(2, Any[1, (#= circular reference @-2 =#)])

julia> v
2-element Vector{Any}:
 1
  (2, Any[#= circular reference @-2 =#])

julia> using GAP
...

julia> GapObj(v)
GAP: [ 1, <Julia: (2, Any[1, (#= circular reference @-2 =#)])> ]

julia> GapObj(v, recursive=true)
GAP: [ 1, [ 2, ~ ] ]

julia> vg = GapObj(v, recursive=true)
GAP: [ 1, [ 2, ~ ] ]

julia> typeof(v)
Vector{Any} (alias for Array{Any, 1})

julia> Vector{Any}(
[99804] signal 11 (2): Segmentation fault: 11
in expression starting at REPL[9]:1
gc_setmark_pool_ at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-XC9YQX9HH2.0/build/default-honeycrisp-XC9YQX9HH2-0/julialang/julia-release-1-dot-12/src/./gc-stock.h:0 [inlined]
gc_setmark_pool at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-XC9YQX9HH2.0/build/default-honeycrisp-XC9YQX9HH2-0/julialang/julia-release-1-dot-12/src/gc-stock.c:307 [inlined]
gc_setmark at /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-XC9YQX9HH2.0/build/default-honeycrisp-XC9YQX9HH2-0/julialang/julia-release-1-dot-12/src/gc-stock.c:314 [inlined]
[...]

Issue #1032, opened 2024-09-04
Still getting GC crashes even with GAP_jll disabling task stack scanning


fingolfin:
... at least in CI, I cannot yet reproduce it locally.

ThomasBreuer:
I get the following on my notebook with Julia 1.8.5 and with GAP.jl as in the current master branch.
```
julia> versioninfo()
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver1)
  Threads: 1 on 8 virtual cores

julia> using GAP
[ Info: Compiling JuliaInterface ...
 ┌───────┐   GAP 4.13.1 of 2024-06-11
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-julia1.8-64-kv9
 Configuration:  gmp 6.2.1, Julia GC, Julia 1.8.5, readline
 Loading the library and packages ...
reading ~/.gap/gaprc
 Packages:   AClib 1.3.2, Alnuth 3.2.1dev, AtlasRep 2.1.9, AutPGrp 1.11, 
             Browse 1.8.21, CRISP 1.4.6, Cryst 4.1.27, CrystCat 1.1.10, 
             CTblLib 1.3.9, FactInt 1.6.3, FGA 1.5.0, Forms 1.2.11, 
             GAPDoc 1.6.7.dev, genss 1.6.8, IO 4.8.2, IRREDSOL 1.4.4, 
             JuliaInterface 0.11.1, LAGUNA 3.9.6, orb 4.9.0, Polenta 1.3.10, 
             Polycyclic 2.17dev, PrimGrp 3.4.4, RadiRoot 2.9, recog 1.4.3, 
             ResClasses 4.7.3, SmallGrp 1.5.3, Sophus 1.27, SpinSym 1.5.2, 
             StandardFF 1.0, TomLib 1.2.11, TransGrp 3.6.5, utils 0.85
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'

julia> using Test

julia> include("test/Aqua.jl") # testfile in GAP.jl
[...]

PR #1422, opened 2026-08-28
Add regression tests for task stack scanning by the GC

### LINKED ISSUE #996: Add tests involving Julia tasks
In particular, add such tests which stress the garbage collector, as it is not quite clear how well our current task stack scanning code works (or if it indeed even is correct).

This is ever more important when multiple GC threads are active. But even with a single GC thread this may reveal hidden issues.

fingolfin:
A GAP object held only in a C stack frame of a Julia task must survive collections that run while the task is suspended (parked in a Julia call, or busy on another thread). This went wrong when the task stack rescan optimization in GAP's julia_gc.c skipped old tasks, the cause of long-elusive crashes; see #1032, #1224 and gap-system/gap#6525.

The tests hold a canary object solely via a wrapper bag in a GAP kernel argument temporary and detect premature collection through per-canary finalizers, covering automatic and explicit incremental collections and, in a -t2 subprocess, a never-switching task on a second thread. Only automatic collections exercise the historically buggy path, so the driver churns allocations rather than calling GC.gc.

Resolves #996

AI disclosure: tests developed with the assistance of Claude Code.

### DIFF
--- a/.github/workflows/gap.yml
+++ b/.github/workflows/gap.yml
@@ -86,9 +86,12 @@ jobs:
           echo "${shimdir}" >> "${GITHUB_PATH}"
       - name: "Build GAP"
         run: |
+          # absolute path: the patches are applied from inside the GAP
+          # checkout below, where a relative path would not be found and the
+          # patches would silently be skipped
+          patchdir=${GITHUB_WORKSPACE}/.github/workflows/GAP_patches/${{ matrix.gap-version }}
           mv GAPROOT /tmp/GAPROOT
           cd /tmp/GAPROOT
-          patchdir=.github/workflows/GAP_patches/${{ matrix.gap-version }}
           if [ -d $patchdir ]; then
             for f in $patchdir/*.patch; do
               echo "Applying patch ${f}"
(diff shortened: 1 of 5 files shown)


.github/workflows/gap.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         run: |
           mv GAPROOT /tmp/GAPROOT
           cd /tmp/GAPROOT
-          patchdir=.github/workflows/GAP_patches/${{ matrix.gap-version }}
+          patchdir=${GITHUB_WORKSPACE}/.github/workflows/GAP_patches/${{ matrix.gap-version }}
```
lgoettgens: This seems a bit unrelated to the rest of this PR (but looks fine anyway)

.github/workflows/gap.yml
```diff
@@ -88,7 +88,7 @@ jobs:
...
```
fingolfin:
```suggestion
          patchdir=.github/workflows/GAP_patches/${{ matrix.gap-version }}
```

fingolfin:
I think this is complete

lgoettgens (approved):
Seems reasonable, even though I don't understand the details

.github/workflows/gap.yml
fingolfin: It's not unrelated, and there is now a comment above it that explains
```
          # absolute path: the patches are applied from inside the GAP
          # checkout below, where a relative path would not be found and the
          # patches would silently be skipped
```

PR #1423, opened 2026-08-29
Speed up `collect` and list comprehensions for GAP lists

fingolfin:
For iterators of unknown size and eltype, the generic `collect` assembles the result via `Base.grow_to!`, which performs several dynamically dispatched calls per element. Iterating a `GapObj` always has unknown size and eltype, so nothing is lost by gathering the elements ourselves: for GAP lists we can use the known length and fill a vector of that size directly, everything else (GAP iterators and collections) goes through the iteration protocol.

The element type of the result is determined as by the generic code: it starts out as the type of the first element and is widened via `promote_typejoin` whenever an element does not fit, holes in lists are skipped, and if there is no element at all the type comes from inference. As there, the `isa` checks let the compiler drop the widening branches when the element type is concrete, so that a comprehension such as `[f(x)::T for x in gaplist]` is still inferred as `Vector{T}` and yields one also for an empty list; Oscar.jl relies on both.

Benchmarks using `Chairmarks.jl` on Julia 1.12:

    julia> L = GAP.evalstr("List([3..6],SymmetricGroup)");

    julia> v = GapObj(collect(1:10000));

    julia> w = GapObj([GapObj([i]) for i in 1:10000]);

Before:

    julia> @b collect($L)
    1.869 μs (24 allocs: 928 bytes)

    julia> @b [x for x in $L]
    1.847 μs (25 allocs: 944 bytes)

    julia> @b collect($v)
    773.416 μs (49003 allocs: 1.423 MiB)

    julia> @b collect($w)
    630.125 μs (30025 allocs: 1.133 MiB)

After:

    julia> @b collect($L)
    257.505 ns (2 allocs: 96 bytes)

    julia> @b [x for x in $L]
    258.178 ns (2 allocs: 96 bytes)

    julia> @b collect($v)
    71.000 μs (9495 allocs: 228.375 KiB)

    julia> @b collect($w)
    32.458 μs (6 allocs: 80.109 KiB)

On Julia 1.10, where the dispatch overhead of the generic code is smaller, the same benchmarks improve from 475.000 ns, 471.218 ns, 743.625 μs and 637.750 μs to 270.047 ns, 266.667 ns, 69.042 μs and 38.416 μs, respectively.

| benchmark | 1.12 before | 1.12 after | 1.10 before | 1.10 after |
|---|---|---|---|---|
| `collect(L)` — 4 groups | 1.869 µs<br>24 allocs | 257 ns<br>2 allocs | 475 ns<br>15 allocs | 270 ns<br>1 alloc |
| `[x for x in L]` | 1.847 µs<br>25 allocs | 258 ns<br>2 allocs | 471 ns<br>16 allocs | 267 ns<br>1 alloc |
| `collect(v)` — 10k ints | 773 µs<br>49003 allocs | 71 µs<br>9495 allocs | 744 µs | 69 µs |
| `collect(w)` — 10k GapObjs | 630 µs<br>30025 allocs | 32 µs<br>**6 allocs** | 638 µs | 38 µs |

Part of issue #736.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/test/adapter.jl
+++ b/test/adapter.jl
@@ -54,6 +54,7 @@
     @test collect(GAP.evalstr("[]")) isa Vector{Any}
     l = GAP.evalstr("[1,,3]")
     @test collect(l) isa Vector{Int}
+    @test collect(l) == [1, 3]
     @test [2 * x for x in l] == [2, 6]
     @test [x for x in GAP.evalstr("[1, 2, 3]")] isa Vector{Int}
 
--- a/test/adapter.jl
+++ b/test/adapter.jl
@@ -45,6 +45,28 @@
 
     # some things cannot be iterated over
     @test_throws ArgumentError collect(GAP.Globals.GAPInfo)
+
+    # `collect` over GAP lists narrows the element type like the generic
+    # iteration based code does
+    @test collect(GAP.evalstr("[SymmetricGroup(3), SymmetricGroup(4)]")) isa Vector{GapObj}
+    @test collect(GAP.evalstr("[1, 2, true]")) isa Vector{Integer}
+    @test collect(GAP.evalstr("[1, SymmetricGroup(3)]")) isa Vector{Any}
+    @test collect(GAP.evalstr("[]")) isa Vector{Any}
+    l = GAP.evalstr("[1,,3]")
+    @test collect(l) isa Vector{Int}
+    @test [2 * x for x in l] == [2, 6]
+    @test [x for x in GAP.evalstr("[1, 2, 3]")] isa Vector{Int}
+
+    # an empty result gets its element type from inference,
+    # just like the generic iteration based code
+    @test [(a::GapObj) for a in GAP.evalstr("[]")] isa Vector{GapObj}
+    @test [2 * (a::Int) for a in GAP.evalstr("[]")] isa Vector{Int}
+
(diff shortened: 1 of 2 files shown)


test/adapter.jl
```diff
@@ -45,6 +45,28 @@
...
     @test_throws ArgumentError collect(GAP.Globals.GAPInfo)
+
+    # `collect` over GAP lists narrows the element type like the generic
+    # iteration based code does
+    @test collect(GAP.evalstr("[SymmetricGroup(3), SymmetricGroup(4)]")) isa Vector{GapObj}
+    @test collect(GAP.evalstr("[1, 2, true]")) isa Vector{Integer}
+    @test collect(GAP.evalstr("[1, SymmetricGroup(3)]")) isa Vector{Any}
+    @test collect(GAP.evalstr("[]")) isa Vector{Any}
+    l = GAP.evalstr("[1,,3]")
+    @test collect(l) isa Vector{Int}
```
lgoettgens, line 56: Could you also add a test of the form 

```suggestion
    @test collect(l) isa Vector{Int}
    @test collect(l) == [1, 3]
```
to check that `collect` handles holes correctly?

lgoettgens (approved):

Issue #902, opened 2023-06-17
GAP command Process hangs after IO_Popen3 is used on macOS


mohamed-barakat:
```
$ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.9.1 (2023-06-07)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

```
```julia
julia> using GAP
 ┌───────┐   GAP 4.12.2 of 2022-12-18
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: aarch64-apple-darwin20-julia1.9-64-kv8
 Configuration:  gmp 6.2.1, Julia GC, Julia 1.9.1, readline
 Loading the library and packages ...
 Packages:   AClib 1.3.2, Alnuth 3.2.1, AtlasRep 2.1.6, AutPGrp 1.11, CRISP 1.4.6, Cryst 4.1.25, CrystCat 1.1.10, CTblLib 1.3.4, FactInt 1.6.3, FGA 1.4.0, Forms 1.2.9,
             GAPDoc 1.6.6, genss 1.6.8, IO 4.8.1, IRREDSOL 1.4.4, JuliaInterface 0.9.6, LAGUNA 3.9.5, orb 4.9.0, Polenta 1.3.10, Polycyclic 2.16, PrimGrp 3.4.3,
             RadiRoot 2.9, recog 1.4.2, ResClasses 4.7.3, SmallGrp 1.5.1, Sophus 1.27, SpinSym 1.5.2, TomLib 1.2.9, TransGrp 3.6.3, utils 0.81
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'

julia> GAP.Globals.str = g""; GAP.Globals.Process( GAP.Globals.DirectoryCurrent(), GAP.Globals.Filename( GAP.Globals.DirectoriesSystemPrograms(), g"sh" ), GAP.Globals.InputTextUser(), GAP.Globals.OutputTextString( GAP.Globals.str, true ), GapObj( [ g"-c", g"echo 2" ] ) ); GAP.Globals.str
GAP: "2\n"

[...]

Issue #741, opened 2021-10-22
effects of GAP's Browse on Julia's screen output?


ThomasBreuer:
I am using Julia 1.6.3 with the develop version of GAP.jl.
The GAP package Browse has been installed via `GAP.Packages.load("Browse", install = true).

When I suspend the Julia session with Ctrl-Z and then enter the session again, the terminal gets cleared, the Julia prompt appears in the first line, and afterwards the distribution of Julia's output to lines does not fit anymore.
This problem disappears as soon as I deinstall the Browse package.

(This problem is sort of a continuation of #456, but since the effects are different, I opened a new issue instead of reopening #456.)

fingolfin:
It seems that pressing ctrl-z while Browse is active is somewhat broken even in regular GAP. E.g. when I enter `?Quotient` I am showing a pager list of help items to select; pressing ctrl-z here doesn't work very well. But in contrast, once I've selected a specific help item, ctrl-z works nicely. Huh.

ThomasBreuer:
I think we have to distinguish three different situations:
- If Browse has not yet been loaded then Ctrl-Z has its default behaviour.
- If Browse has beed loaded then some initialization has happened, and we get the effects sketched in the beginning of this discussion.
- If the visual mode of ncurses is active, the behaviour of Ctrl-Z is apparently different.

Concerning the example mentioned above:
Entering `?Quotient` at the `gap> ` prompt starts a Browse application if Browse is loaded, and then ncurses' visual mode is active. After entering Ctrl-Z, one is back at the shell prompt, but the cursor is still hidden. Calling `fg` takes one back to the application, but no redraw of the screen is triggered. (I have never thought about whether we could deal better with this situation.)
If one selects one of the entries shown in the window that appears when one enters `?Quotient` then ncurses' visual mode gets switched off, and the text is shown in a GAP pager, which is not based on ncurses.

paemurru:
> When I suspend the Julia session with Ctrl-Z and then enter the session again, the terminal gets cleared

The same behaviour occurs when using OSCAR without any other packages.

fieker:
On Sun, Mar 10, 2024 at 01:15:15AM -0800, Erik Paemurru wrote:
> > When I suspend the Julia session with Ctrl-Z and then enter the session again, the terminal gets cleared
> 
> The same behaviour occurs when using OSCAR without any other packages.
> 
polymake and Gap play with signal handlers?

fingolfin:
[...]

Issue #74, opened 2018-10-15
GAP catches interrupt


fingolfin:
Import from https://github.com/oscar-system/LibGAP.jl/issues/2

@ThomasBreuer wrote:
> When one loads libgap into julia then a double ''<ctr>-C'' entered in a long julia computation
> is apparently caught by GAP,
> with the effect that julia is left, with the well-known message that is printed by GAP:
> "gap: you hit '<ctr>-C' twice in a second, goodbye."
> 
> Is this intended?

@wbhart replied:
> We had the same problem when we used to call Pari from Nemo. We will have to work on handlers so that the two can cooperate.
> […](#)

@sebasguts replied:
> I think this might need some modification in LibGAP. A library should never implement a signal handler.
> 
> Since GAP sets the signal handler at start-up (in `sysfiles.c`), and GAP is loaded after Julia, and programs can only have one handler per signal, the Julia handler certainly gets overwritten. The proper solution would be that LibGAP exposes an interrupt-function, which could be used by the Julia signal handler to interrupt the GAP computation. I will talk to Markus about this issue. Probably a first step is to disable the signal handler for LibGAP, and then think about implementing some interrupt function.

fingolfin:
The problem exists in the other direction, too: When using `JuliaInterface`, using Ctrl-C immediately kills GAP instead of entering a break loop.

sebasguts:
Julia supports setting a signal handler, using `jl_install_sigint_handler`. Should we set it to the GAP `SIGINT` handler in JuliaInterface after `Julia` is loaded?

PR #1426, opened 2026-09-04
Signal handling: keep Julia's state, make Ctrl-C interrupt GAP

fingolfin:
GAP, Julia/libuv, and GAP packages (io, Browse/ncurses) each assume they own the process-wide signal state, child reaping, and the terminal. This PR sets an ownership contract for embedded GAP and implements the parts that work with released GAP and io. The contract and the facts about Julia it rests on are in the new manual page `docs/src/signals.md`.

### Startup and `GAP.prompt()`

GAP installs `SIGCHLD` and `SIGWINCH` handlers regardless of the `handleSignals` argument of `GAP_Initialize`, and autoloading Browse makes ncurses install a `SIGTSTP` handler. In an embedded session these belong to Julia (a foreign `SIGCHLD` handler breaks libuv's child handling on Linux; the terminal handlers garble the REPL, #741). `initialize` now saves and restores these dispositions and the terminal attributes (written back only if they changed, since `tcsetattr` from a background job raises `SIGTTOU`); `prompt()` hands the terminal and the error-handling state back in a `finally` block.

### Ctrl-C (#74, gap-system/gap#3256)

Julia's listener thread re-dispatches every `SIGINT` through the process disposition before deciding whether to raise an `InterruptException`. JuliaInterface installs a chaining `SIGINT` handler that uses this: while GAP code executes (tracked by a counter maintained around every entry into GAP and reset while GAP calls back into Julia) it calls GAP's `InterruptExecStat`, otherwise it forwards to Julia's handler. GAP's resulting `user interrupt` error surfaces as a Julia `InterruptException`.

- GAP acts at its next statement. If a call ends before that, e.g. a single kernel function, the interrupt is disarmed and the `InterruptException` is thrown when the call returns, instead of firing in a later, unrelated call.
- In `GAP.prompt()` Ctrl-C enters the break loop; at an idle prompt it is ignored, as in standalone GAP.
- `SyInstallAnswerIntr` is no longer used, so a double Ctrl-C can no longer terminate the Julia process from inside a signal handler.

No signal mask manipulation is involved. A trivial GAP call costs 6.9 ns, against 6.8 ns on master.

### Diagnostics and tests

`GAP.signal_report()` prints the current dispositions with the owning library of every handler. `test/signals.jl` pins the contract, checks that Julia subprocesses still work after GAP startup, and tests the interrupt bridge in-process and via `SIGINT` sent to subprocesses; `etc/julia.expect` gains the prompt scenarios.

### Not in this PR

io up to 4.10.0 reaps with `waitpid(-1)` from its own `SIGCHLD` handler, which hangs Julia's subprocess handling afterwards (#902) and made the `utils`/`curlInterface` distro tests hang (both excluded on master). Fixed by gap-packages/io#137 and gap-packages/io#138; once io 4.10.1 ships in `GAP_pkg_io_jll`, a follow-up can re-enable those tests and the Julia `ExecuteProcess`. Kernel-side wishes and the Browse ncurses initialization are listed under "Known problems" in the manual page.

Open question beyond this PR: if Singular.jl, Polymake.jl etc. copy the chaining handler, the chain depends on load order and on nobody replacing a link. A single dispatcher with registered runtimes would avoid that; GAP.jl's handler could become it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

Disclaimer: I have not yet proof-read most of this myself, so it might be full of slop. But I wanted it here so that CI can run, so that others can see it, and so that I don't forget about it (at least not permanently 😂).

---

CC @ChrisJefferson

### DIFF
--- a/.github/workflows/CI-distro.yml
+++ b/.github/workflows/CI-distro.yml
@@ -60,6 +60,7 @@ jobs:
           - gap-package: 'atlasrep'                   # random gc-related segfaults during testing
           - gap-package: 'autodoc'                    # tries and fails to build its own documentation, inside the (partially write protected) artifacts dir
           - gap-package: 'ctbllib'                    # random gc-related segfaults during testing
+          - gap-package: 'curlinterface'              # hangs: tests fork an HTTP server via `io`, but the forked child never receives the SIGTERM meant to stop it (julia blocks signals; its signal listener thread does not survive the fork), so a blocking IO_WaitPid never returns
           - gap-package: 'example'                    # no jll
           - gap-package: 'examplesforhomalg'          # `Error, found no GAP executable in PATH`
           - gap-package: 'io'                         # segfaults, most likely due to child process handling
@@ -68,6 +69,7 @@ jobs:
           - gap-package: 'packagemanager'             # many "#I  No sysinfo.gap found" warnings
           - gap-package: 'profiling'                  # segfaults during testing
           - gap-package: 'ringsforhomalg'             # `Error, found no GAP executable in PATH`
+          - gap-package: 'utils'                      # hangs: same forked HTTP test server problem as curlinterface (see above)
           - gap-package: 'xgap'                       # no jll
 
     steps:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,14 @@
 
 ## Version 0.17.6-DEV (released YYYY-MM-DD)
 
+- Ctrl-C now interrupts a running GAP computation (as an `InterruptException`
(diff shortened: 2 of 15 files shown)


lgoettgens:
I just un-requested my review. As long as it may be full of AI slop, I don't really wanna give this a review. Better just cc me in such cases. That way, you can use the "request review" feature once you actually want feedback on this. (and it leaves my "waiting for reviews" list on github in a more manageable state with less slop).

But once it actually works, this would be great to have!

PR #1430, opened 2026-09-05
Fix timeouts in REPL integration test on nightly

fingolfin:
The "Test REPL integration" step fails at random on Julia nightly, see e.g. [this run](https://github.com/oscar-system/GAP.jl/actions/runs/33941729929/job/101240365232) or [this one](https://github.com/oscar-system/GAP.jl/actions/runs/33931819774/job/101211794206). It always dies at the same spot: the tab completion of `GAP.Globals.GAPInfo.MaxNrArgs`. In passing nightly jobs that tab takes 19.4 to 19.8 s, against a 20 s expect timeout. So it is a coin flip, not a race.

Cause: the step runs julia with `--code-coverage`, i.e. mode `user`, which tracks every non-sysimage package, REPL included. The first completion needing type inference then recompiles the REPL completion machinery. Locally on 1.14-DEV: 14 s with coverage, 0.3 s without. On 1.13 the same cost lands on the first tab instead (13 s in CI).

Restricting coverage to `src` keeps REPL's precompiled code while `src/*.cov` files are still written, so Codecov is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -67,7 +67,7 @@ jobs:
           depwarn: error
       - name: "Test REPL integration"
         if: matrix.julia-version != '1.10'
-        run: etc/expect.sh --project=@. --code-coverage
+        run: etc/expect.sh --project=@. --code-coverage=@src
       - name: "Add optional Julia dependencies for GAP tests"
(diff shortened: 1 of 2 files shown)


lgoettgens (approved):
Nice! I didn't know one could restrict code coverage to some paths like that

fingolfin:
Neither did I before this 😂

PR #1432, opened 2026-09-07
Fix typos in comments and documentation

fingolfin:
Also correct stale references to renamed functions and files, a `##` comment line in JuliaInterface.gd that hid part of a sentence from the generated manual, and outdated type names in convert.gd.

Co-authored-by: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ GAP: SymmetricGroup( [ 1 .. 3 ] )
 ```
 
 The Julia types `Int64` and `Bool` are automatically converted to GAP
-objects when passed as arguments to GAP functions. Several others basic
+objects when passed as arguments to GAP functions. Several other basic
 types of objects can be converted using the `GapObj` constructor. For
 example, here we convert a `Vector{Int}` to a GAP list:
 ```julia
@@ -89,7 +89,7 @@ to report any issues you may encounter when using it. You can also submit
 feature requests and general help requests via that tracker.
 
 GAP.jl is being maintained by
-- Thomas Breuer <sam@math.rwth-aachen.de>>
+- Thomas Breuer <sam@math.rwth-aachen.de>
 - Lars Göttgens <goettgens@art.rwth-aachen.de>
 - Max Horn <mhorn@rptu.de>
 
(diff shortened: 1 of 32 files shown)


pkg/JuliaInterface/gap/convert.gd
```diff
@@ -253,7 +253,7 @@
 #!  </Item>
 #!  <Item>
 #!    <C>IsList</C> to
-#!    <C>Vector{Union{Any,Nothing}}</C> (default),
+#!    <C>Vector{Any}</C> (default),
```
lgoettgens, line 256: This is a non-obvious change. Why do we need it?

README.md
```diff
@@ -107,10 +107,10 @@ or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public
 License for more details.
 
 You should have received a copy of the GNU Lesser General Public License
-along with GAP.jl in form of the file `LICENSE`, or see
+along with GAP.jl in the form of the file `LICENSE`, or see
 <https://www.gnu.org/licenses/lgpl.html>.
 
-Copyright (C) 2017-2025 by its authors, which include:
+Copyright (C) 2017-2026 by its authors, which include:
```
lgoettgens, line 113: This is also something that just IMO not be buried in a random typo fixing PR

pkg/JuliaInterface/gap/convert.gd
ThomasBreuer: The change means to adjust the documentation to the code, see the function `GAP._default_type`.

pkg/JuliaInterface/gap/calls.gi
```diff
@@ -14,7 +14,7 @@ BindGlobal("_JL_Vector_Any", JuliaType( Julia.Vector, [ Julia.Any ] ));
 BindGlobal("_JL_Dict_Any", JuliaType( Julia.Dict, [ Julia.Symbol, Julia.Any ] ));
 
 ##
-##  We want to use &GAP's function call syntax also for certain Julia objects
+##  We want to use &GAP;'s function call syntax also for certain Julia objects
```
ThomasBreuer: if &GAP; then also &Julia;?

README.md
ThomasBreuer: According to the discussions in Oscar triage meetings, should the years range perhaps better get removed?

fingolfin: In GAP we keep the authors list and copyright year in a single file, and GAP.jl deliberately imitates that.

@lgoettgens what is the problem with "burying" this here? What do we gain by having an extra PR for changing a single digit?

pkg/JuliaInterface/gap/calls.gi
fingolfin: Actually this is not a doc comment, so...
```suggestion
##  We want to use GAP's function call syntax also for certain Julia objects
```

README.md
lgoettgens: There are people who like to discuss things like this change over and over again. And for that case, I prefer to have a dedicated place for such a discussion.
But, to be honest, I did not really think about the fact that this is the GAP.jl repo, and only the three of us really care about it.

lgoettgens (approved):

PR #1434, opened 2026-09-07
Make gap.sh honor use_active_project

fingolfin:
`create_gap_sh(dir; use_active_project=true)` computed the project directory but the generated script still used `--project=$(dirname "$0")`. As `dir` contains no Project.toml in that case, Julia fell back to the default environment, which need not contain GAP.jl. The script used by PackageManager to build GAP packages is generated this
way.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,9 +6,7 @@
 - Avoid boxing immediate integers in wrapped GAP calls (reduces
   allocations and improves performance)
 - Avoid `Core.Box` allocations in closures
-- Fix the `gap.sh` script that GAP.jl generates for building GAP packages
-  to run in the Julia project that was active when it was created, instead
-  of falling back to the default environment
+- Several "behind the scenes" changes that should be invisible for most users
 - Fix non-recursive conversion of GAP records to `Dict{Symbol,T}`: nested
   entries are now treated like those of lists, i.e. no further conversion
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -6,6 +6,9 @@
 - Avoid boxing immediate integers in wrapped GAP calls (reduces
   allocations and improves performance)
 - Avoid `Core.Box` allocations in closures
+- Fix the `gap.sh` script that GAP.jl generates for building GAP packages
```
lgoettgens: This is not the only purpose of the script. Maybe just drop that justification from the changelog

lgoettgens (approved):

CHANGES.md
```diff
@@ -6,6 +6,9 @@
...
+  to run in the Julia project that was active when it was created, instead
+  of falling back to the default environment
```
fingolfin:
```suggestion
- Several "behind the scenes" changes that should be invisible for most users
```

fingolfin: @lgoettgens like this perhaps?

lgoettgens: The version from the github email is different from the one here... Ahh, you edited it again.
I am fine with both versions

PR #1438, opened 2026-09-14
Support GAP's precise GC mode in JuliaInterface

fingolfin:
GAP master can be built with --enable-precise-gc, where the Julia GC no longer scans the C stack and every C local holding a GAP or Julia object across an allocation must be registered in a GC frame. Root the locals of JuliaInterface accordingly: the arguments converted one by one for a call in either direction, where converting one can collect the ones before it; the function, its name and body while a Julia function is wrapped; the error message while it is reported; the caller's local variables across each call into Julia, which a GAP error unwinding into Julia can leave otherwise unreferenced; and the error buffer until the module holds it. Annotate the functions for Julia's GC and safepoint analyses, which now pass on the package.

Against older kernels the macros and annotations compile away, in gc_compat.h, so the package still builds with GAP 4.14 to 4.16.

The sysinfo.gap written for package builds now keeps the defines of the GAP it links against, so a kernel extension compiled against a precise GAP sees DISABLE_STACK_SCAN and roots its locals too. The override script gains --precise for building such a GAP.

Assisted-by: Claude Code (Fable 5.1)
Assisted-by: Claude Code (Opus 5)


lgoettgens:
> GAP 4.16 can be built with --enable-precise-gc

do you mean GAP 4.17 after https://github.com/gap-system/gap/pull/6558 is merged?

PR #1443, opened 2026-09-19
Version 0.17.7

fingolfin:
Quick update to help fix the OSCAR test suite against Julia nightly.

Good to get this out of the way before further work is merged.


fingolfin:
still seeing issues with Julia nightly, caused by Singular errors

lgoettgens (approved):

lgoettgens:
The failing jobs now have different other errors that need some investigation

fingolfin:
Indeed the ABI and now also API of Julia just underwent another round of breaking changes

lgoettgens:
I think we are mostly done with the bandaids against the nightly breakage. I just started another rebuild of GAP_pkg_juliainterface_jll so that it uses the same libjulia version as GAP_jll (ref https://github.com/JuliaPackaging/Yggdrasil/pull/14906).
Once that is done, I think this release is good to go.

